PR #18, opened 2018-10-16
add the branch to the context of the status

jmcampanini:
this is the only instantiation of `github.RepoStatus` in policy-bot.

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -49,12 +49,13 @@ type PullEvaluationOptions struct {
 	AppName    string `yaml:"app_name"`
 	PolicyPath string `yaml:"policy_path"`
 
-	// StatusCheckContext will be used to create the status context.
-	// It will be used in the following pattern: <StatusCheckContext>: <Branch Name>
+	// StatusCheckContext will be used to create the status context. It will be used in the following
+	// pattern: <StatusCheckContext>: <Base Branch Name>
 	StatusCheckContext string `yaml:"status_check_context"`
 
-	// StatusCheckLegacyContext enables the sending of a second status using just StatusCheckContext, no templating.
-	// This is to support legacy workflows that depend on the legacy context behaviour. This is turned off by default.
+	// StatusCheckLegacyContext enables the sending of a second status using just StatusCheckContext as the context,
+	// no templating. This is to support legacy workflows that depend on the original context behaviour. This is turned
+	// off by default.
 	StatusCheckLegacyContext bool `yaml:"status_check_legacy_context"`
 }
 
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -74,8 +74,6 @@ func (p *PullEvaluationOptions) FillDefaults() {
 }
 
 func (b *Base) PostStatus(ctx context.Context, client *github.Client, owner, repo, ref string, state, message string, pr *github.PullRequest) error {
-	logger := zerolog.Ctx(ctx)
-
 	var detailsURL string
 	if pr != nil {
 		publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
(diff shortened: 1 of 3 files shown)


bluekeyes:
For compatibility, I think it's important that we continue to post the status check with the fixed context as well, so there's a transition period where people can update their branch protection settings without everything breaking. We'd eventually remove this in favor of only posting the named contexts.

My initial thought was to make this a configuration option that defaults to off. New users of the OSS version would get the new behavior only, but we could turn on this flag internally while we transition people. The `1.3.0` release would include a note that this has to be set to preserve backwards compatibility and then we remove it in a future `2.0.0` release.

jmcampanini:
i like this plan. may require more work, since i think we post from different places. but ill clean it up

jmcampanini:
alrighty, should be good for a review. centralized all status calls to `Base.PostStatus` and from there we post 1 or 2 statuses, depending on the legacy flag

server/handler/base.go
```diff
@@ -46,9 +46,17 @@ type Base struct {
...
+
+	// StatusCheckContext will be used to create the status context. It will be used in the following
+	// pattern: <StatusCheckContext>: <Base Branch Name>
 	StatusCheckContext string `yaml:"status_check_context"`
-	AppName            string `yaml:"app_name"`
+
+	// StatusCheckLegacyContext enables the sending of a second status using just StatusCheckContext as the context,
+	// no templating. This is turned off by default. This is to support legacy workflows that depend on the original
+	// context behaviour, and will be removed in 2.0
+	StatusCheckLegacyContext bool `yaml:"status_check_legacy_context"`
```
bluekeyes: I think it would be better to name this more like an option. The current name makes it seem like it should be a string.

```suggestion
	PostInsecureStatusChecks bool `yaml:"post_insecure_status_checks"`
```

server/handler/base.go
```diff
@@ -65,19 +73,45 @@ func (p *PullEvaluationOptions) FillDefaults() {
...
+		TargetURL:   &detailsURL,
+	}
+
+	logger.Info().Msgf("Setting status context=%s state=%s description=%s target_url=%s", contextWithBranch, state, message, detailsURL)
+	if _, _, err := client.Repositories.CreateStatus(ctx, owner, repo, ref, status); err != nil {
+		return err
 	}
 
-	return &github.RepoStatus{
+	if !b.PullOpts.StatusCheckLegacyContext {
```
bluekeyes: style: it might be more natural to do this affirmatively:

```go
if b.PullOpts.PostInsecureStatusChecks {
    // ... post the check
}
```

server/handler/base.go
```diff
@@ -65,19 +73,45 @@ func (p *PullEvaluationOptions) FillDefaults() {
...
+	legacyStatus := &github.RepoStatus{
 		Context:     &b.PullOpts.StatusCheckContext,
 		State:       &state,
 		Description: &message,
-		TargetURL:   &details,
+		TargetURL:   &detailsURL,
+	}
+
+	logger.Info().Msgf("Setting status context=%s state=%s description=%s target_url=%s", b.PullOpts.StatusCheckContext, state, message, detailsURL)
+	if _, _, err := client.Repositories.CreateStatus(ctx, owner, repo, ref, legacyStatus); err != nil {
```
bluekeyes: Is it worth defining a private method that creates the `RepoStatus`, logs this message, and makes the API call to avoid duplication?

server/handler/base.go
```diff
@@ -100,19 +134,15 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, mbrCtx pull.Membership
 
 	if fetchedConfig.Invalid() {
 		logger.Warn().Err(fetchedConfig.Error).Msgf("invalid policy: %s", fetchedConfig)
-
-		s := b.MakeStatus("error", fetchedConfig.Description(), pr)
-		_, _, err := client.Repositories.CreateStatus(ctx, fetchedConfig.Owner, fetchedConfig.Repo, srcSHA, s)
+		err := b.PostStatus(ctx, client, fetchedConfig.Owner, fetchedConfig.Repo, srcSHA, "error", fetchedConfig.Description(), pr)
```
bluekeyes, line 132: Looking at this in context, another design might be:

```go
status := b.MakeStatus(...)
b.PostStatus(ctx, client, owner, repo, sha, s)
```

`PostStatus` would clone the status object and change the context as needed.

jmcampanini: feel like that gives you the impression you can post your own status that you made with `MakeStatus`. since we always call both, feels simpler to have it be one method.

server/handler/base.go
jmcampanini: yea, going to clean this up a bit more, which should make the flow less exit-y

server/handler/base.go
```diff
@@ -65,19 +73,40 @@ func (p *PullEvaluationOptions) FillDefaults() {
...
+		status.Context = &b.PullOpts.StatusCheckContext
+		if err := b.postGitHubRepoStatus(ctx, client, status, owner, repo, ref); err != nil {
+			return err
+		}
 	}
+
+	return nil
+}
+
+func (b *Base) postGitHubRepoStatus(ctx context.Context, client *github.Client, status *github.RepoStatus, owner, repo, ref string) error {
```
bluekeyes: style: I think it's slightly more idiomatic to take the `status` as the last argument

bluekeyes (approved):
I'm not a fan of how many arguments `PostStatus` takes, but I'm not sure how else to do it. The fact that we pass a `nil` pull request in one spot mean we can't get the other info from the PR.

jmcampanini:
agreed, i dont like how many arguments and also dont like that we pass nil PR objects :(

PR #20, opened 2018-10-18
Fix status auditing and improve post status API

bluekeyes:
Fixes an NPE and a bad comparison when posting statuses for failed
status audits. Because the status audit code now mirrors the bad context
(to avoid looking up PR information), the base PostStatus method can
take fewer arguments and require a pull request.

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -73,12 +73,13 @@ func (p *PullEvaluationOptions) FillDefaults() {
 	}
 }
 
-func (b *Base) PostStatus(ctx context.Context, client *github.Client, owner, repo, ref string, state, message string, pr *github.PullRequest) error {
-	var detailsURL string
-	if pr != nil {
-		publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
-		detailsURL = fmt.Sprintf("%s/details/%s/%d", publicURL, pr.GetBase().GetRepo().GetFullName(), pr.GetNumber())
-	}
+func (b *Base) PostStatus(ctx context.Context, client *github.Client, pr *github.PullRequest, state, message string) error {
+	owner := pr.GetBase().GetRepo().GetOwner().GetLogin()
+	repo := pr.GetBase().GetRepo().GetName()
+	sha := pr.GetHead().GetSHA()
+
+	publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
+	detailsURL := fmt.Sprintf("%s/details/%s/%d", publicURL, pr.GetBase().GetRepo().GetFullName(), pr.GetNumber())
 
 	contextWithBranch := fmt.Sprintf("%s: %s", b.PullOpts.StatusCheckContext, pr.GetBase().GetLabel())
 	status := &github.RepoStatus{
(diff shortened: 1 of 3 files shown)


server/handler/base.go
```diff
@@ -73,12 +73,13 @@ func (p *PullEvaluationOptions) FillDefaults() {
...
-		publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
-		detailsURL = fmt.Sprintf("%s/details/%s/%d", publicURL, pr.GetBase().GetRepo().GetFullName(), pr.GetNumber())
-	}
+func (b *Base) PostStatus(ctx context.Context, client *github.Client, pr *github.PullRequest, state, message string) error {
+	owner := pr.GetBase().GetRepo().GetOwner().GetLogin()
+	repo := pr.GetBase().GetRepo().GetName()
+	sha := pr.GetHead().GetSHA()
+
+	publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
+	detailsURL := fmt.Sprintf("%s/details/%s/%d", publicURL, pr.GetBase().GetRepo().GetFullName(), pr.GetNumber())
```
jmcampanini: non blocking: could fullname be replaced by owner/repo ?

server/handler/base.go
```diff
@@ -73,12 +73,13 @@ func (p *PullEvaluationOptions) FillDefaults() {
 	}
 }
 
-func (b *Base) PostStatus(ctx context.Context, client *github.Client, owner, repo, ref string, state, message string, pr *github.PullRequest) error {
-	var detailsURL string
-	if pr != nil {
-		publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
-		detailsURL = fmt.Sprintf("%s/details/%s/%d", publicURL, pr.GetBase().GetRepo().GetFullName(), pr.GetNumber())
-	}
+func (b *Base) PostStatus(ctx context.Context, client *github.Client, pr *github.PullRequest, state, message string) error {
```
jmcampanini, line 76: i like having PR be required

server/handler/issue_comment.go
```diff
@@ -108,10 +108,7 @@ func (h *IssueComment) detectAndLogTampering(ctx context.Context, client *github
 		msg := fmt.Sprintf("Entity %s edited approval comment by %s", eventAuthor, commentAuthor)
 		logger.Warn().Str(LogKeyAudit, "issue_comment").Msg(msg)
 
-		repo := pr.GetBase().GetRepo()
-		srcSHA := pr.GetHead().GetSHA()
-
-		err := h.PostStatus(ctx, client, repo.GetOwner().GetLogin(), repo.GetName(), srcSHA, "failure", msg, nil)
+		err := h.PostStatus(ctx, client, pr, "failure", msg)
```
jmcampanini, line 111: so clean!

server/handler/status.go
```diff
@@ -60,10 +61,24 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
...
+			event.GetContext(),
+			commitSHA,
+			event.GetState(),
+			event.GetDescription(),
+			event.GetTargetURL(),
+		)
 		logger.Warn().Str(LogKeyAudit, eventType).Msg(auditMessage)
 
-		err := h.PostStatus(ctx, client, ownerName, repoName, commitSHA, "failure", auditMessage, nil)
+		status := &github.RepoStatus{
```
jmcampanini, line 78: wouldnt this only post 1 context

i guess the alternative is to create a PR object and fill in the data we have, but that feels brittle :( maybe just call the `postGitHubRepoStatus` call twice to at least have consistency?

bluekeyes: Posting a single context is intentional. In this code path, we're responding to someone who has overwritten any single context that we post. After detection, we want to replace that now-invalid context with a failure. If they overwrite both of our contexts, we'll handle two events, one for each context value. Otherwise, they probably only overwrote the context that is required by branch protection, so changing the other to be a failure isn't important.

I'll add comments about this to the code.

jmcampanini (approved):

PR #23, opened 2018-10-19
use branch from ref and not label

jmcampanini:
BUGFIX: Current behavior will create: `policy-bot: org-name:branch-name` which is incorrect. We do not want the `org-name` present.


asvoboda (approved):

jmcampanini:
re-eval

PR #30, opened 2018-10-31
Add download badges and links to README

bluekeyes:


bluekeyes:
Maybe the badges should be on the same line as title? I tried both and couldn't decide what was best.

jmcampanini (approved):
next line works for me. lgtm! :)

PR #37, opened 2018-12-06
Fix invalidate_on_push, use GraphQL for timeline

### LINKED ISSUE #9: switch to using github api v4
we should look into switching to using the graphql version of the API for GHE. this would give us `pushed` timestamps, making `invalidate_on_push` a lot simpler.

a sample query for this would look like:

```
{
  repository(owner: "gradle", name: "gradle") {
    pullRequest(number: 5813) {
      comments(first: 100) {
        pageInfo {
          endCursor
          hasNextPage
        }
        nodes {
          body
          createdAt
        }
      }
      reviews(first: 100) {
        nodes {
          createdAt
          author {
            login
          }
          state
        }
      }
      commits(first: 100) {
        nodes {
          commit {
            pushedDate
            author {
              user {
                login
              }
            }
            committer {
              user {
                login
              }
            }
          }
        }
      }
    }
  }
}
```

As a note, we still need to use the v3 API for changed files, since these are not exposed via GraphQL at this time.

bluekeyes:
The primary point of this PR is to fix a bug where commits are ordered by commit time (which is easily forged by clients) instead of push time. As a result, the `invalidate_on_push` option is currently a suggestion rather than a guarantee.

To fix this, we need the `pushedDate`, which as far as I can tell is only exposed in the v4 GraphQL API. The second part of the PR converts `pull.GitHubContext` to use GraphQL. This coincidentally cleans up some commit fetching logic, since we can get all the information at once now. Unfortunately, we still need the v3 client because PR file information is still in preview and even that isn't in GHE 2.15.x.

The final part of this change is updating the tests to work with GraphQL. This required rewriting how our mock `http.Transport` worked.

Some points for discussion:

* I pass two clients around everywhere. Would it be better to define an aggregate struct?
  ```go
  type Client struct {
    V3 *github.Client
    V4 *githubv4.Client
  }
  ```

* All the testing tools are defined in the test package via `*_test.go` files. Should this be extracted to a real package? If so, where? I'm not sure it's mature enough to move to `go-githubapp`.

* I pull in a GraphQL parsing library to implement the GraphQL matcher. This is potentially overkill. Is there another equally safe way to match on GraphQL query bodies?

Fixes #9.


pull/github.go
```diff
@@ -225,55 +175,146 @@ func (ghc *GitHubContext) Branches() (base string, head string, err error) {
...
+	var q struct {
+		Repository struct {
+			PullRequest struct {
+				Timeline struct {
+					PageInfo struct {
+						EndCursor   githubv4.String
+						HasNextPage bool
+					}
+					Nodes []*timelineEvent
+				} `graphql:"timeline(first: 100, after: $cursor)"`
```
asvoboda, line 187: Where does this `100` come from?

pull/github.go
```diff
@@ -225,55 +175,146 @@ func (ghc *GitHubContext) Branches() (base string, head string, err error) {
...
 		}
+		qvars["cursor"] = githubv4.NewString(q.Repository.PullRequest.Timeline.PageInfo.EndCursor)
+	}
 
-		opts.Page = res.NextPage
+	// GitHub only records pushedDate for the HEAD commit in a batch push
+	// Backfill this to all other commits for the purpose of sorting
+	var lastPushed *time.Time
+	for i := len(allEvents) - 1; i >= 0; i-- {
+		if event := allEvents[i]; event.Type == "Commit" {
```
asvoboda, line 215: Should we use `timelineevent#CreatedAt()` here?

pull/github.go
bluekeyes: It's an arbitrary page size, but 100 is the [maximum allowed by GitHub](https://developer.github.com/v4/guides/resource-limitations/#node-limit).

pull/github.go
bluekeyes: I don't think so - this loop fills in the underlying property that makes `CreatedAt()` function correctly for commit events, so I think the clarity of using the low-level property is desired.

bluekeyes:
I just realized that with the addition of valid timestamps on everything, we can probably remove the `Order` field from all of these types and use the times directly.

ajlake:
👍

PR #38, opened 2018-12-06
Correctly handle dismissed GitHub reviews

### LINKED ISSUE #28: Handle review dismissal event
If a GH review is performed with the result of request changes policybot will correctly fail the status check. If a person dismisses that review policybot will not reset the status check to pending.

bluekeyes:
When a review is dismissed, remove it from the review list as though it
never happened. This should be equivalent to a user deleting an approval
comment, although unlike in that case, we don't audit for this event.
Because GitHub provides the ability to restrict review dismissal, this
seems safe.

Fixes #28.

### DIFF
--- a/pull/context.go
+++ b/pull/context.go
@@ -125,4 +125,7 @@ type Review struct {
 	Author    string
 	State     ReviewState
 	Body      string
+
+	// ID is the GitHub node ID of the review, used to resolve dismissals
+	ID string
 }
--- a/pull/context.go
+++ b/pull/context.go
@@ -114,10 +114,10 @@ type ReviewState string
 
 const (
 	ReviewApproved         ReviewState = "approved"
-	ReviewChangesRequested             = "changes_requested"
-	ReviewCommented                    = "commented"
-	ReviewDismissed                    = "dismissed"
-	ReviewPending                      = "pending"
+	ReviewChangesRequested ReviewState = "changes_requested"
+	ReviewCommented        ReviewState = "commented"
+	ReviewDismissed        ReviewState = "dismissed"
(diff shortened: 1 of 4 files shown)


bluekeyes:
Note that #37 needs to merge first, then I'll update this to reference `develop`.

ajlake (approved):

bluekeyes:
After further testing, this may not be the best solution. GitHub does change the state of the review in the timeline to `DISMISSED` after a dismissal, but I suspect these is a delay here. When we process the webhook for the dismissal, the timeline still shows the original review, leading to the incorrect status. Any refresh event after the timeline is updated will show the correct status.

If the timeline already includes the dismissed event at the time of the webhook, this should solve it, although it would be better to change the state of the old review rather than remove it.

pull/context.go
```diff
@@ -114,15 +114,18 @@ type ReviewState string
...
 const (
 	ReviewApproved         ReviewState = "approved"
-	ReviewChangesRequested             = "changes_requested"
-	ReviewCommented                    = "commented"
-	ReviewDismissed                    = "dismissed"
-	ReviewPending                      = "pending"
+	ReviewChangesRequested ReviewState = "changes_requested"
+	ReviewCommented        ReviewState = "commented"
+	ReviewDismissed        ReviewState = "dismissed"
+	ReviewPending          ReviewState = "pending"
```
bluekeyes, line 120: @asvoboda sorry I asked you to change code like this in another project; I was wrong and the type declaration only carries in the specific case of using `iota`, i.e. you have to omit everything after the name for the type information of the first line to carry to the other lines.

asvoboda: No worries!

PR #39, opened 2018-12-11
Add IgnoreUpdateMerges option

### LINKED ISSUE #12: Allow branch updates/merges from target without invalidation
This makes it possible to keep branches up-to-date without (1) requiring reapproval and (2) counting the person who clicked merge as a contributor.

Thoughts on the implementation from the internal issue:

> I'm not sure how to detect this... maybe by allowing merge commits that don't themselves add any new changes or files to the branch?
> 
> In terms of implementation, there are two parts to consider: (1) how to identify a safe/clean commit and (2) how to ignore those.
> 
> The second is easiest: take a look at the `InvalidateOnPush` handling in [`IsApproved`](https://github.com/palantir/policy-bot/blob/develop/policy/approval/approve.go#L107). You can walk back the `lastCommitOrder` until the first unsafe commit. You'll also need to adjust the logic for finding contributors to ignore the authors of safe commits; that's [here](https://github.com/palantir/policy-bot/blob/develop/policy/approval/approve.go#L142).
> 
> For identifying commits, a good starting point is finding merge commits. This should be easy as they'll have more than one parent. Unfortunately, merge commits can also modify code, as when conflicts are resolved. I think you could do this by looking up the tree associated with the commit and seeing if it is empty, but I'm not quite sure how this is represented in Git.
> 
> We'll also have to decide if we only want to allow commits created by GitHub (or cleverly faked by users) or if local merges can count as well.
> 
> The last part is integrating that into the [GitHub implementation of `pull.Context`](https://github.com/palantir/policy-bot/blob/develop/pull/github.go), which wraps all the API logic and handles caching for the duration of a request. Try to minimize API calls, but correctness is more important than efficiency.

bluekeyes:
When this option is true, a commit on the PR branch is ignored if it meets these conditions:
    
1. Has two parents (i.e. is a simple merge commit)
2. Was created via the UI or the API
3. Has one parent in the last 100 commits on the target branch
    
An ignored commit does not invalidate approvals and the author/committer do not count as contributors.

Closes #12. See that issue for background on the problem and this solution.

This PR also refactors some of the test code to avoid destructively modifying the default test data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ UI to view the detailed approval status of any pull request.
     - [Disapproval is Disabled by Default](#disapproval-is-disabled-by-default)
     - [`or`, `and`, and `if` (Rule Predicates)](#or-and-and-if-rule-predicates)
     - [Cross-organization Membership Tests](#cross-organization-membership-tests)
+    - [Update Merges](#update-merges)
 * [Deployment](#deployment)
 * [Development](#development)
 * [Contributing](#contributing)
@@ -163,6 +164,13 @@ options:
   # approvals for this rule. False by default.
   invalidate_on_push: false
 
+  # If true, "update merges" do not invalidate approval (if invalidate_on_push
+  # is enabled) and their authors/committers do not count as contributors. An
+  # "update merge" is a merge commit that was created in the UI or via the API
+  # and merges the target branch into the pull request branch. These are
+  # commonly created by using the "Update branch" button in the UI.
+  ignore_update_merges: false
+
   # "methods" defines how users may express approval. The defaults are below.
   methods:
     comments:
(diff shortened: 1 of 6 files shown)


bluekeyes:
This also needs to be updated after #37 merges.

ajlake (approved):

bmoylan:
Are we able to check that no "merge conflict resolution" changes were added to the merge commit?

bluekeyes:
I haven't found a way to distinguish between an automatically resolved conflict in a file and a manually resolved conflict short of having policy-bot simulate the merge commit. If the merge is no a fast-forward, it will reference a new tree, and if a given file had changes in both parents, the tree will reference a new blob for the file that exists in neither parent. That blob is the final state of the file, so we don't have any information about how it was created.

By performing a tree comparison, I _could_ skip any merges where both branches modify the same file, but that (1) might block some common safe cases, leading to confusion and (2) is expensive in terms of API calls, requiring up to one call for each directory in the repository for each parent for each merge.

I don't actually know how GitHub implements this for GitHub Reviews, but I suspect they either have extra information that isn't exposed via the API or this is easier to solve if you have all the state about the PR history (e.g. if the merge state was conflict, and after a merge commit is pushed the state is mergeable, the merge commit must have resolved the conflict). We don't track any kind of state at the moment, so supporting that approach would require more work to set up and maintain a database or something.

README.md
```diff
@@ -294,6 +302,24 @@ Effectively, skipped rules are treated as if they don't exist.
...
+1. The commit must have exactly two parents
+2. The commit must have the `committedViaWeb` property set to `true`
+3. One parent must exist in the last 100 commits on the target branch of the
+   pull request
+
+These will all be true after updating a branch using the UI, but historic
+merges on long-running branches or merges created with the API may not be
+ignored. If this happens, you will need to reapprove the pull request.
+
+Note that `policy-bot` cannot detect if an update merge contains any merge
```
jmcampanini, line 319: that last sentence sounds a bit dangerous. i wonder how many of the use cases motivating this feature could be addressed with just requiring 2 approvals and allowing authors/contributors to be approvers

jmcampanini (approved):
lgtm

README.md
bluekeyes: I think the main use case is when you get approval, and then someone else merges another branch first, so now you have to update your branch and get approval again. That doesn't change based on the approval conditions.

For the case where you want to ignore other contributors who click update, then the double approval approach could work.

bluekeyes:
I'm going to merge this with the current limitations so we can start testing it with real usage. If it turns out that it needs to be stricter to be useful, we can revisit how to detect manual conflict resolution.

PR #40, opened 2018-12-12
Add a link to the policy file from the detail UI

bluekeyes:
The repository badge in the upper left corner now includes the target
branch name and is a link to the policy file.

Note that this does not resolve references: it always links to the `policy.yml` in your repository, not the one that defines the actual rules. While resolving references shows the actual policy definition, it may not be obvious that the current repository uses a reference if we link directly to the referenced file.

Here's a preview of the new button:

policy-definition-link

### DIFF
--- a/server/assets/css/main.css
+++ b/server/assets/css/main.css
@@ -3,7 +3,7 @@
 
 /* custom components */
 
-a, a:visited {
+a {
   color: config('colors.blue3');
 }
 
--- a/server/handler/details.go
+++ b/server/handler/details.go
@@ -17,6 +17,8 @@ package handler
 import (
 	"fmt"
 	"net/http"
+	"net/url"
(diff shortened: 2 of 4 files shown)


jmcampanini (approved):
nice! 

nit: i know that the button is clickable since i read this. i wonder if there's a way to make it more apparent, like an icon, or something? not sure what that would be though

bluekeyes:
When you hover over it, the text and background change color and the cursor icon switches to a pointer. Do you think that's enough? There's also title text that appears that says "View policy definition on GitHub".

I could also look at changing the colors to be blue or have an underline to make it more link-y.

jmcampanini:
pretty sure im just over-optimizing, as is should be good

asvoboda:
👍 snazzy

PR #43, opened 2018-12-17
Update permissions to require repository content

### LINKED ISSUE #42: Permission failure for private repos
In 1.4.0 we started using the timeline API, which seems to have some slightly unexpected permission changes for private repos. The result is that policy-bot is unable to evaluate policy ymls for private repos.

Error in the UI:
```
failed to compute approval status: failed to get approval candidates: failed to fetch comments: failed to list pull request timeline: Resource not accessible by integration
```

I think the fix is to probably update the required permissions, but we should test it out first.

bluekeyes:
With the switch to the GraphQL API, read-only permission for repository
contents is required read commit metadata.

Fixes #42.

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,12 +339,14 @@ that terminates TLS connections.
 
 `policy-bot` requires the following permissions as a GitHub app:
 
-* Issues - read-only
-* Repository metadata - read-only
-* Pull requests - read-only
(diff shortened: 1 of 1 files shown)


bluekeyes:
Note that I removed the single file permissions from the table, since they are redundant now.

asvoboda (approved):

PR #45, opened 2018-12-21
GitHub checks functionality

jamestoyer:
This integrates the ability to use the GitHub Checks API to post the results of the policies. This means the results will be included in the GH UI instead of jumping to a different UI, although you can still jump to the old view at the bottom of the checks page.

Here is an example of what this would look like screen shot 2018-12-21 at 14 45 39

Please go gentle as this is my first time using Go in anger


jamestoyer:
Replaced by #46

PR #46, opened 2019-01-07
GitHub checks functionality

jamestoyer:
This integrates the ability to use the GitHub Checks API to post the results of the policies. This means the results will be included in the GH UI instead of jumping to a different UI, although you can still jump to the old view at the bottom of the checks page.

Here is an example of what this would look like on the PR:

screen shot 2018-12-21 at 16 16 43

And on the checks page:
 
screen shot 2018-12-21 at 14 45 39

Replaces #45

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/google/go-github/github"
 	"github.com/palantir/go-baseapp/baseapp"
@@ -111,6 +112,63 @@ func (b *Base) postGitHubRepoStatus(ctx context.Context, client *github.Client, 
 	return err
 }
 
+func (b *Base) PostCheckResults(ctx context.Context, client *github.Client, pr *github.PullRequest, evaluationResult common.Result) error {
+	sha := pr.GetHead().GetSHA()
+	owner := pr.GetBase().GetRepo().GetOwner().GetLogin()
+	repo := pr.GetBase().GetRepo().GetName()
+	checkRunsRequest := &github.ListCheckRunsOptions{}
+
+	existingChecks, _, err := client.Checks.ListCheckRunsForRef(ctx, owner, repo, sha, checkRunsRequest)
+	if err != nil {
+		return err
+	}
+
+	if existingChecks.GetTotal() >= 0 {
+		if err := b.createGitHubRepoCheck(ctx, client, pr, owner, repo, sha, evaluationResult); err != nil {
+			return err
+		}
+	}
+
+	return nil
+
+}
+
+func (b *Base) createGitHubRepoCheck(ctx context.Context, client *github.Client, pr *github.PullRequest, owner string, repo string, sha string, evaluationResult common.Result) error {
+	publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
+	detailsURL := fmt.Sprintf("%s/details/%s/%s/%d", publicURL, owner, repo, pr.GetNumber())
+	defaultStatus := "completed"
+	time := github.Timestamp{
+		Time: time.Now(),
+	}
+
+	conclusion := "neutral"
+	switch evaluationResult.Status.String() {
+	case "approved":
+		conclusion = "success"
+	case "disapproved":
+		conclusion = "failure"
+	case "pending":
+		conclusion = "action_required"
+	}
+
+	status := &github.CreateCheckRunOptions{
+		Name: b.PullOpts.AppName,
+		DetailsURL: &detailsURL,
+		Status: &defaultStatus,
+		Output: &github.CheckRunOutput{
+			Title: &evaluationResult.Name,
+			Summary: &evaluationResult.Description,
+		},
+		StartedAt: &time,
+		CompletedAt: &time,
+		Conclusion: &conclusion,
+		HeadSHA: sha,
+	}
+
+	_, _, err := client.Checks.CreateCheckRun(ctx, owner, repo, *status)
+	return err
+}
+
 func (b *Base) Evaluate(ctx context.Context, mbrCtx pull.MembershipContext, client *github.Client, v4client *githubv4.Client, pr *github.PullRequest) error {
 	fetchedConfig, err := b.ConfigFetcher.ConfigForPR(ctx, client, pr)
 	if err != nil {
@@ -168,5 +226,10 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, mbrCtx pull.Membership
 	}
 
 	err = b.PostStatus(ctx, client, pr, statusState, statusDescription)
+	if err != nil {
+		return err
+	}
+
+	err = b.PostCheckResults(ctx, client, pr, result)
 	return err
 }
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -141,10 +141,11 @@ func (b *Base) createGitHubRepoCheck(ctx context.Context, client *github.Client,
 	case "skipped":
 		conclusion = "neutral"
 	case "disapproved":
+		conclusion = "failure"
 	case "unknown":
 		conclusion = "failure"
 	case "pending":
-		conclusion = "action_required"
+		conclusion = "failure"
 	}
 
 	var builder strings.Builder
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -124,7 +124,6 @@ func (b *Base) PostCheckResults(ctx context.Context, client *github.Client, pr *
 	}
 
 	return nil
-
 }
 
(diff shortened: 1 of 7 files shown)


jamestoyer:
I've added contextual information to the name of the check run, including the name of the Policy Bot and then name of branch

server/handler/base.go
```diff
@@ -58,6 +59,16 @@ type PullEvaluationOptions struct {
...
+
+	// DisableGitHubStatusChecks disable the sending of a status update to GitHub. This is enable by default as
+	// traditionally status checks were the only way to add status to a Pull Request. This can be replaced by the new
+	// GitHub Checks feature. See EnableGitHubCheckRuns
+	DisableGitHubStatusChecks bool `yaml:"disable_github_status_updates"`
+
+	// EnableGitHubCheckRuns enables posting of the results of the policy evaluations as GitHub checks. This means that
+	// that full results are visible on the checks tab of a GitHub pull request. By default this is turned off as the
+	// `checks:write` permission is required for the GitHub application
+	EnableGitHubCheckRuns bool `yaml:"enable_github_check_runs"`
```
bluekeyes, line 71: I like that this is configurable (and that you can disable the "old" method if necessary), but it's unfortunate you can configure the server to not post any results. Is there an alternate configuration that could avoid this, like a single option that takes 3 values? Or validation on startup?

server/handler/base.go
```diff
@@ -111,6 +122,87 @@ func (b *Base) postGitHubRepoStatus(ctx context.Context, client *github.Client,
...
+	case "approved":
+		conclusion = "success"
+	case "skipped":
+		conclusion = "neutral"
+	case "disapproved":
+		conclusion = "failure"
+	case "unknown":
+		conclusion = "failure"
+	case "pending":
+		conclusion = "failure"
```
bluekeyes: Rather than `pending` resulting in `completed` with a `failure` conclusion, does it make more sense to have the run remain in the `queued` or `in_progress` state in this case? That better matches the old commit statuses and suggests that the app is still waiting for something to happen.

server/handler/base.go
```diff
@@ -167,6 +259,15 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, mbrCtx pull.Membership
 		return errors.Errorf("evaluation resulted in unexpected state: %s", result.Status)
 	}
 
-	err = b.PostStatus(ctx, client, pr, statusState, statusDescription)
```
bluekeyes, line 170: Note that there are other locations in the code that call `PostStatus`; those probably need to be updated to handle check runs as well.

server/handler/check_run.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	switch event.GetAction() {
+	case "rerequested":
+		for _, pullRequest := range event.GetCheckRun().PullRequests {
+			ctx, _ = githubapp.PreparePRContext(ctx, installationID, event.GetRepo(), pullRequest.GetNumber())
+
+			// HACK: This gets around a lack of context from the PR associated with a check run. As the API might
+			// change later, this should be re-evaluated at a later date
+			pullRequest.Base.Repo = event.GetRepo()
```
bluekeyes: It might be safer to make an additional API call here to load the full PR object. As is, I could see us adding other code that uses missing parts of this object and miss it during testing, particularly if check runs are disabled.

server/handler/check_suite.go
```diff
@@ -0,0 +1,66 @@
...
+		return err
+	}
+
+	v4client, err := h.NewInstallationV4Client(installationID)
+	if err != nil {
+		return err
+	}
+
+	switch event.GetAction() {
+	case "rerequested":
```
bluekeyes, line 53: Does this also need to handle the `requested` event?

server/handler/base.go
```diff
@@ -111,6 +122,87 @@ func (b *Base) postGitHubRepoStatus(ctx context.Context, client *github.Client,
...
+	return err
+}
+
+func (b *Base) generateCheckRunSummary(results []*common.Result, level int) string {
+	var summaryBuilder strings.Builder
+	headerValue := strings.Repeat("#", 3+level)
+	indentValue := strings.Repeat(" ", 2*level)
+
+	for _, childResult := range results {
+		_, _ = summaryBuilder.WriteString("\n")
```
bluekeyes: style: you shouldn't have to ignore these return values... the latest version of `errcheck` includes exceptions for `strings.Builder` methods, but unfortunately the Godel asset containing that hasn't been released yet.

For now, you can add excludes to the `godel/config/checks.yml` file:

```yaml
checks:
  errcheck:
    config:
      exclude:
        - (*strings.Builder).Write
        - (*strings.Builder).WriteByte
        - (*strings.Builder).WriteRune
        - (*strings.Builder).WriteString
```

server/handler/base.go
```diff
@@ -111,6 +122,87 @@ func (b *Base) postGitHubRepoStatus(ctx context.Context, client *github.Client,
...
+		_, _ = summaryBuilder.WriteString(fmt.Sprintf("%s- %s %s (%s)", indentValue, headerValue, strings.Title(childResult.Name), childResult.Status.String()))
+		_, _ = summaryBuilder.WriteString("\n")
+		_, _ = summaryBuilder.WriteString(fmt.Sprintf("  %s%s", indentValue, childResult.Description))
+		childSummaries := b.generateCheckRunSummary(childResult.Children, level+1)
+		_, _ = summaryBuilder.WriteString(childSummaries)
+	}
+
+	return summaryBuilder.String()
```
bluekeyes, line 203: Did you look at using `text/template` for this, similar to how the UI version is rendered? I'm not sure if it would be simpler, but a recursive template might be easier to understand than the string formatting indentation here.

server/handler/base.go
```diff
@@ -111,6 +122,87 @@ func (b *Base) postGitHubRepoStatus(ctx context.Context, client *github.Client,
 	return err
 }
 
+func (b *Base) PostCheckResults(ctx context.Context, client *github.Client, pr *github.PullRequest, evaluationResult common.Result) error {
+	sha := pr.GetHead().GetSHA()
+	owner := pr.GetBase().GetRepo().GetOwner().GetLogin()
+	repo := pr.GetBase().GetRepo().GetName()
+
+	for _, result := range evaluationResult.Children {
+		if err := b.createGitHubRepoCheck(ctx, client, pr, owner, repo, sha, *result); err != nil {
```
bluekeyes, line 131: This ends up creating both approval and disapproval runs, right? Would it be better to combine them into a single run or does the fact that they are part of the same suite effectively combine them?

bluekeyes (commented):
Overall, this looks good. I'll have to do some research into how check runs are added to branch protection to make sure that part still works.

server/handler/base.go
jamestoyer: Updated the Godel config to ignore `strings.Builder` and revoked the previous changes to make this look a little less cluttered

server/handler/base.go
jamestoyer: I've changed `pending` to have a `action_required` conclusion. This better follows the checks workflow whereby a check that has been run must have a completed status when it is not in progress. It also makes sure that an "update" isn't needed on a check, thus ensuring we don't (currently) need to store state in the bot.

Below is the workflow suggested by [GitHub](https://developer.github.com/v3/checks/runs/)

server/handler/check_suite.go
jamestoyer: The `check_run` and `check_suite` events exist to capture when a user manually re-requests a suite/run to re-run. The other statuses would actually be triggered by the initial creation of a run/suite which are done automatically when posting a new run is created or a new suite is created. So in this case `rerequested` is the only event we care about

See:
- https://developer.github.com/v3/activity/events/types/#checksuiteevent
- https://developer.github.com/v3/activity/events/types/#checkrunevent

server/handler/base.go
jamestoyer: That's correct. In my mind it more clearly splits out what has been the failed checks and the checks that have not. However, as they are both part of the same suite, which is tied to a ref, it shouldn't really matter underneath

jamestoyer:
@bluekeyes With branch permissions it works exactly the same as traditional statuses do. You should be able to create protections in the same way you do normal statuses

server/handler/base.go
jamestoyer: Just to add, this does now mean there is no manual way to force the checks to re-run, it will always require an update to the rest of the PR. 

In the example below, the first check run has a status of `action_required` and the second has one of `failure`

image

NargiT:
is this dead ?

wrslatz:
Looks like a new issue for this has been opened: https://github.com/palantir/policy-bot/issues/348

Issue #47, opened 2019-01-18
invalidate_on_push fails open if GitHub APIs are delayed


bluekeyes:
We're currently investigating an internal issue where GitHub Enterprise seems to be sending webhooks before the commits referenced by those hooks are available in other APIs. Because policy-bot uses only the PR timeline to find new commits, it can end up posting an incorrect approved status on the latest commit.

At minimum, we should make sure that the commit referenced by the webhook we are processing appears in the commit list. If it does not, we could delay and try again or immediately fail. Either option will be more correct than the current behavior.

PR #54, opened 2019-03-12
Remove usage of pull request timeline API

bluekeyes:
We originally used this API to have consistent ordering between comments, commits, and reviews. However, this didn't respect commit push times, so we switched to using timestamps for ordering with the switch to GraphQL. Later, we discovered that the timeline API has a random delay before commits, particularly from forks, appear in the result. This causes the bot to evaluate the status for the wrong commit, breaking the `invalidate_on_push` option.

Instead, get commits, comments, and reviews from the pull request API, which should be up-to-date when we respond to a `pull_request` hook. Because all three of these sources are used in most evaluations, continue to load them with a single query instead of making separate queries for each type.

See #47 for more details. This will hopefully fix the problem, but I'll submit a separate PR that adds a check for the head commit.


bluekeyes:
This also updates `stretchr/testify` to fix a bug that caused failed assertions to panic instead of printing a diff in certain cases.

ajlake (approved):

PR #58, opened 2019-04-05
Improve logging of commit information

bluekeyes:
Attach the head SHA of the pull request being evaluated to log messages,
log when candidates are removed due to a pushed commit, and log the
target ref when posting a commit status.

### DIFF
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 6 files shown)


server/handler/base.go
```diff
@@ -106,11 +106,20 @@ func (b *Base) PostStatus(ctx context.Context, client *github.Client, pr *github
...
-	logger.Info().Msgf("Setting status context=%s state=%s description=%s target_url=%s", status.GetContext(), status.GetState(), status.GetDescription(), status.GetTargetURL())
+	logger.Info().Msgf("Setting %q status on %s to %s: %s", status.GetContext(), ref, status.GetState(), status.GetDescription())
 	_, _, err := client.Repositories.CreateStatus(ctx, owner, repo, ref, status)
 	return err
 }
 
+func (b *Base) PreparePRContext(ctx context.Context, installationID int64, pr *github.PullRequest) (context.Context, zerolog.Logger) {
+	ctx, logger := githubapp.PreparePRContext(ctx, installationID, pr.GetBase().GetRepo(), pr.GetNumber())
+
+	logger = logger.With().Str("github_sha", pr.GetHead().GetSHA()).Logger()
```
bluekeyes, line 117: I decided to put this here instead of in `githubapp` because there are situations when you might take action in response to a PR event that have nothing to do with the current SHA. It's also unclear if you would always want to log the head SHA.

policy/approval/approve.go
```diff
@@ -128,15 +129,20 @@ func (r *Rule) IsApproved(ctx context.Context, prctx pull.Context) (bool, string
...
 		for _, candidate := range candidates {
-			if candidate.CreatedAt.After(lastCommitTime) {
+			if candidate.CreatedAt.After(lastCommit.CreatedAt) {
 				allowedCandidates = append(allowedCandidates, candidate)
 			}
 		}
+
+		log.Debug().Msgf("discarded %d candidates invalidated by push of %s at %s",
+			len(candidates)-len(allowedCandidates),
+			lastCommit.SHA,
```
jamestoyer, line 143: Is this the full SHA of the commit or just substring of it? If it's the former, do we need to have the full SHA? May places like GitHub use the first 7-8 characters

bluekeyes: This should be the full SHA: it comes from the API, which I believe only returns full values. Since this is for debugging, I don't mind having the full value, even if it makes the message longer.

ajlake (approved):

policy/approval/approve.go
jamestoyer: Cool, makes sense in this case

PR #59, opened 2019-04-05
Fix repository content permissions in README

bluekeyes:
When the table was copied from palantir/bulldozer, I missed fixing the
repository content permission, which should be read-only. This is
because we read commit metadata to support invalidate_on_push and
author/contributor checks.


ajlake:
👍

PR #61, opened 2019-04-12
Fix invalidate_on_push for fork pull requests

bluekeyes:
GitHub does not propagate all commit information from forks to the
upstream repository when creating a pull request. In particular,
pushedDate is always null when listing commits in a fork via the pull
request APIs. To fix this, always list commits directly from the head
repository.

As a consequence, this also removes the 250 commit limit, although any
PR with more than 250 commits is likely to also hit the 300 file limit.

Finally, reorganize the test data and construct the default test PR in
code instead of optionally loading it from a mock request.


bluekeyes:
I've testing manually with an app token to make sure it can read the commit information from forks, and it seems to be able to, but I'll also test with my development server just to make sure before this goes to staging.

pull/github.go
```diff
@@ -149,30 +149,60 @@ func (ghc *GitHubContext) ChangedFiles() ([]*File, error) {
...
+		var commits []*v4Commit
+		for len(commits) < totalCommits {
+			if err := ghc.v4client.Query(ghc.ctx, &q, qvars); err != nil {
+				return nil, errors.Wrap(err, "failed to list commits")
+			}
+			commits = append(commits, q.Repository.Object.Commit.History.Nodes...)
 
-	// verify that the head of the PR being evaluated exists in commit list
-	// some GitHub APIs have a delay propagating commit information
-	headSHA := ghc.pr.GetHead().GetSHA()
```
bluekeyes, line 163: I removed this check because we now list commits starting with the head SHA directly. If it doesn't exist, we'll get an empty response from the API call, which will either fail to deserialize or will fail on the commit count check.

bluekeyes:
Testing with a fork and my development sever looks good. I've also added a check to make sure push dates are set, which will fail if this style of thing happens again.

pull/github.go
```diff
@@ -149,30 +149,65 @@ func (ghc *GitHubContext) ChangedFiles() ([]*File, error) {
...
+			} `graphql:"repository(owner: $owner, name: $name)"`
+		}
+		qvars := map[string]interface{}{
+			// always list commits from the head repository
+			// some commit details do not propagate from forks to the upstream
+			"owner": githubv4.String(ghc.pr.GetHead().GetRepo().GetOwner().GetLogin()),
+			"name":  githubv4.String(ghc.pr.GetHead().GetRepo().GetName()),
+			"oid":   githubv4.GitObjectID(ghc.pr.GetHead().GetSHA()),
+
+			"limit":  githubv4.Int(min(totalCommits, MaxPageSize)),
```
ajlake: Could we remove the `min` logic and just ask for `MaxPageSize` every time? This assumes that Github will return _at most_ `MaxPageSize` results and not error out.

ajlake (approved):
Looks good. It's minor but I think the query loop would be clearer written this way:
```go
        qvars := map[string]interface{}{
            // always list commits from the head repository
            // some commit details do not propagate from forks to the upstream
            "owner": githubv4.String(ghc.pr.GetHead().GetRepo().GetOwner().GetLogin()),
            "name":  githubv4.String(ghc.pr.GetHead().GetRepo().GetName()),
            "oid":   githubv4.GitObjectID(ghc.pr.GetHead().GetSHA()),
            "cursor": (*githubv4.String)(nil),
        }

        var commits []*v4Commit
        for {
            qvars["limit"] = githubv4.Int(min(totalCommits-len(commits), MaxPageSize))
            if err := ghc.v4client.Query(ghc.ctx, &q, qvars); err != nil {
                return nil, errors.Wrap(err, "failed to list commits")
            }
            commits = append(commits, q.Repository.Object.Commit.History.Nodes...)
            
            if len(commits) == totalCommits {
                break
            }

            if !q.Repository.Object.Commit.History.PageInfo.UpdateCursor(qvars, "cursor") {
                return nil, errors.Errorf("pull request has %d commits, but API only returned %d", totalCommits, len(commits))
            }
        }
```

PR #65, opened 2019-04-16
Use values, not pointers, in GraphQL queries

bluekeyes:
I'm not sure if this is needed, but after switching how commits are
loaded in 1.7.0, we started seeing weird race-condition-like issues
where the program would crash or randomly be missing information. One
incidental change was that commits switched from being values to being
pointers in the GraphQL query struct. My best theory right now is that
something about how `shurcoolL/graphql` unmarshals responses led to
unintentional data sharing, but I don't have a full explanation.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -157,7 +157,7 @@ func (ghc *GitHubContext) Commits() ([]*Commit, error) {
 					Commit struct {
 						History struct {
 							PageInfo v4PageInfo
-							Nodes    []*v4Commit
+							Nodes    []v4Commit
 						} `graphql:"history(first: $limit, after: $cursor)"`
 					} `graphql:"... on Commit"`
 				} `graphql:"object(oid: $oid)"`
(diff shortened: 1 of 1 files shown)


bluekeyes:
On second thought, I'm going to try to get the race detector running on our production instance before merging this. Everything I found so far suggests memory corruption, but I haven't found it via inspection and the race detector doesn't see anything in our tests or in my development server.

ajlake:
👍

ajlake (approved):

bluekeyes:
After investigation, this is not needed, but I think we'll leave it in anyway for consistency with the `githubv4` examples.

The actual problem was an index-out-of-bounds error (fixed in #68 and #69) that manifested as a segfault due to a bug with bounds check elimination in `go1.11`. That was fixed in `go1.11.1`, so upgrading our Go version would have revealed the actual bug; this implicitly happened when I built the race-detector binary on my local machine.

PR #66, opened 2019-04-17
Fix data race using default approval methods

bluekeyes:
The default approval and disapproval methods were stored as global
variables but then modified by other methods. While the written data was
always the same and it seemed to work fine historically, we started
seeing memory corruption errors in the 1.7.0 release and the race
detector caught this problem.


bluekeyes:
An alternative approach is defining functions that return the defaults, but it seemed easier to inline them since they were only used in one place.

ajlake (approved):

PR #67, opened 2019-04-17
Upgrade Go to 1.12 and Godel to 2.16.0

bluekeyes:
Switch to the standard CircleCI Docker images, as I don't think we
used anything that was part of the custom images we used before.


asvoboda:
👍 we can always enroll in excavator at a later date

PR #68, opened 2019-04-19
Load full PR if it is missing required fields

### LINKED ISSUE #63: Contributors ignored during a single evaluation
**Edit:** see [comment](https://github.com/palantir/policy-bot/issues/63#issuecomment-483497069) for a summary of the underlying bug.

----

I left a comment on a single line of this PR (https://github.com/palantir/gradle-conjure/pull/136), but it seems it caused the entire thing to merge.

Screenshot 2019-04-15 at 13 15 10

Interestingly, the UI seems to still say the PR has not yet been approved, but the status check went green so bulldozer mnerged it.

Screenshot 2019-04-15 at 11 57 15

bluekeyes:
The PR objects included in webhook payloads are oddly inconsistent: the
pull_request_review payload does not include the "commits" field, which
is require for policy-bot to function. Instead of loading PRs in each
handler, modify the context to load the PR if the one it was given is
missing any required fields. To make this harder to mess up, use a new
pull request object internally based on the GraphQL query.

Fixes #63.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -16,7 +16,6 @@ package pull
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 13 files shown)


bluekeyes:
The next step here would be to refactor `NewGitHubContext` to take something that minimally provides the locator information but might be cast to a full `github.PullRequest`. That would save us from loading the pull request explicitly in the `issue_comment` handler.

pull/github.go
```diff
@@ -103,7 +135,11 @@ func (ghc *GitHubContext) RepositoryName() string {
 }
 
 func (ghc *GitHubContext) Author() (string, error) {
-	return ghc.pr.GetUser().GetLogin(), nil
+	pr, err := ghc.get()
+	if err != nil {
```
ajlake: Should we just load the PR at construction time? Maybe `v3PullRequestToV4` can also take in the Github client necessary to populate the `v4` object if there is insufficient information in the v3 object, instead of returning `nil`.

In that case, we'd probably want to move the `case` checks into a helper:
```
v3PullRequestToV4(...):
    if isSufficient(v3) {
      // seed from v4
    }

    // else graphql query
```

bluekeyes: I introduced the concept of a `Locator` which contains an optional `github.PullRequest`. If that is missing or incomplete, the full value is now loaded at construction time. I also split out the completeness test as you suggested.

ajlake (approved):

PR #71, opened 2019-04-19
Add modified_lines predicate

### LINKED ISSUE #70: FR: Add a min_lines_of_code predicate
In my repository, I'd like to require special approval for pull requests over a certain size. E.g. if more than 1000 lines of code have been modified, a certain set of people must approve the pull request.

Would it be possible to add a predicate to support this?

bluekeyes:
Allows enabling rules based on the number of lines added or deleted by a
pull request. Each property takes simple expression representing a
comparison and the predicate is satisfied if any expression is matched.

Fixes #70.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,10 +145,18 @@ if:
     organizations: ["org1", "org2", ...]
     teams: ["org1/team1", "org2/team2", ...]
 
-  # "targets_branch" is satisfied if the target branch on the pull request
+  # "targets_branch" is satisfied if the target branch of the pull request
   # matches the regular expression
   targets_branch:
     pattern: "^(master|regexPattern)$"
+
+  # "modified_lines" is satisfied if the number of lines added or deleted by
+  # the pull request matches any of the listed conditions. Each expression is
+  # an operator (one of '<' or '>'), an optional space, and a number.
+  modified_lines:
+    additions: "> 100"
+    deletion: "> 100"
+    total: "> 200"
 
 # "options" specifies a set of restrictions on approvals. If the block does not
 # exist, the default values are used.
--- a/policy/predicate/file.go
+++ b/policy/predicate/file.go
@@ -16,7 +16,9 @@ package predicate
 
 import (
 	"context"
+	"fmt"
 	"regexp"
+	"strconv"
 
 	"github.com/pkg/errors"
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -145,11 +145,19 @@ if:
...
   targets_branch:
     pattern: "^(master|regexPattern)$"
 
+  # "modified_lines" is satisfied if the number of lines added or deleted by
+  # the pull request matches any of the listed conditions. Each expression is
+  # an operator (one of '<' or '>'), an optional space, and a number.
+  modified_lines:
+    additions: "> 100"
+    deletion: "> 100"
+    total: "> 200"
```
jmcampanini, line 159: are all combinations of these 3 allowed?

jmcampanini (approved):

README.md
bluekeyes: Yes, the code allows all combinations of the three, as there's a clear way to evaluate all of them and combine the result. That said, I'm not sure if there are any practical policies that would make use of all three in all combinations.

policy/predicate/file.go
```diff
@@ -85,6 +87,75 @@ func (pred *OnlyChangedFiles) Evaluate(ctx context.Context, prctx pull.Context)
...
+)
+
+func (exp ComparisonExpr) IsEmpty() bool {
+	return exp == ""
+}
+
+func (exp ComparisonExpr) Evaluate(n int64) (bool, error) {
+	match := numCompRegexp.FindStringSubmatch(string(exp))
+	if match == nil {
+		return false, errors.Errorf("invalid comparsion expression: %q", exp)
```
ajlake:
```suggestion
		return false, errors.Errorf("invalid comparison expression: %q", exp)
```

policy/predicate/file.go
```diff
@@ -85,6 +87,75 @@ func (pred *OnlyChangedFiles) Evaluate(ctx context.Context, prctx pull.Context)
...
+		return false, errors.Wrapf(err, "invalid commparison expression: %q", exp)
+	}
+
+	switch op {
+	case "<":
+		return n < v, nil
+	case ">":
+		return n > v, nil
+	}
+	return false, errors.Errorf("invalid comparsion expression: %q", exp)
```
ajlake:
```suggestion
	return false, errors.Errorf("invalid comparison expression: %q", exp)
```

policy/predicate/file.go
```diff
@@ -85,6 +87,75 @@ func (pred *OnlyChangedFiles) Evaluate(ctx context.Context, prctx pull.Context)
...
+
+	for expr, v := range map[ComparisonExpr]int64{
+		pred.Additions: additions,
+		pred.Deletions: deletions,
+		pred.Total:     additions + deletions,
+	} {
+		if !expr.IsEmpty() {
+			ok, err := expr.Evaluate(v)
+			if err != nil {
+				return false, "", err
```
ajlake, line 147: Is the returned string only for errors or should it also indicate the positive cases for why something matched? I don't recall.

policy/predicate/file.go
```diff
@@ -85,6 +87,75 @@ func (pred *OnlyChangedFiles) Evaluate(ctx context.Context, prctx pull.Context)
...
+			}
+			if ok {
+				return true, "", nil
```
ajlake, line 150: It's probably worth calling out in the README that the options are combined using logical `or` (and not _all_ must be true).

policy/predicate/file.go
bluekeyes: Yeah, it's only used when the predicate does not match to describe why it did not match.

policy/predicate/file.go
bluekeyes: The current comment says `... matches any of the listed conditions` - do you think it needs to be more obvious?

PR #79, opened 2019-05-06
Clarify error messages when listing commits

bluekeyes:
After observation, the head commit is only missing when the pull request
is updated by a force-push immediately after some other event. The first
event includes a payload with the old commit information, but only the
new commit is available by the time the server lists the pull request
commits from the API. Still return an error in this case (since we can't
correctly evaluate the PR as requested), but avoid retrying and clarify
the error message. In practice, this error won't appear to users because
the failed commit was removed from the pull request.


jmcampanini (approved):

bluekeyes:
// policy-bot

PR #84, opened 2019-06-04
Support policy files up to 100 MB

### LINKED ISSUE #83: Use Github Data API to support policy files greater than 1 MB
We have a generated policy file that is currently about 1.08 MB, and policy-bot fails to deal with it.

The error reported is:
```
Failed to fetch configuration at ref=develop: failed to fetch content of org/repo@master/central-policy.yml: GET https://githuburl/api/v3/repos/org/repo/contents/central-policy.yml?ref=master: 403 This API returns blobs up to 1 MB in size. The requested blob is too large to fetch via the API, but you can use the Git Data API to request blobs up to 100 MB in size. [{Resource:Blob Field:data Code:too_large Message:}]
```

dsw2127:
Previous usage of Git Repos API only supports policy files with max size of 1 MB. This adds a fall-back to use the Git Data API if the initial call fails as the file is too large.

The DownloadContents method could be used in place of GetContents, but since this code is called very often and DownloadContents makes an additional API call to list the tree to acquire the file SHA,
and this is an edge case, I have implemented it as a fallback instead (per Billy's suggestion)

Fixes #83

### DIFF
--- a/server/handler/fetcher.go
+++ b/server/handler/fetcher.go
@@ -159,12 +159,15 @@ func (cf *ConfigFetcher) fetchConfigContents(ctx context.Context, client *github
 		}
 		if ok && cf.hasTooLargeError(rerr) {
 			// GetContents only supports file sizes up to 1 MB, DownloadContents supports files up to 100 MB (with an additional API call)
-			reader, downloadErr := client.Repositories.DownloadContents(ctx, owner, repo, path, opts)
-			if downloadErr != nil {
-				return nil, errors.Wrapf(downloadErr, "failed to download content of %s/%s@%s/%s", owner, repo, ref, path)
+			reader, err := client.Repositories.DownloadContents(ctx, owner, repo, path, opts)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to download content of %s/%s@%s/%s", owner, repo, ref, path)
 			}
 			downloadedContent, readErr := ioutil.ReadAll(reader)
-			reader.Close()
+			closeError := reader.Close()
+			if closeError != nil {
+				return nil, errors.Wrapf(closeError, "failed to close reader for %s/%s@%s/%s", owner, repo, ref, path)
+			}
 			if readErr != nil {
 				return nil, errors.Wrapf(readErr, "failed to read content of %s/%s/@%s/%s", owner, repo, ref, path)
 			}
--- a/server/handler/fetcher.go
+++ b/server/handler/fetcher.go
@@ -17,6 +17,7 @@ package handler
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


dsw2127:
When GetContents is called on a file > 1 MB, the response has a 403 http status and body:
```
{
  "message": "This API returns blobs up to 1 MB in size. The requested blob is too large to fetch via the API, but you can use the Git Data API to request blobs up to 100 MB in size.",
  "errors": [
    {
      "resource": "Blob",
      "field": "data",
      "code": "too_large"
    }
  ],
  "documentation_url": "https://developer.github.com/enterprise/2.16/v3/repos/contents/#get-contents"
}
```

server/handler/fetcher.go
```diff
@@ -179,3 +197,12 @@ func (cf *ConfigFetcher) unmarshalConfig(bytes []byte) (*policy.Config, error) {
 
 	return &config, nil
 }
+
+func (cf *ConfigFetcher) hasTooLargeError(errorResponse *github.ErrorResponse) bool {
```
bluekeyes: style: I'd call this `isTooLargeError` and make it a plain non-exported function instead of attaching it to `ConfigFetcher`

server/handler/fetcher.go
```diff
@@ -152,9 +153,26 @@ func (cf *ConfigFetcher) fetchConfigContents(ctx context.Context, client *github
...
 			return nil, nil
 		}
+		if ok && cf.hasTooLargeError(rerr) {
+			// GetContents only supports file sizes up to 1 MB, DownloadContents supports files up to 100 MB (with an additional API call)
+			reader, err := client.Repositories.DownloadContents(ctx, owner, repo, path, opts)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to download content of %s/%s@%s/%s", owner, repo, ref, path)
+			}
+			downloadedContent, readErr := ioutil.ReadAll(reader)
+			closeError := reader.Close()
```
bluekeyes: I think it's a bit more idiomatic to handle this in a `defer` and either set a named return value or log it as a warning/error:

```go
reader, err := client.Repositories.DownloadContents(ctx, owner, repo, path, opts)
if err != nil {
  return nil, errors.Wrapf(err, "failed to download content of %s/%s@%s/%s", owner, repo, ref, path)
}

defer func() {
  if cerr := reader.Close(); cerr != nil {
    logger.Error().Err(cerr).Msgf(...)
  }
}()

content, err := ioutil.ReadAll(reader)
// ...
```

I'd go for the log over returning an error since I don't think a failure to close this reader should block further evaluation of the request.

bluekeyes (commented):
Thanks for implement this! Two minor style comments.

I didn't realize `DownloadContent` was a function in the `github` package and that it only requires one additional call - that's useful to know about.

server/handler/fetcher.go
dsw2127: Makes sense, thanks for the advice. I agree!

server/handler/fetcher.go
dsw2127: done

dsw2127:
Yeah I started to dig into the API until I realized that method existed - I was grateful to not have to implement that

bluekeyes (approved):

PR #91, opened 2019-06-24
Use go-githubapp client caching

asvoboda:
Similar in purpose to https://github.com/palantir/bulldozer/pull/129


bluekeyes (approved):
Note that this will be less helpful in policy-bot than it was in Bulldozer due to the use of the v4 API. I think we'll only end up cacheing policy file lookups and the files in each PR, at least until we move file lookup to the v4 API as well.

PR #97, opened 2019-07-03
Add debug logging for GitHub request retries

bluekeyes:
While we log individual requests already, it is hard to separate these
retries from other requests because all GraphQL requests have the same
path. We also don't have any way to determine how often retries are
required and if they succeed, since we currently only log the error when
all attempts fail.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/google/go-github/github"
 	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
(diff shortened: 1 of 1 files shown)


bluekeyes:
Based on the data from this, the next step to fix this missing date problem will probably be to try loading the commits directly if the pull request is missing the `pushedDate`. We only do this for forks right now but should fallback to it in all cases, perhaps removing or reducing the retires.

asvoboda (approved):

PR #114, opened 2019-08-14
only_has_contributors_in rule

alexodle:
Use case is when you have more than one set of users (or bots!) that can make pre-approved changes, all of which you want to allow to go through.

In our case, we have one bot that creates PRs, and another bot that ensures the PRs stay up to date with master. In some cases the timing works out where a PR ends up with commits from both bots before it's able to be merged. We want to allow these to push through.

### DIFF
--- a/policy/predicate/author.go
+++ b/policy/predicate/author.go
@@ -24,6 +24,41 @@ import (
 	"github.com/palantir/policy-bot/pull"
 )
 
+type OnlyHasAuthorsIn struct {
+	common.Actors `yaml:",inline"`
+}
+
+var _ Predicate = &OnlyHasAuthorsIn{}
+
+func (pred *OnlyHasAuthorsIn) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+	author := prctx.Author()
+
+	result, err := pred.IsActor(ctx, prctx, author)
+	if err != nil {
+		return false, "", errors.Wrap(err, "failed to validate author")
+	}
+	if !result {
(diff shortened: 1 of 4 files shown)


policy/predicate/author.go
```diff
@@ -24,6 +24,41 @@ import (
...
+		return false, fmt.Sprintf("The pull request author %q does not meet the required membership conditions", author), nil
+	}
+
+	commits, err := prctx.Commits()
+	if err != nil {
+		return false, "", errors.Wrap(err, "failed to get commits")
+	}
+
+	for _, c := range commits {
+		result, err := pred.IsActor(ctx, prctx, c.Author)
```
bluekeyes: I'd suggest copying and modifying the logic from the existing `HasContribtorIn` predicate, as it has a few advantages:

- It properly considers both the author and the committer of a commit as contributors
- It avoids running `IsActor` on the same user more than once (although the internal caching makes this less important)

bluekeyes (commented):
While I'm happy to support this, I think the behavior is closer to the existing `has_contributor_in`, so I'd like to call it `only_has_contributors_in` instead. The [contributor logic](https://github.com/palantir/policy-bot/blob/bbf4a281c4182fd8928d160ef37ebdb808fc45be/policy/predicate/author.go#L59) considers the author a contributor, so the end behavior is the same.

policy/predicate/author.go
alexodle: DONE!

bluekeyes (approved):

PR #117, opened 2019-08-22
Extract base-path from public URLs and use it as prefix for all service endpoint URLs

bgutjahr:
We want to run policy-bot docker container in a kubernetes pod.
In our enviroment, we tried to configure a public URL like this:
  https://out-host-name:443/policy-bot
Initially we configured the ingress (nginx) web proxy to map that URL to the
policy-bot kubernetes service (mapped to the policy-bot container's local port,
and to strip off the external-URLs /policy-bot base-path from the URL.

This approach didn't work, since the policy-bot container used it's own URLs (not the public URL)
for oauth redirects.

My solution is to extend policy-bot to extract the path from the public URL as use
is as a base-path for all handled and redirected URLs.

Then we could configure the ingress proxy to just pass the .../policy-bot URL unchanged to the
container, and everything worked smoothly.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/alexedwards/scs"
@@ -65,6 +66,7 @@ func New(c *Config) (*Server, error) {
 		return nil, errors.Wrap(err, "failed parse public URL")
 	}
 
+	basePath := strings.TrimSuffix(publicURL.Path, "/")
 	forceTLS := publicURL.Scheme == "https"
 
 	sessions := scs.NewCookieManager(c.Sessions.Key)
@@ -128,7 +130,13 @@ func New(c *Config) (*Server, error) {
 		return nil, errors.Wrap(err, "failed to load templates")
 	}
 
-	mux := base.Mux()
+	var mux *goji.Mux
+	if basePath == "" {
+		mux = base.Mux()
+	} else {
+		mux = goji.SubMux()
+		base.Mux().Handle(pat.New(basePath+"/*"), mux)
+	}
 
 	// webhook route
 	mux.Handle(pat.Post(githubapp.DefaultWebhookRoute), dispatcher)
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @bgutjahr! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

server/server.go
```diff
@@ -131,11 +135,11 @@ func New(c *Config) (*Server, error) {
 	mux := base.Mux()
 
 	// webhook route
-	mux.Handle(pat.Post(githubapp.DefaultWebhookRoute), dispatcher)
+	mux.Handle(pat.Post(basePath+githubapp.DefaultWebhookRoute), dispatcher)
```
bluekeyes: Instead of adding `basePath` to each route, I'd use a sub-mux instead, something like this (untested):

```golang
var mux *goiji.Mux
if basePath == "" {
  mux = base.Mux()
} else {
  mux = goji.SubMux()
  base.Mux().Handle(pat.New(basePath+"/*"), mux)
}
```

That way, all the route definitions can remain the same.

server/server.go
```diff
@@ -79,6 +81,8 @@ func New(c *Config) (*Server, error) {
 		return nil, errors.Wrap(err, "failed to initialize base server")
 	}
 
+	logger.Info().Msg("BasePath = " + basePath)
```
bluekeyes: Since we don't log any other configuration right now (although maybe we should), can you remove this?

bluekeyes (changes requested):
Thanks for implementing this! We hadn't considered deployments that use prefix paths since we usually use subdomains. I had a few suggestions on the implementation, but the concept is sound.

server/server.go
bgutjahr: I have made the suggested changes.
What I'm not fully sure about: is prepending the basePath in "handler.Static(basePath+"/static/", ..." (server.go, line 157) correct?

bluekeyes (approved):
Yes, the change to the static paths looks correct to me.

PR #122, opened 2019-10-07
Select all repository collaborators

asvoboda:


pull/github.go
```diff
@@ -356,7 +356,7 @@ func (ghc *GitHubContext) loadCollaboratorData() error {
 					Permission string  `graphql:"permission"`
 					Node       v4Actor `graphql:"node"`
 				}
-			} `graphql:"collaborators(first: 100, after: $collaboratorCursor, affiliation: DIRECT)"`
```
asvoboda, line 359: I need to rethink the flow "user is an org admin, but we don't want to ask them for review". Perhaps I can use https://developer.github.com/v4/object/permissionsource/, or run another check later

bluekeyes: Using `PermissonSource` and `PermissionGranter` feels like it would work and should avoid additional queries.

bluekeyes (approved):
Not sure if you want to merge this as is or try to fix the organization admins thing at the same time.

pull/github.go
asvoboda: I've opened #123 to hopefully deal with org owners as a separate feature. I think it would be more friendly to merge this, and iterate there for the "best" solution if we can't use the v4 api.

PR #123, opened 2019-10-07
Org owners can be excluded from auto-review

asvoboda:


bluekeyes:
Can you explain the background of this more? My understanding was that we only wanted to exclude organization admins if the rule requires admin approval _and_ they are not admins of the repository either directly or via a team.

As currently implemented, I believe this will remove org admins all the time, even if they are also admins on the repo through a team or directly. In our internal case, all the infrastructure team members are all organization admins, but removing them from review selection feels counter to the initial feature request.

asvoboda:
You're right about the goal to exclude org admins if they aren't admins of the repo directly or by team.

policy/common/result.go
```diff
@@ -37,13 +37,22 @@ func (s EvaluationStatus) String() string {
 	return "unknown"
 }
 
+type AdminScope string
+
+const (
```
asvoboda: Should we bail if an option provided doesn't match here?

README.md
```diff
@@ -203,6 +203,8 @@ options:
   request_review:
     # False by default
     enabled: true
+    # Include organization owners in automatic review. False by default.
+    org_owner: false
```
bluekeyes: Update this for the new `admin_scope` property

policy/common/result.go
bluekeyes: I think that's best, but I'm not sure how easy it is to implement. Maybe if `AdminScope` implemented the `MarshalText`/`UnmarshalText` methods, the YAML parsing would fail automatically? I don't remember if there's a separate validation step after parsing.

We'll also need to decide how the default (empty string) is handled.

policy/common/result.go
```diff
@@ -37,13 +37,22 @@ func (s EvaluationStatus) String() string {
...
+	User AdminScope = "user"
+	Team AdminScope = "team"
+	Org  AdminScope = "org"
```
bluekeyes: style: let's call these `ScopeUser` or `AdminScopeUser`... seeing `common.User` in another file could mean a lot of things.

policy/reviewer/reviewer.go
```diff
@@ -34,6 +34,9 @@ func findLeafChildren(result common.Result) []common.Result {
 		}
 	} else {
 		for _, c := range result.Children {
+			if c == nil {
```
bluekeyes: When are children `nil`? Is that a result of this change or was that always the case?

pull/context.go
```diff
@@ -36,6 +36,9 @@ type MembershipContext interface {
 
 	// OrganizationMembers returns the list of usernames in the given organization.
 	OrganizationMembers(org string) ([]string, error)
+
+	// OrganizationMembers returns the list of owner usernames in the given organization.
```
bluekeyes: `OrganizationOwners`

policy/reviewer/reviewer.go
```diff
@@ -124,31 +127,94 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
 			shoveIntoMap(allUsers, orgMembers)
 		}
 
-		allCollaboratorPermissions, err := prctx.RepositoryCollaborators()
+		allCollaboratorPermissions := make(map[string]string)
+		allCollaboratorAndPermissions, err := prctx.RepositoryCollaborators()
```
bluekeyes: The distinction between these variables is unclear... what does the extra `And` in the second one mean for the functionality?

policy/reviewer/reviewer.go
```diff
@@ -124,31 +127,94 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
...
 			shoveIntoMap(allUsers, repoCollaborators)
 		}
 
+		// When admins are selected for review, only collect the desired set of admins instead of
+		// everyone, which includes org admins
+		if !child.ReviewRequestRule.Admins {
+			// When not looking for admins, we want to check with everyone
+			allCollaboratorPermissions = allCollaboratorAndPermissions
+		} else {
+			// Determine what the scope of requested admins should be
```
bluekeyes: This feels significant enough to merit a separate helper function, maybe `selectAdmins` to match `selectTeamMembers` and `selectOrgMembers`.

policy/reviewer/reviewer.go
```diff
@@ -124,31 +127,94 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
...
+			switch child.ReviewRequestRule.AdminScope {
+			case common.User:
+				logger.Debug().Msg("Selecting admin users with direct collaboration rights")
+				directCollaborators, err := prctx.DirectRepositoryCollaborators()
+				if err != nil {
+					logger.Error().Err(err).Msgf("Unable to get list of direct collaborators on %s", prctx.RepositoryName())
```
bluekeyes: Seeing these error logs that continue on with execution is surprising (even if it works); I think this is another reason to make a helper function.

policy/reviewer/reviewer.go
```diff
@@ -124,31 +127,94 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
...
+					if perm == common.GithubAdminPermission {
+						repoAdmins = append(repoAdmins, user)
+					}
+				}
+				shoveIntoMap(allUsers, repoAdmins)
+				for _, a := range repoAdmins {
+					allCollaboratorPermissions[a] = common.GithubAdminPermission
+				}
+
+				break
```
bluekeyes: style: `break` is unnecessary in Go unless terminating the `case` early

policy/reviewer/reviewer.go
```diff
@@ -124,31 +127,94 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
...
+			case common.Org:
+				logger.Debug().Msg("Selecting admin users from the org")
+				orgOwners, err := prctx.OrganizationOwners(prctx.RepositoryOwner())
+				if err != nil {
+					logger.Error().Err(err).Msgf("Unable to get list of org owners for %s", prctx.RepositoryOwner())
+				}
+
+				for _, o := range orgOwners {
+					allUsers[o] = struct{}{}
```
bluekeyes: Why do the other cases use `shoveInToMap`, but this adds the users directly in a single loop? I think the single loop version (this one) is preferable, but would rather have consistency.

pull/github.go
```diff
@@ -289,6 +289,30 @@ func (ghc *GitHubContext) HasReveiwers() (bool, error) {
...
+		for _, t := range teams {
+			allTeams[t.GetName()] = t.GetPermission()
+		}
+		if resp.NextPage == 0 {
+			break
+		}
+		opt.Page = resp.NextPage
+	}
+
+	return allTeams, nil
```
bluekeyes, line 314: Should this be cached like other list output in the context?

pull/github.go
```diff
@@ -356,20 +380,36 @@ func (ghc *GitHubContext) loadCollaboratorData() error {
...
+
+	if direct {
+		if ghc.directCollaborators != nil {
+			return ghc.directCollaborators, nil
+		}
+	} else {
+		if ghc.collaborators != nil {
+			return ghc.collaborators, nil
+		}
+	}
```
bluekeyes: Instead of implementing this switching in this function, what if `loadCollaborators` only returned the map of collaborators and the callers were responsible for checking and initializing the cache?

That would also let you replace the `direct` parameter and take the `affiliation` value directly.

pull/github_membership.go
```diff
@@ -130,8 +130,9 @@ func (mc *GitHubMembershipContext) IsCollaborator(org, repo, user, desiredPerm s
 	return perm.GetPermission() == desiredPerm, nil
 }
 
-func (mc *GitHubMembershipContext) OrganizationMembers(org string) ([]string, error) {
+func (mc *GitHubMembershipContext) organizationMembers(org, perm string) ([]string, error) {
```
bluekeyes: style: `role` instead of `perm`

pull/pulltest/context.go
```diff
@@ -171,6 +190,14 @@ func (c *Context) OrganizationMembers(org string) ([]string, error) {
 	return inverted[org], nil
 }
 
+func (c *Context) OrganizationOwners(org string) ([]string, error) {
+	if c.OrgOwnerError != nil {
+		return nil, c.OrgOwnerError
+	}
+
+	return c.OrgOwners, nil
```
bluekeyes: style: simplify to `return c.OrgOwners, c.OrgOwnerError`

bluekeyes (commented):
When I originally proposed the `AdminScope` idea, I was thinking it would be cumulative, so that the `team` scope could also select users, and the `org` scope could also select users and teams. Do you think this is useful? Will people using admin rules be satisfied with picking a single type of admin for review requests? If you do want to implement this, I think it's relatively simple: invert the order of the `switch` and add `fallthrough` at the end of each `case`.

And unrelated to your code, it's really annoying how the GitHub API let's you filter on things like affiliation but provides no fields in the result object that would let client code do the same filtering. It would be a lot nicer to combine some of these query methods and return an object with the users grouped however we wanted.

policy/reviewer/reviewer.go
asvoboda: I was trying to come up with a better name here. The idea is that we want to only consider some collaborators, and not all, when checking which admins can be auto-assigned. I've tried to rename these to `allCollaborators` and `collaboratorsToConsider` to be a little more clear.

policy/reviewer/reviewer.go
asvoboda: I'll remove the `shoveIntoMap` calls

pull/github.go
asvoboda: I felt like it was a little more clunky if the clients need to worry about the caching, but I'm open to changing it if you feel like this is a worse outcome.

pull/github.go
asvoboda: We should get v3 caching by default, but I'm open to also caching it more explicitly here.

policy/reviewer/reviewer.go
asvoboda: I changed this because I was inserting a nil `Result` in the tests

policy/common/result.go
asvoboda: I think a sane default might be to assume `org` if unset? If you require admins to +1, but otherwise don't hint about what kind of admin, we should probably pick from anyone.

README.md
```diff
@@ -203,6 +203,11 @@ options:
   request_review:
     # False by default
     enabled: true
+    # If admins are required for review, they can be automatically selected.
+    # Admin of a repo is granted in three ways: direct user admin on the Repository, team admin on the repository,
+    # or as an Organization Owner.
+    # To reflect these permission sources, admin_scope can be set to (user|team|org)
+    admin_scope: "org"
```
iamdanfox: Are you sure you want to commit to this configurability? Seems fine to keep the settings internally, but I'd be tempted to just hard-code this setting so that only people who have admin perms via the `user` or `team` scope get requested (and org level admins never get requested).

asvoboda: Fair q. I think in general we prefer to have policy bot accept a variety of configurations, and for users to make their own decisions about how they want to achieve the best pull review ux. I don't think its super great to hard code assumptions about how to best review things into the bot.

This feature is actually pretty simple if I could figure out the right incantation for https://developer.github.com/v4/object/permissionsource/, but it appears to always return an empty list for github apps.

asvoboda: The more I think about it, the more I think you're right. We don't need to expose the complexity today, and we can always make it more configurable later. I'll push up a commit to reflect this.

policy/reviewer/reviewer.go
bluekeyes: Would it possible to not use `nil` children in the tests if this is not a case that occurs in the application? If children can legitimately be `nil` then I believe there are other places that need to be updated.

policy/reviewer/reviewer.go
```diff
@@ -97,6 +94,28 @@ func selectOrgMembers(prctx pull.Context, allOrgs []string, r *rand.Rand) ([]str
 	return orgMembers, nil
 }
 
+func selectAdmins(prctx pull.Context) ([]string, error) {
+	// Resolve all admin teams on the repo, and resolve their user membership
+	var adminUsers []string
+	teams, err := prctx.Teams()
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to get list of teams collaborators")
```
bluekeyes: style: almost all other `errors.Wrap` calls in this project use a message like `failed to ...` - can you update these to match?

policy/reviewer/reviewer.go
```diff
@@ -106,54 +125,70 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
 
 	for _, child := range pendingLeafNodes {
 		allUsers := make(map[string]struct{})
-		shoveIntoMap(allUsers, child.ReviewRequestRule.Users)
+
+		if len(child.ReviewRequestRule.Users) > 0 {
```
bluekeyes: style: this check is redundant, `range` on an empty loop won't do anything

pull/github.go
bluekeyes: Ok, we can try it with only the v3 caching for now.

policy/reviewer/reviewer.go
```diff
@@ -106,54 +125,70 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
...
-			var repoCollaborators []string
-			for user, perm := range allCollaboratorPermissions {
-				if perm == common.GithubWritePermission {
-					repoCollaborators = append(repoCollaborators, user)
-				}
+		// When admins are selected for review, only collect the desired set of admins instead of
+		// everyone, which includes org admins
+		if !child.ReviewRequestRule.Admins {
+			// When not looking for admins, we want to check with all possible collaborators
+			collaboratorsToConsider = allCollaborators
```
bluekeyes: These names are better now, but seeing it in the new limited context makes me think the distinction is not necessary at all, and is actually incorrect.

The `allUsers` set is what we actually use to select reviewers, while `collboratorsToConsider` is only used to figure out which elements of `allUsers` are collaborators. In the admin case, we add the team admins to `allUsers` but then we additionally limit the set of collaborators to admins as well. When an approval rule allows both admins and other conditions, I think this artificially limits the set of reviewers to only the admins. And when the rule only allows admins, the admin/not-admin filtering is performed by `selectAdmins`, so the additional filtering at this level is redundant.

Am I missing something?

bluekeyes (commented):
Two logic questions, two small style things, and then I think this is good to go.

policy/reviewer/reviewer.go
asvoboda: Fair point. I'll fix up the tests and remove this.

policy/reviewer/reviewer.go
asvoboda: I think you're right. I added the distinction previously when trying to find the set of admins and then remove certain users like org owners. Now that we're going with a purely additive approach, we can simplify and remove this bug.

policy/reviewer/reviewer.go
```diff
@@ -73,96 +73,130 @@ func selectRandomUsers(n int, users []string, r *rand.Rand) []string {
...
+		return nil, errors.Wrap(err, "failed to get list of teams collaborators")
 	}
-	return orgMembers, nil
+
+	for team, perm := range teams {
+		if perm == common.GithubAdminPermission {
+			fullTeamName := fmt.Sprintf("%s/%s", prctx.RepositoryOwner(), team)
+			admins, err := prctx.TeamMembers(fullTeamName)
+			if err != nil {
+				return nil, errors.Wrapf(err, "Unable to get list of members for %s", team)
```
bluekeyes, line 115: style: one missed `failed to`

policy/reviewer/reviewer.go
```diff
@@ -73,96 +73,130 @@ func selectRandomUsers(n int, users []string, r *rand.Rand) []string {
...
+		if child.ReviewRequestRule.Admins {
+			logger.Debug().Msg("Selecting admins for review")
+			admins, err := selectAdmins(prctx)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to select admins")
+			}
+
+			for _, admin := range admins {
+				allUsers[admin] = struct{}{}
+				collaboratorsToConsider[admin] = common.GithubAdminPermission
```
bluekeyes: I thought these users would already be in `collaboratorsToConsider` as part of listing all collaborators. Is that not true?

bluekeyes (approved):

policy/reviewer/reviewer.go
asvoboda: you're correct, I missed this in the last commit.

bluekeyes (approved):

iamdanfox:
Awesome :) lmk when we can try this out on some of our oss/internal repos 🎉🎉

PR #126, opened 2019-10-14
Double check if there are existing reviewers

asvoboda:
For complex approvals, policy bot may take a few seconds, and a human might request review manually. Run another check just before we assign users so we don't clash with a user's decision.

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -214,7 +214,13 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, pe
 				logger.Warn().Err(err).Msg("Unable to select random request reviewers")
 			}
 
-			if len(requestedUsers) > 0 {
+			// check again if someone assigned a reviewer while we were calculating users to request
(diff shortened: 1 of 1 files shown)


server/handler/base.go
```diff
@@ -214,7 +214,13 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, pe
 				logger.Warn().Err(err).Msg("Unable to select random request reviewers")
 			}
 
-			if len(requestedUsers) > 0 {
+			// check again if someone assigned a reviewer while we were calculating users to request
+			hasReviewersAfter, err := prctx.HasReveiwers()
+			if err != nil {
+				logger.Warn().Err(err).Msg("Unable to list request reviewers a second time")
```
bluekeyes: I believe `hasReviewerAfter` will be false in this case? If so, let's update this message to indicate the consequences of that, something like `Unable to double-check existing reviewers, assuming original state is still valid.`

bluekeyes (approved):

PR #127, opened 2019-10-14
Don't request reviewers if there are existing reviews

asvoboda:
Github removes the set of users requested for review once they finish adding a review, which can cause the `HasReviewers()` flow to feel pretty klunky.

To fix this, we can also check if there are any reviews left on the PR, and reasonably assume that if a user has already left some reviews that the bot can safely move on.

### DIFF
--- a/pull/context.go
+++ b/pull/context.go
@@ -95,7 +95,10 @@ type Context interface {
 	Teams() (map[string]string, error)
 
 	// HasReviewers returns true if the Pull Request has reviewers
-	HasReveiwers() (bool, error)
+	HasReviewers() (bool, error)
+
+	// HasReviews returns true if the Pull Request has reviews
+	HasReviews() (bool, error)
 }
 
 type FileStatus int
--- a/pull/context.go
+++ b/pull/context.go
@@ -94,11 +94,8 @@ type Context interface {
 	// their respective permission on a repo.
 	Teams() (map[string]string, error)
 
-	// HasReviewers returns true if the Pull Request has reviewers
+	// HasReviewers returns true if the Pull Request has reviewers or any reviews
 	HasReviewers() (bool, error)
-
-	// HasReviews returns true if the Pull Request has reviews
-	HasReviews() (bool, error)
 }
 
 type FileStatus int
--- a/pull/github.go
+++ b/pull/github.go
@@ -277,7 +277,7 @@ func (ghc *GitHubContext) RepositoryCollaborators() (map[string]string, error) {
 	return ghc.collaborators, nil
 }
 
-func (ghc *GitHubContext) HasReveiwers() (bool, error) {
+func (ghc *GitHubContext) HasReviewers() (bool, error) {
 	// Intentionally kept to a small result size, since we just want to determine if there are existing reviewers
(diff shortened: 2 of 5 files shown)


pull/github.go
```diff
@@ -277,7 +277,7 @@ func (ghc *GitHubContext) RepositoryCollaborators() (map[string]string, error) {
 	return ghc.collaborators, nil
 }
 
-func (ghc *GitHubContext) HasReveiwers() (bool, error) {
+func (ghc *GitHubContext) HasReviewers() (bool, error) {
```
asvoboda, line 280: Also fixed a type while I was here

server/handler/base.go
```diff
@@ -202,20 +202,25 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, pe
...
-		hasReviewers, err := prctx.HasReveiwers()
+		hasReviewers, err := prctx.HasReviewers()
 		if err != nil {
 			logger.Warn().Err(err).Msg("Unable to list request reviewers")
 		}
 
-		if !hasReviewers {
+		hasReviews, err := prctx.HasReviews()
+		if err != nil {
+			logger.Warn().Err(err).Msg("Unable to list reviews")
```
asvoboda: Maybe this should return an error so that we return and stop processing?

pull/github.go
```diff
@@ -290,6 +290,19 @@ func (ghc *GitHubContext) HasReveiwers() (bool, error) {
...
+	// Intentionally kept to a small result size, since we just want to determine if there are existing reviews
+	subsetCurrentReviews, _, err := ghc.client.PullRequests.ListReviews(ghc.ctx, ghc.owner, ghc.repo, ghc.number, &github.ListOptions{
+		Page:    0,
+		PerPage: 1,
+	})
+	if err != nil {
+		return false, errors.Wrap(err, "Unable to list request reviewers")
+	}
+
+	return len(subsetCurrentReviews) > 0, nil
```
bluekeyes: I recommend implementing this by calling `ghc.Reviews()` instead of making a new API call. In almost all cases, that data will already be loaded by the time `HasReviews()` is called as a result of the normal evaluation.

server/handler/base.go
bluekeyes: That seems better to me, since the end result is the same... logging an error doesn't allow any best effort attempt later on.

pull/context.go
```diff
@@ -95,7 +95,10 @@ type Context interface {
 	Teams() (map[string]string, error)
 
 	// HasReviewers returns true if the Pull Request has reviewers
-	HasReveiwers() (bool, error)
+	HasReviewers() (bool, error)
+
+	// HasReviews returns true if the Pull Request has reviews
+	HasReviews() (bool, error)
```
bluekeyes: Does this need to be a separate function, or should it be combined with `HasReviewers`? It seems completely reasonable to me that `HasReviewers` would be true if there are requested reviewers or if there are actual reviews.

bluekeyes (commented):
A more robust fix might be to only take action if the `pull_request` event is `opened` or `ready_for_review` and otherwise process the event without taking action. That should further help to avoid race conditions between server instances if there are multiple events in quick succession (say pushing and editing, or the auto-edit from something like Reviewable.)

pull/context.go
asvoboda: Good point. I'll move this into `HasReviewers`.

bluekeyes (approved):

PR #130, opened 2019-10-22
Add option for requesting all possible reviewers

### LINKED ISSUE #129: FR: include ability to add all authorized reviewers to pull request
We work in a monorepo with very disparate teams. It would be nice if every authorized reviewer was added to a PR instead of picking at random until the `requires.count` limit is met. In particular, we'd like to be able to automatically request reviews from entire teams that own those portions of the codebase.

asvoboda:
Resolves #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,13 +199,15 @@ options:
   # Automatically request reviewers when a Pull Request is opened
   # if this rule is pending, there are no assigned reviewers, and if the
   # Pull Request is not in Draft.
-  # Reviewers are selected randomly based on the set of requirements for this rule.
+  # Reviewers are selected based on the set of requirements for this rule
+  # and reviewers can be augmented using the mode option.
   request_review:
     # False by default
     enabled: true
     # mode modifies how users are selected. `all-users` will request all users
     # who are able to approve the pending rule. `random-users` selects a small
     # set of random users based on the required count of approvals.
+    # defaults to 'random-users'
     mode: all-users|random-users
 
   # "methods" defines how users may express approval. The defaults are below.
@@ -375,7 +377,7 @@ when Pull Requests are opened by setting the `request_review` option.
 
 The `mode` enum modifies how users are selected. There are currently two
 supported options:
- * `all-users` to requests all users who can approve
+ * `all-users` to request all users who can approve
  * `random-users` to randomly select the number of users that are required
 
 ```yaml
--- a/README.md
+++ b/README.md
@@ -198,11 +198,13 @@ options:
 
   # Automatically request reviewers when a Pull Request is opened
   # if this rule is pending, there are no assigned reviewers, and if the
-  # Pull Request is not in Draft. Reviewers are selected randomly
-  # based on the set of requirements for this rule.
+  # Pull Request is not in Draft. If add_everyone is false,
+  # Reviewers are selected randomly based on the set of requirements for this rule.
+  # If add_everyone is true, all possible reviewers are requested for review.
   request_review:
     # False by default
     enabled: true
+    add_everyone: false
 
   # "methods" defines how users may express approval. The defaults are below.
   methods:
@@ -374,11 +376,15 @@ options:
   request_review:
     # False by default
     enabled: true
+    add_everyone: false
 ```
 
-A number of reviewers will be randomly requested based on the `requires` rules
-so that all rules will be satisfied once the set of requested reviewers complete
-and approve the Pull Request.
+If `add_everyone` is false, a number of reviewers will be randomly requested
+based on the `requires` rules so that all rules will be satisfied once the
+set of requested reviewers complete and approve the Pull Request.
+
+If `add_everone` is true, all possible reviewers that match the pending
+criteria will be added as reviewers.
 
 The set of requested reviewers will not include the author of the Pull Request or
 users who are not collaborators on the repository.
--- a/README.md
+++ b/README.md
@@ -198,13 +198,15 @@ options:
 
   # Automatically request reviewers when a Pull Request is opened
   # if this rule is pending, there are no assigned reviewers, and if the
-  # Pull Request is not in Draft. If add_everyone is false,
+  # Pull Request is not in Draft.
   # Reviewers are selected randomly based on the set of requirements for this rule.
-  # If add_everyone is true, all possible reviewers are requested for review.
   request_review:
     # False by default
     enabled: true
-    add_everyone: false
+    # mode modifies how users are selected. `all-users` will request all users
+    # who are able to approve the pending rule. `random-users` selects a small
+    # set of random users based on the required count of approvals.
+    mode: all-users|random-users
 
   # "methods" defines how users may express approval. The defaults are below.
   methods:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -198,11 +198,13 @@ options:
...
   # if this rule is pending, there are no assigned reviewers, and if the
-  # Pull Request is not in Draft. Reviewers are selected randomly
-  # based on the set of requirements for this rule.
+  # Pull Request is not in Draft. If add_everyone is false,
+  # Reviewers are selected randomly based on the set of requirements for this rule.
+  # If add_everyone is true, all possible reviewers are requested for review.
   request_review:
     # False by default
     enabled: true
+    add_everyone: false
```
asvoboda: I'm not sold on the name of this, but I'm not sure what else to call it.

jamestoyer: Perhaps it could be `everyone` or `all`. I think it should also default to true instead of `false` to me it feels like it the most expected behaviour if I was turning this feature on

bluekeyes: Maybe this should be an enum instead?

```
mode: random-users|all-users|teams
```

Where `random-users` is the current behavior, `all-users` is the behavior added in this PR, and future extensions would add more options, like `teams` to request review from teams rather than adding their individual members.

Also, minor style thing, but can you move the documentation for how this options controls behavior to be above the option?

policy/reviewer/reviewer.go
```diff
@@ -189,9 +189,14 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
 		}
 
 		if len(possibleReviewers) > 0 {
-			logger.Debug().Msgf("Found %d total candidates for review after removing author and non-collaborators; randomly selecting %d", len(possibleReviewers), child.ReviewRequestRule.RequiredCount)
-			randomSelection := selectRandomUsers(child.ReviewRequestRule.RequiredCount, possibleReviewers, r)
-			usersToRequest = append(usersToRequest, randomSelection...)
+			if child.ReviewRequestRule.AddEveryone {
```
bluekeyes: I can't comment on the line but `FindRandomRequesters` is no longer an accurate name for this function.

If we do go with the `mode` option approach, a nice refactor might be to wrap these in strategy functions so that this becomes a `switch` on the mode that calls the correct strategy for the mode.

policy/reviewer/reviewer.go
```diff
@@ -189,9 +189,16 @@ func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common
...
 		if len(possibleReviewers) > 0 {
-			logger.Debug().Msgf("Found %d total candidates for review after removing author and non-collaborators; randomly selecting %d", len(possibleReviewers), child.ReviewRequestRule.RequiredCount)
-			randomSelection := selectRandomUsers(child.ReviewRequestRule.RequiredCount, possibleReviewers, r)
-			usersToRequest = append(usersToRequest, randomSelection...)
+			switch child.ReviewRequestRule.Mode {
+			case common.RequestModeAllUsers:
+				logger.Debug().Msgf("Found %d total reviewers after removing author and non-collaborators; requesting all", len(possibleReviewers))
+				usersToRequest = append(usersToRequest, possibleReviewers...)
+			case common.RequestModeRandomUsers:
+			default:
```
asvoboda: Should this be the default to conform with current behaviour? Should we bail earlier?

README.md
asvoboda: I like the idea of an enum here!

policy/reviewer/reviewer.go
asvoboda: We you thinking of a function to wrap `SelectReviewers`?

README.md
```diff
@@ -369,17 +373,18 @@ limitations. Please file an issue if this functionality is important to you.
 `policy-bot` can automatically request reviewers for all pending rules
 when Pull Requests are opened by setting the `request_review` option.
 
+The `mode` enum modifies how users are selected. There are currently two
+supported options:
+ * `all-users` to requests all users who can approve
```
swanysimon: "to request"

swanysimon (commented):
does "all-users" include teams or admins (if `admins: true`) at the moment?

README.md
```diff
@@ -198,11 +198,15 @@ options:
 
   # Automatically request reviewers when a Pull Request is opened
   # if this rule is pending, there are no assigned reviewers, and if the
-  # Pull Request is not in Draft. Reviewers are selected randomly
-  # based on the set of requirements for this rule.
+  # Pull Request is not in Draft.
+  # Reviewers are selected randomly based on the set of requirements for this rule.
```
bluekeyes: This actually depends on the value of `mode` now

policy/reviewer/reviewer.go
bluekeyes: If this was the first implementation, I'd pick `all-users` as the default. Given we already released with random selection, let's leave that as the default to preserve compatibility.

Note that cases do not fallthrough by default in Go - use the `fallthrough` keyword. That said, it might be better to fill in a default value earlier so that the `default` case here can be for invalid modes. Take a look at `GetMethods` on the `approve.Options` type for one way to do this. `FillDefaults` on the server configuration is another.

policy/reviewer/reviewer.go
bluekeyes: I was thinking of something like this:

```golang
switch mode {
case common.RequestModeRandomUsers:
  usersToRequest = selectRandomly(users)
case common.RequestModeAllUsers:
  usersToRequest = selectAll(users)
}
```

The benefit here is minimal since the selection logic is simple. It's more valuable as complexity increases and you might be returning teams _and_ users or doing other logic.

bluekeyes:
@swanysimon yes, `all-users` takes the list of all users that satisfy a particular rule (including admins, members of teams, etc.) and requests them for review. This is the same list of users that was selected from randomly in the initial implementation.

policy/reviewer/reviewer.go
asvoboda: We will probably need to change up some of the selection if we want to support team requests, since the gh api expects the team names to be passed as a separate parameter: https://developer.github.com/v3/pulls/review_requests/#input

Maybe we can note this for now as work that should definitely be done later, but only once we've got a good usecase for it?

bluekeyes (approved):
Looks good, let's try it out and then consider teams mode later on.

README.md
TheBeruriahIncident: As a bystander, I can't imagine a case when you'd want the current behavior of picking a random user. The point of being an owner is to know when code changes

jamestoyer: @akroy being an owner of a repo in GitHub brings the ability to do admin things to the repo. There are times when the repo admin (i.e. owner) is there to ensure settings are enforced but they don't actually participate in the code being generated. An example of this might be that there are no human owners on the repo as the admin duties are delegated to bots which maintain the settings. This is certainly the case on some of the repos I work on.

TheBeruriahIncident: Sorry to clarify, I'm talking about being a code owner, not a repo owner. (the native GitHub feature that we're replacing with Policy-Bot) I think the goal there is see the code changes. If you didn't want to see the code changes, and the work was done by bots, would you still be a code owner/policy member?

At the very least, it seems like the default repo's code owners want to see changes to the code they own?

PR #131, opened 2019-10-31
Add better leaf node debugging logs

asvoboda:


policy/reviewer/reviewer.go
```diff
@@ -122,18 +126,22 @@ func selectAdmins(prctx pull.Context) ([]string, error) {
...
 
 func FindRandomRequesters(ctx context.Context, prctx pull.Context, result common.Result, r *rand.Rand) ([]string, error) {
-	logger := zerolog.Ctx(ctx)
+	logger := *zerolog.Ctx(ctx)
 	var usersToRequest []string
 	pendingLeafNodes := findLeafChildren(result)
 
+	logger.Debug().Msgf("Found %d pending leaf nodes for review selection", len(pendingLeafNodes))
 	for _, child := range pendingLeafNodes {
+		logger = logger.With().Str(LogKeyLeafNode, child.Name).Logger()
```
bluekeyes: This will keep stacking repeated fields on the same logger for each child, since `zerolog` does not deduplicate fields. It's probably better to get this from the context at the start of each loop.

bluekeyes (approved):

PR #139, opened 2019-11-15
Add has_successful_status predicate

### LINKED ISSUE #102: [Enhancement] Approve if other status check has passed
Github status check enforcement is an "all must pass" system. This poses a problem for our workflow since we only trigger builds if that portion of the code has changed - we would prefer to use a "one or the other" style system as a minimum, and preferably a "if this code section changed, require this build" setup down the line

I would like to setup an approval like so:

```yaml
policy:
  approval:
    - and:
      - code has changed
      - or:
        - the API build has passed
        - the Client build has passed
      - the deploy has passed

approval_rules:
    - name: the API build has passed
      if:
        status_check_passed: My.Company.Project.API - CI
```

policy-bot already has read access for Status checks, but it does not appear to be currently possible to setup the above approval step.

pilbot:
Fixes #102

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,10 @@ if:
     additions: "> 100"
     deletions: "> 100"
     total: "> 200"
+  
+  # "has_successful_status" is satisfied if the status check specified is marked
+  # successful on the head commit of the pull request.
+  has_successful_status: "status-name"
 
 # "options" specifies a set of restrictions on approvals. If the block does not
 # exist, the default values are used.
--- a/README.md
+++ b/README.md
@@ -174,9 +174,12 @@ if:
     deletions: "> 100"
     total: "> 200"
   
-  # "has_successful_status" is satisfied if the status check specified is marked
-  # successful on the head commit of the pull request.
-  has_successful_status: "status-name"
+  # "has_successful_status" is satisfied if the status checks that are specified
+  # are marked successful on the head commit of the pull request.
+  has_successful_status:
+    - "status-name-1"
+    - "status-name-2"
+    - "status-name-3"
 
 # "options" specifies a set of restrictions on approvals. If the block does not
 # exist, the default values are used.
--- a/policy/predicate/status.go
+++ b/policy/predicate/status.go
@@ -1,0 +1,45 @@
+// Copyright 2018 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package predicate
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/palantir/policy-bot/pull"
+)
+
+type HasSuccessfulStatus string
+
+var _ Predicate = HasSuccessfulStatus("")
+
+func (pred HasSuccessfulStatus) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+	statuses, err := prctx.LatestStatuses()
+	if err != nil {
+		return false, "", errors.Wrap(err, "failed to list commit statuses")
+	}
+
+	status, ok := statuses[string(pred)]
+	if !ok {
+		return false, fmt.Sprintf("%q has not been executed on this commit", pred), nil
+	}
+	if status == "success" {
+		return true, "", nil
+	}
+
+	return false, fmt.Sprintf("%q has not passed", pred), nil
+}
--- a/policy/predicate/status.go
+++ b/policy/predicate/status.go
@@ -17,15 +17,16 @@ package predicate
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 
 	"github.com/palantir/policy-bot/pull"
 )
 
-type HasSuccessfulStatus string
+type HasSuccessfulStatus []string
 
-var _ Predicate = HasSuccessfulStatus("")
+var _ Predicate = HasSuccessfulStatus([]string{})
 
 func (pred HasSuccessfulStatus) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
 	statuses, err := prctx.LatestStatuses()
@@ -33,13 +34,30 @@ func (pred HasSuccessfulStatus) Evaluate(ctx context.Context, prctx pull.Context
 		return false, "", errors.Wrap(err, "failed to list commit statuses")
 	}
 
-	status, ok := statuses[string(pred)]
-	if !ok {
-		return false, fmt.Sprintf("%q has not been executed on this commit", pred), nil
-	}
-	if status == "success" {
-		return true, "", nil
+	missingResults := make([]string, 0)
+	failingStatuses := make([]string, 0)
+	for _, status := range pred {
+		result, ok := statuses[status]
+		if !ok {
+			missingResults = append(missingResults, status)
+		}
+		if result != "success" {
+			failingStatuses = append(failingStatuses, status)
+		}
 	}
 
-	return false, fmt.Sprintf("%q has not passed", pred), nil
+	if len(missingResults) == 1 {
+		return false, fmt.Sprintf("%q has not been executed on this commit", missingResults[0]), nil
+	} else if len(missingResults) > 1 {
+		missingResultsStr := strings.Join(missingResults, ",")
+		return false, fmt.Sprintf("%q have not been executed on this commit", missingResultsStr), nil
+	}
+
+	if len(failingStatuses) == 1 {
+		return false, fmt.Sprintf("%q has not passed", failingStatuses[0]), nil
+	} else if len(failingStatuses) > 1 {
+		failingStatusesStr := strings.Join(failingStatuses, ",")
+		return false, fmt.Sprintf("The statuses %q have not passed", failingStatusesStr), nil
+	}
+	return true, "", nil
 }
--- a/pull/github.go
+++ b/pull/github.go
@@ -125,6 +125,7 @@ type GitHubContext struct {
 	collaborators map[string]string
 	teamIDs       map[string]int64
 	membership    map[string]bool
+	statuses      map[string]string
 }
 
 // NewGitHubContext creates a new pull.Context that makes GitHub requests to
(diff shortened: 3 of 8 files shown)


README.md
```diff
@@ -173,6 +173,10 @@ if:
     additions: "> 100"
     deletions: "> 100"
     total: "> 200"
+  
+  # "has_successful_status" is satisfied if the status check specified is marked
+  # successful on the head commit of the pull request.
+  has_successful_status: "status-name"
```
bluekeyes: I suspect it will be valuable and not much extra work to accept a list of status here, where all statuses must be successful to satisfy the rule.

pull/github.go
```diff
@@ -319,6 +319,25 @@ func (ghc *GitHubContext) Teams() (map[string]string, error) {
...
+func (ghc *GitHubContext) LatestStatuses() (map[string]string, error) {
+	latestStatuses := make(map[string]string)
+
+	statuses, _, err := ghc.client.Repositories.ListStatuses(ghc.ctx, ghc.owner, ghc.repo, ghc.HeadSHA(), &github.ListOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get statuses for %s", ghc.HeadSHA())
+	}
+
+	// Statuses are returned in reverse chronological order (i.e. latest item first), only keep the latest status
+	for _, s := range statuses {
```
bluekeyes: I think you could simplify this by using the the [combined status](https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref) API, which only returns the latest value for each context. That should avoid loading extra data.

pull/github.go
```diff
@@ -319,6 +319,25 @@ func (ghc *GitHubContext) Teams() (map[string]string, error) {
 	return allTeams, nil
 }
 
+func (ghc *GitHubContext) LatestStatuses() (map[string]string, error) {
+	latestStatuses := make(map[string]string)
+
+	statuses, _, err := ghc.client.Repositories.ListStatuses(ghc.ctx, ghc.owner, ghc.repo, ghc.HeadSHA(), &github.ListOptions{})
```
bluekeyes: It's important to [paginate](https://godoc.org/github.com/google/go-github/github#hdr-Pagination) these requests to handle cases where there are more statuses than can fit on a single page. It's also usually a good idea to set a larger default page size - most of the other methods here use 100, the maximum size.

pull/github.go
```diff
@@ -319,6 +319,25 @@ func (ghc *GitHubContext) Teams() (map[string]string, error) {
...
+
+	// Statuses are returned in reverse chronological order (i.e. latest item first), only keep the latest status
+	for _, s := range statuses {
+		_, ok := latestStatuses[s.GetContext()]
+		if !ok {
+			latestStatuses[s.GetContext()] = s.GetState()
+		}
+	}
+
+	return latestStatuses, nil
```
bluekeyes: Consider caching this value in the context (see other methods for reference) so that statuses are only loaded once even if multiple rules use a status predicate.

server/handler/status.go
```diff
@@ -50,44 +61,68 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
...
-		// replace a forged context with a failure, not post a general status
-		// if multiple contexts are forged, we will handle multiple events
-		status := &github.RepoStatus{
-			Context:     event.Context,
-			State:       github.String("failure"),
-			Description: &desc,
-		}
-
-		_, _, err := client.Repositories.CreateStatus(ctx, ownerName, repoName, commitSHA, status)
+	client, err := h.NewInstallationClient(installationID)
```
bluekeyes, line 108: Before doing any work, we should check the status of the context. Since we only care about successful status checks, we can skip a lot of work by ignoring everything else.

server/handler/status.go
```diff
@@ -50,44 +61,68 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
...
+	if err != nil {
 		return err
 	}
 
-	return nil
+	ctx, logger := githubapp.PrepareRepoContext(ctx, installationID, repo)
+
+	// Get all PRs for which the event commit references the head of a PR
+	prs, _, err := client.PullRequests.List(ctx, ownerName, repoName, &github.PullRequestListOptions{Head: commitSHA})
```
bluekeyes: In practice, I don't think paging is necessary here, but I'd like to include a comment explaining why along with an explicit page size. Something like "In practice, there should only be one or two open PRs for a given commit. In exceptional cases, if there are more than N open PRs, only process the most recent N."

We should also include the `state` filter here, to only list open PRs - no point updating something that has been closed.

server/handler/status.go
```diff
@@ -50,44 +61,68 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
...
+	if err != nil {
+		return err
```
bluekeyes: `errrors.Wrapf(err, "failed to list pull requests for SHA %s", commitSHA)`

server/handler/status.go
```diff
@@ -50,44 +61,68 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
...
+
+	for _, pr := range prs {
+		err = h.Evaluate(ctx, installationID, true, pull.Locator{
+			Owner:  ownerName,
+			Repo:   repoName,
+			Number: pr.GetNumber(),
+			Value:  pr,
+		})
+	}
+	return err
```
bluekeyes: As implemented, this will only return the the last error and hide any errors returned by earlier evaluations. One way to solve this is to log the intermediate errors and count the number that occurred. Then, return a new error at the end saying something like `failed to evaluate N pull requests`.

bluekeyes (changes requested):
This looks like a good start. I had several comments about some edge cases and ways to improve the efficiency. I also think we should consider taking a list of statuses instead of a single one.

pull/github.go
pilbot: 🤦‍♂ I read that API doc and thought it was all statuses rolled up for a particular commit... Will change.

server/handler/status.go
pilbot: I've put this check in the caller as think that makes it more obvious.

pull/github.go
```diff
@@ -319,6 +320,30 @@ func (ghc *GitHubContext) Teams() (map[string]string, error) {
 	return allTeams, nil
 }
 
+func (ghc *GitHubContext) LatestStatuses() (map[string]string, error) {
+	if ghc.statuses == nil {
+		opt := &github.ListOptions{
+			PerPage: 100,
+		}
+		// get all pages of results
+		ghc.statuses = make(map[string]string)
```
bluekeyes, line 329: It's best to do this after the `for` loop so that an empty map is not cached in the event of an error.

policy/predicate/status.go
```diff
@@ -0,0 +1,63 @@
...
+var _ Predicate = HasSuccessfulStatus([]string{})
+
+func (pred HasSuccessfulStatus) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+	statuses, err := prctx.LatestStatuses()
+	if err != nil {
+		return false, "", errors.Wrap(err, "failed to list commit statuses")
+	}
+
+	missingResults := make([]string, 0)
+	failingStatuses := make([]string, 0)
```
bluekeyes, line 38: style: `var missingResults []string` for both of these

policy/predicate/status.go
```diff
@@ -0,0 +1,63 @@
...
+			failingStatuses = append(failingStatuses, status)
+		}
+	}
+
+	if len(missingResults) == 1 {
+		return false, fmt.Sprintf("%q has not been executed on this commit", missingResults[0]), nil
+	} else if len(missingResults) > 1 {
+		missingResultsStr := strings.Join(missingResults, ",")
+		return false, fmt.Sprintf("%q have not been executed on this commit", missingResultsStr), nil
+	}
```
bluekeyes, line 54: I think we could simplify this messaging to something like:

```golang
if len(missingResults) > 0 {
  return false, "One or more status are missing: " + strings.Join(missingResults, ", ")
}
```

And similarly for the failing case: `One or more status have not passed: ...`

server/handler/status.go
```diff
@@ -50,44 +64,83 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
...
+
+	// In practice, there should only be one or two open PRs for a given commit. In exceptional cases, if there are
+	// more than 100 open PRs, only process the most recent 100.
+	prs, _, err := client.PullRequests.List(
+		ctx,
+		ownerName,
+		repoName,
+		&github.PullRequestListOptions{
+			Head:        commitSHA,
+			ListOptions: github.ListOptions{PerPage: 100},
```
bluekeyes, line 123: Unfortunately, I don't think this works. The `head` filter is documented as

> Filter pulls by head user or head organization and branch name in the format of `user:ref-name` or `organization:ref-name`.

not by the commit SHA. Assuming that's correct and it doesn't also accept SHA values, I think that leaves two options:

1. List _all_ open pull requests (with paging) until we find ones containing the SHA
2. Use the GraphQL API instead so we can use the `associatedPullRequests` connection on the [`commit`](https://developer.github.com/v4/object/commit/) type. 

I think my preference is for the GraphQL approach, since it should only require a single API call, although it doesn't support filtering by state, which we'll have to do client-side.

server/handler/status.go
```diff
@@ -50,44 +64,83 @@ func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, paylo
...
+	evaluationFailures := 0
+	for _, pr := range prs {
+		err = h.Evaluate(ctx, installationID, true, pull.Locator{
+			Owner:  ownerName,
+			Repo:   repoName,
+			Number: pr.GetNumber(),
+			Value:  pr,
+		})
+		if err != nil {
+			evaluationFailures++
```
bluekeyes, line 139: It's important to log the error here as well so that we have it for debugging.

server/handler/status.go
```diff
@@ -31,17 +33,29 @@ type Status struct {
...
-// Handle pull_request
+// Handle status
 // https://developer.github.com/v3/activity/events/types/#statusevent
 func (h *Status) Handle(ctx context.Context, eventType, deliveryID string, payload []byte) error {
 	var event github.StatusEvent
 	if err := json.Unmarshal(payload, &event); err != nil {
 		return errors.Wrap(err, "failed to parse status event payload")
 	}
 
+	if !strings.HasPrefix(event.GetContext(), h.PullOpts.StatusCheckContext) {
```
bluekeyes, line 44: Given the introduction of `processOthers` and `processOwn`, I think it would be better to invert this condition: if the context has the prefix, call `processOwn` and return.

bluekeyes (changes requested):

server/handler/status.go
sjrand: There's a `ListPullRequestsWithCommit`, which seems like what we want, so switched to use that

bluekeyes:
Closing in favor of #141.

PR #141, opened 2019-12-05
Add has_successful_status predicate

### LINKED ISSUE #102: [Enhancement] Approve if other status check has passed

sjrand:
Continues the work started by @pilbot in https://github.com/palantir/policy-bot/pull/139. I don't have permission to push to that branch, thus the new one.

Fixes #102.

### DIFF
--- a/server/handler/status.go
+++ b/server/handler/status.go
@@ -121,7 +121,6 @@ func (h *Status) processOthers(ctx context.Context, event github.StatusEvent) er
 		repoName,
 		commitSHA,
 		&github.PullRequestListOptions{
-			Head:        commitSHA,
 			ListOptions: github.ListOptions{PerPage: 100},
 		})
 	if err != nil {
--- a/server/handler/status.go
+++ b/server/handler/status.go
@@ -138,7 +138,7 @@ func (h *Status) processOthers(ctx context.Context, event github.StatusEvent) er
 		})
 		if err != nil {
 			evaluationFailures++
-			logger.Error().Err(err).Msgf("failed to evaluate pull request '%d' for SHA '%s'", pr.GetNumber(),
+			logger.Error().Err(err).Msgf("Failed to evaluate pull request '%d' for SHA '%s'", pr.GetNumber(),
 				commitSHA)
 
 		}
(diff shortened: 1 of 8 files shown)


server/handler/status.go
```diff
@@ -114,12 +115,12 @@ func (h *Status) processOthers(ctx context.Context, event github.StatusEvent) er
 
 	// In practice, there should only be one or two open PRs for a given commit. In exceptional cases, if there are
 	// more than 100 open PRs, only process the most recent 100.
-	prs, _, err := client.PullRequests.List(
+	prs, _, err := client.PullRequests.ListPullRequestsWithCommit(
```
bluekeyes, line 118: Nice, I didn't know this endpoint existed in the v3 API. Per the [docs](https://developer.github.com/v3/repos/commits/#list-pull-requests-associated-with-commit), it looks like it returns both closed and open PRs:

> Lists all pull requests containing the provided commit SHA, which can be from any point in the commit history. The results will include open and closed pull requests.

I think the 100 PR limit is _probably_ still safe in practice (although the comment is inaccurate now), but we do need to filter the results to only evaluate open PRs in the loop on L132.

server/handler/status.go
```diff
@@ -137,6 +138,9 @@ func (h *Status) processOthers(ctx context.Context, event github.StatusEvent) er
 		})
 		if err != nil {
 			evaluationFailures++
+			logger.Error().Err(err).Msgf("failed to evaluate pull request '%d' for SHA '%s'", pr.GetNumber(),
```
bluekeyes: style: `Failed to ...`. In this codebase, log message usually start with a captial letter, errors with a lowercase letter.

bluekeyes (changes requested):
This is almost ready, but I think there's one more issue with the PR listing. Once that's resolved it should be good to merge.

bluekeyes (approved):

PR #142, opened 2019-12-05
Add policy yaml validation endpoint

### LINKED ISSUE #10: Verify the validity of policy.yml files before PRs merge
Context:

- currently it is possible to open a PR to a repo containing an invalid policy.yml file, because there is no pre-merge validation on this

Desired:

- a red CI light (a status) from policy-bot preventing me from merging invalid policy.yml files.

asvoboda:
One way to resolve #10. 

Another approach could also take in the org/repo and current PR to run a more complete validation of all settings, similar to the details endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,27 @@ disapproval:
     teams: ["org1/team1", "org2/team2"]
 ```
 
+### Testing and Debugging Policies
+
+Sometimes it is useful to test if a given policy file is valid, especially in a CI environment.
+
+An API endpoint exists at `/api/validate` to validate the syntax of the yaml and policy configuration,
+however it cannot validate that the rules are semantically correct for a given use case.
+
+The API can be used as such:
+
+```sh
+$ curl https://policybot.domain/api/validate -XPOST -F @path/to/policy.yml
+{"message":"failed to parse approval policy: failed to parse subpolicies for 'and': policy references undefined rule 'the devtools team has approved', allowed values: [the devtools team has]","version":"1.12.5"}
+```
+
+You can combine the HTTP response code to automatically detect failures
+
+```sh
+$ rcode=$(curl https://policybot.domain/api/validate -XPOST -F @path/to/policy.yml -s -w "%{http_code}" -o /tmp/response)
+$ if [[ "${rcode}" -gt 299 ]]; then cat /tmp/response && exit 1; fi
+```
+
 ### Caveats and Notes
 
 There are several additional behaviors that follow from the rules above that
--- a/README.md
+++ b/README.md
@@ -335,7 +335,7 @@ $ curl https://policybot.domain/api/validate -XPUT -T path/to/policy.yml
 {"message":"failed to parse approval policy: failed to parse subpolicies for 'and': policy references undefined rule 'the devtools team has approved', allowed values: [the devtools team has]","version":"1.12.5"}
 ```
 
-You can combine the HTTP response code to automatically detect failures
+You can examine the HTTP response code to automatically detect failures
 
 ```sh
 $ rcode=$(curl https://policybot.domain/api/validate -XPUT -T path/to/policy.yml -s -w "%{http_code}" -o /tmp/response)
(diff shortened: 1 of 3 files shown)


server/handler/validate.go
```diff
@@ -0,0 +1,75 @@
...
+	Version string `json:"version"`
+}
+
+func Validate() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		logger := zerolog.Ctx(ctx)
+
+		logger.Info().Msg("Attempting to validate policy file")
+		file, _, err := r.FormFile("policy")
```
bluekeyes: It might be better to PUT/POST the content of the file as the request body (and then read it from `r.Body` here), rather than use the form file upload functionality. Maybe there's no difference in practice, but form file handling has always felt more complicated to me, particularly when using `curl` to make requests.

server/handler/validate.go
```diff
@@ -0,0 +1,75 @@
...
+		if err != nil {
+			baseapp.WriteJSON(w, http.StatusBadRequest, &ValidateCheck{Message: "Unable to read policy file from request", Version: version.GetVersion()})
```
bluekeyes: Would it make sense to construct the `ValidateCheck` upfront (so you only set the version once) and then only set the message before writing it in the other places?

server/handler/validate.go
```diff
@@ -0,0 +1,75 @@
...
+		var policyConfig policy.Config
+		err = yaml.UnmarshalStrict(policyBuf.Bytes(), &policyConfig)
+		if err != nil {
+			baseapp.WriteJSON(w, http.StatusBadRequest, &ValidateCheck{Message: err.Error(), Version: version.GetVersion()})
+			return
+		}
+
+		_, err = policy.ParsePolicy(&policyConfig)
+		if err != nil {
+			baseapp.WriteJSON(w, http.StatusBadRequest, &ValidateCheck{Message: err.Error(), Version: version.GetVersion()})
```
bluekeyes: I think `http.StatusUnprocessableEntity` is more appropriate for the invalid policy cases, leaving the more generic 400 for other errors with processing.

server/handler/validate.go
asvoboda: Initially I had issues using `curl -d`, and used the form uploads. I realized that my issue was with the extra processing curl does, and I needed to specify `--data-binary`. I've also noted this in the README, but it is a bit of a frog trap.

bluekeyes: Do we have the same problem if we use `PUT` instead of `POST` and recommend using `-T` / `--upload-file` with `curl`? I think you can make a reasonable semantic argument for either method in this case.

asvoboda: There doesn't seem to be an issue using PUTs and it cleans up the usage, so lets go with that instead.

README.md
```diff
@@ -321,6 +321,27 @@ disapproval:
...
+however it cannot validate that the rules are semantically correct for a given use case.
+
+The API can be used as such:
+
+```sh
+$ curl https://policybot.domain/api/validate -XPUT -T path/to/policy.yml
+{"message":"failed to parse approval policy: failed to parse subpolicies for 'and': policy references undefined rule 'the devtools team has approved', allowed values: [the devtools team has]","version":"1.12.5"}
+```
+
+You can combine the HTTP response code to automatically detect failures
```
bluekeyes: `You can examine the HTTP response code to automatically detect failures`?

server/handler/validate.go
```diff
@@ -0,0 +1,67 @@
...
+		ctx := r.Context()
+		logger := zerolog.Ctx(ctx)
+
+		logger.Info().Msg("Attempting to validate policy file")
+		check := ValidateCheck{Version: version.GetVersion()}
+
+		requestPolicy, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			check.Message = "Unable to read policy file buffer"
+			baseapp.WriteJSON(w, http.StatusBadRequest, &check)
```
bluekeyes: This might be more appropriate as a 500 error? I'm not sure the client did anything wrong here.

bluekeyes (approved):
Two minor comments, but looks good otherwise.

iamdanfox:
❤❤❤

ferozco:
It would be really sweet if I could leave a comment on a PR to ask policy bot to do the validation and post back whether it is valid

PR #145, opened 2019-12-11
Status handler shouldn't take actions on PRs

asvoboda:


bluekeyes (approved):
Oops, I totally missed that in review.

PR #146, opened 2020-01-08
Add ignore list to changed_files predicate

jmcampanini:
this would allow us to ignore specific files. a common scenario is:

* Team Infra owns the infrastructure code for a project (`build.gradle` files, etc.)
* another team, Team Coder owns the code in a project folder

i'd want the policy to reflect that Team Coder owns everything except the `build.gradle` file. this allows me to use a regex for the folder for Team Coder, and add an ignore pattern to remove files that aren't owned by Team Coder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,8 +130,7 @@ if:
       - "config/special\\.file"
 
   # "only_changed_files" is satisfied if all files changed by the pull request
-  # match at least one regular expression in the list. The "ignore" list is
-  # optional and will ignore any files that match the regular expressions listed.
+  # match at least one regular expression in the list.
   only_changed_files:
     paths:
       - "config/.*"
--- a/README.md
+++ b/README.md
@@ -120,8 +120,9 @@ name: "example rule"
 # exist, the rule applies to every pull request.
 if:
   # "changed_files" is satisfied if any file in the pull request matches any
-  # regular expression in the list. The "ignore" list is optional and will
-  # ignore any files that match the regular expressions listed.
+  # regular expression in the "paths" list. If the "ignore" list is present,
+  # files in the pull request matching these regular expressions are ignored
+  # by this rule.
   changed_files:
     paths:
       - "config/.*"
--- a/README.md
+++ b/README.md
@@ -134,8 +134,6 @@ if:
   only_changed_files:
     paths:
       - "config/.*"
-    ignore:
-      - "config/special\\.file"
 
   # "has_author_in" is satisfied if the user who opened the pull request is in
   # the users list or belongs to any of the listed organizations or teams.
--- a/README.md
+++ b/README.md
@@ -120,17 +120,23 @@ name: "example rule"
 # exist, the rule applies to every pull request.
 if:
   # "changed_files" is satisfied if any file in the pull request matches any
-  # regular expression in the list.
+  # regular expression in the list. The "ignore" list is optional and will
+  # ignore any files that match the regular expressions listed.
   changed_files:
     paths:
       - "config/.*"
       - "server/views/.*\\.tmpl"
+    ignore:
+      - "config/special\\.file"
 
   # "only_changed_files" is satisfied if all files changed by the pull request
-  # match at least one regular expression in the list.
+  # match at least one regular expression in the list. The "ignore" list is
+  # optional and will ignore any files that match the regular expressions listed.
   only_changed_files:
     paths:
       - "config/.*"
+    ignore:
+      - "config/special\\.file"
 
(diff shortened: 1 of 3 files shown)


policy/predicate/file.go
```diff
@@ -37,13 +38,25 @@ func (pred *ChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (boo
...
+	if err != nil {
+		return false, "", errors.Wrap(err, "failed to parse ignore paths")
+	}
+
 	files, err := prctx.ChangedFiles()
 	if err != nil {
 		return false, "", errors.Wrap(err, "failed to list changed files")
 	}
 
+	var cleanedFilenames []string
```
bluekeyes: Maybe `filteredFiles` or `relevanatFiles`? "Cleaned" makes me think the names have been changed in some way.

bluekeyes (commented):
I like this idea. We'll probably want to consider something similar for `only_changed_files`.

policy/predicate/file.go
```diff
@@ -37,13 +38,25 @@ func (pred *ChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (boo
...
+
 	files, err := prctx.ChangedFiles()
 	if err != nil {
 		return false, "", errors.Wrap(err, "failed to list changed files")
 	}
 
+	var relevantFiles []string
 	for _, f := range files {
-		if anyMatches(paths, f.Filename) {
+		if !anyMatches(ignorePaths, f.Filename) {
```
bluekeyes: Could this actually be part of the original loop?

```go
for _, f := range files {
  if anyMatches(ignorePaths, f.Filename) {
    continue
  }
  if anyMatches(paths, f.Filename) {
    return true, "", nil
  }
}
```

README.md
```diff
@@ -120,17 +120,23 @@ name: "example rule"
 # exist, the rule applies to every pull request.
 if:
   # "changed_files" is satisfied if any file in the pull request matches any
-  # regular expression in the list.
+  # regular expression in the list. The "ignore" list is optional and will
+  # ignore any files that match the regular expressions listed.
```
bluekeyes: Maybe `in the "paths" list. If the "ignore" list is present, files in the pull request matching these regular expressions are ignored by this rule.`?

I think clarifying the `paths` list important, but I don't feel as strongly about the rest (I do think it's a bit more explicit about the behavior.)

README.md
```diff
@@ -120,17 +120,23 @@ name: "example rule"
...
     paths:
       - "config/.*"
       - "server/views/.*\\.tmpl"
+    ignore:
+      - "config/special\\.file"
 
   # "only_changed_files" is satisfied if all files changed by the pull request
-  # match at least one regular expression in the list.
+  # match at least one regular expression in the list. The "ignore" list is
+  # optional and will ignore any files that match the regular expressions listed.
```
bluekeyes: Thinking about this now, I can see this behaving in two ways and I'm not sure which is correct:

1. Files in the `ignore` list are completely ignored - it doesn't matter whether they changed or not for the purposes of the rule. This is what you implemented.

2. Files in the `ignore` list are removed from the set of files that are allowed to change, i.e. changes to them cause the predicate to fail.

I think (2) is the closer behavior semantically (not in implementation) to `changed_files`. In the example below, if I make a PR that changes `config/server.yml` and `config/special.file`, should this rule apply?

Right now, I believe the predicate is true because `config/special.file` is ignored, meaning I only changed files in `config/`. This makes it no different from only specifying `paths`.

If the intent is that you ignore things that would otherwise match patterns in `paths`, then I think we need to count "ignored" files as if they are other changed files in the PR.

Or we remove the `ignore` flag from `only_changed_files`.

policy/predicate/file.go
```diff
@@ -64,20 +78,32 @@ func (pred *OnlyChangedFiles) Evaluate(ctx context.Context, prctx pull.Context)
...
```
bluekeyes: Similar to above, I think this could be combined into one loop, but it depends on what we decide the behavior should be.

bluekeyes (commented):
Looking at this more, I'm not sure it makes as much sense for `only_changed_files`, at least as implemented. I expanded on that more in another comment.

The `changed_files` part looks good, aside from a style comment or two.

jmcampanini:
yea, i didnt like how it felt in `only_changed_files`... the tests felt wonky.

i do have confidence it'll bring value being in just `changed_files` though.

bluekeyes (approved):
Sounds good, let's try this out and revisit `only_changed_files` if someone finds a use for it.

PR #151, opened 2020-01-28
add has_labels predicate

### LINKED ISSUE #29: add a has_label_in or similar approval rule predicate
I'd like to be able to apply a set of approval rules based on a label, commit message content, or similar marker on a PR that indicates some special set of circumstances that require different approval rules. The use case is something like the following:

- a user wants to merge a PR that would ordinarily be blocked by some failing automated tests or other policy
- they might have the ability to silence or bypass the automated checks in some way
- if they do this, the PR should require special approval from some set of important people who are vouching for the change being OK to submit, even though checks are being explicitly skipped
- to indicate that, the user adds a "Requires-Lead-Approval" label (or similar) to the PR

I'd imagine the configuration looking something like:
```
approval_rules:
  - name: important people have approved a scary change
    if:
      has_label_in:
        label: ["Requires-Lead-Approval"]
    requires:
      count: 1
      teams:
        - "ImportantGatekeepers"
```

Alternative to the label approach might be something where a commit message or the PR description has a special marker, e.g.
```
approval_rules:
  - name: important people have approved a scary change
    if:
      description_has: "==API_BREAK=="
...
```

ferozco:
## Before this PR
There was no way to create policies that took into account the labels that were applied onto a PR

## After this PR
==COMMIT_MSG==
Add `has_label_applied` field to a clause allowing users to define predicates which are only applied if all of the specified labels are applied
==COMMIT_MSG==

Fixes #29.

## Possible downsides?
There is some additional computation since we now respond to pull request `labeled` and `unlabeled` events, but I don't think this should be a big deal

### DIFF
--- a/policy/approval/predicate.go
+++ b/policy/approval/predicate.go
@@ -33,7 +33,7 @@ type Predicates struct {
 
 	HasSuccessfulStatus *predicate.HasSuccessfulStatus `yaml:"has_successful_status"`
 
-	HasLabelApplied *predicate.HasLabelApplied `yaml:"has_label_applied"`
+	HasLabels *predicate.HasLabels `yaml:"has_labels"`
 }
 
 func (p *Predicates) Predicates() []predicate.Predicate {
@@ -70,8 +70,8 @@ func (p *Predicates) Predicates() []predicate.Predicate {
 		ps = append(ps, predicate.Predicate(p.HasSuccessfulStatus))
 	}
 
-	if p.HasLabelApplied != nil {
-		ps = append(ps, predicate.Predicate(p.HasLabelApplied))
+	if p.HasLabels != nil {
+		ps = append(ps, predicate.Predicate(p.HasLabels))
 	}
 
 	return ps
--- a/policy/approval/predicate.go
+++ b/policy/approval/predicate.go
@@ -32,6 +32,8 @@ type Predicates struct {
 	ModifiedLines *predicate.ModifiedLines `yaml:"modified_lines"`
 
 	HasSuccessfulStatus *predicate.HasSuccessfulStatus `yaml:"has_successful_status"`
+
+	HasLabelApplied *predicate.HasLabelApplied `yaml:"has_label_applied"`
 }
 
 func (p *Predicates) Predicates() []predicate.Predicate {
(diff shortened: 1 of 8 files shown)


policy/approval/predicate.go
```diff
@@ -32,6 +32,8 @@ type Predicates struct {
 	ModifiedLines *predicate.ModifiedLines `yaml:"modified_lines"`
 
 	HasSuccessfulStatus *predicate.HasSuccessfulStatus `yaml:"has_successful_status"`
+
+	HasLabelApplied *predicate.HasLabelApplied `yaml:"has_label_applied"`
```
bluekeyes: I think I'd prefer to call this `HasLabels`, since all of the labels are required.

policy/predicate/label.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/palantir/policy-bot/pull"
+)
+
+type HasLabelApplied []string
+
+var _ Predicate = HasLabelApplied([]string{})
+
+func (pred HasLabelApplied) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+	if len(pred) > 0 {
+
```
bluekeyes: style: remove empty line here

policy/predicate/label.go
```diff
@@ -0,0 +1,55 @@
...
+		labels, err := prctx.Labels()
+		if err != nil {
+			return false, "", errors.Wrap(err, "failed to list pull request labels")
+		}
+
+		for _, requiredLabel := range pred {
+			if !contains(labels, strings.ToLower(requiredLabel)) {
+				return false, "Missing label:" + requiredLabel, nil
```
bluekeyes: style: missing space after the `:` in the message

pull/github.go
```diff
@@ -346,6 +347,25 @@ func (ghc *GitHubContext) LatestStatuses() (map[string]string, error) {
 	return ghc.statuses, nil
 }
 
+func (ghc *GitHubContext) Labels() ([]string, error) {
+	if ghc.labels == nil {
+		var labels []string
+		issueLabels, _, err := ghc.client.Issues.ListLabelsByIssue(ghc.ctx, ghc.owner, ghc.repo, ghc.number, &github.ListOptions{
+			Page:    0,
+			PerPage: 1,
```
bluekeyes: Will this fail if the PR contains more than 1 label? I can understand not supporting arbitrary paging for labels, but it feels like this should be higher, maybe using the maximum page size of 100?

bluekeyes (commented):
Can you also add a test for the new predicate to make sure things like the "all labels are required" behavior work? The existing predicate tests should have good examples.

ferozco:
@bluekeyes ready for another look

bluekeyes (approved):

PR #158, opened 2020-02-10
Add comment_patterns option

### LINKED ISSUE #155: Allow regex matching for comment based approval
Hi, I am looking to have comments approve a PR but be able to setup a rule that matches on a regular expression of the comment rather than whether the comment just contains the string.

For example, I'd like my comment to be able to say "This has been manually tested with build 1.0.12" and have the policy validate against "This has been (manually)? tested with build [1.0.\d+].*"

bluekeyes:
This allows users to specify regular expressions for identifying
approval comments and can also be used to force full/exact string
matches instead of substring matches.

Closes #155.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ if:
     users: ["user1", "user2", ...]
     organizations: ["org1", "org2", ...]
     teams: ["org1/team1", "org2/team2", ...]
-    
+
   # "only_has_contributors_in" is satisfied if all of the commits on the pull
   # request have an author or committer in the users list or that belong to
   # any of the listed organizations or teams.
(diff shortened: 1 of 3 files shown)


asvoboda (approved):
lgtm.

Do you think users might also be interested in using regex capture groups at some point in the future? I can't think of an initial use case right away, so maybe its not necessary to worry about today.

bluekeyes:
Right now, I'm not sure what capture groups in comments would be used for, but it should be possible to support later if we find a use. Nothing prevents defining expressions with groups, but we'd have to modify the `Regexp` type to expose the groups and adjust the relevant match functions.

PR #173, opened 2020-03-16
Make it possible to write status checks without the base branch

jackkleeman:
When you have multiple protected branches without necessarily
predictable names, its useful to still be able to have a required status
check. To do so, we need to have the status check not vary in name
depending on base branch.

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -58,6 +58,10 @@ type PullEvaluationOptions struct {
 	// pattern: <StatusCheckContext>: <Base Branch Name>
 	StatusCheckContext string `yaml:"status_check_context"`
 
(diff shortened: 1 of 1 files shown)


asvoboda:
Hi @jackkleeman ,

This is actually the purpose of the `PostInsecureStatusChecks` [option](https://github.com/palantir/policy-bot/blob/develop/server/handler/base.go#L61-L64), which was added back in #18. The impetus was to prevent anyone with write access to the repo merging changes without approval by pushing fake policies to non-protected branches.

While you can use the legacy option, the best solution is to update the set of required status checks to use the status context with the base branch name.

jackkleeman:
Wow, how did I miss that. Thanks!

PR #176, opened 2020-04-06
add support for check runs

ferozco:
## Before this PR
`has_successful_status` did not support modern check runs.

## After this PR
==COMMIT_MSG==
`has_successful_status` now considers check runs as statuses
==COMMIT_MSG==

## Possible downsides?
N/A


bluekeyes (approved):

bluekeyes:
Thinking about this more, I think we'll also need to add a handler for check run events, since I believe they don't trigger the normal status handler. I can do that in a follow-up PR.

PR #178, opened 2020-04-06
Add handler for check_run events

bluekeyes:
This is functionally equivalent to the status handler, but shares little code because the events have different structures.

Follow-up to #176.

### DIFF
--- a/server/handler/check_run.go
+++ b/server/handler/check_run.go
@@ -1,0 +1,72 @@
+// Copyright 2020 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


server/handler/check_run.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (h *CheckRun) Handles() []string { return []string{"check_run"} }
+
+func (h *CheckRun) Handle(ctx context.Context, eventType, deliveryID string, payload []byte) error {
+	var event github.CheckRunEvent
+	if err := json.Unmarshal(payload, &event); err != nil {
+		return errors.Wrap(err, "failed to parse check_run event payload")
+	}
+
+	if event.GetAction() != "completed" || event.GetCheckRun().GetConclusion() != "success" {
```
asvoboda, line 40: Wouldn't we want to post a status check regardless of the conclusion of the check run?

asvoboda (approved):

server/handler/check_run.go
bluekeyes: At the moment, we care about check runs only for the `has_successful_status` predicate. Evaluating on success saves some computation (the status handler does the same thing) without hurting correctness.

Issue #166, opened 2020-02-24
request_review: mode: team


NargiT:
I would love to see a mode in which request_review would affect a team instead of users of a team.

In my use case our teams are ldap based (running github entreprise on perm) and we have managers in our teams. We do not want create new teams with only tech people because we have to maintain them.

Affecting a group instead of users is also more future proof as is allow new comers to see existing issues. 

How do you think we could leverage this ?

asvoboda:
Hey @NargiT ,

Thanks for the request. We've talked about doing this internally, and while it makes sense we just haven't yet had the time to implement the feature.

NargiT:
from your experience is the feature hard to implement ? I never wrote golang before so I need to know how "big" the feature might be before to help contributing.

AdamDeHovitz:
This has been implemented and can now be closed

PR #183, opened 2020-04-24
Feature: add request-mode `teams`

AdamDeHovitz:
Resolves this issue: https://github.com/palantir/policy-bot/issues/166

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,16 +433,17 @@ limitations. Please file an issue if this functionality is important to you.
 `policy-bot` can automatically request reviewers for all pending rules
 when Pull Requests are opened by setting the `request_review` option.
 
-The `mode` enum modifies how users are selected. There are currently two
+The `mode` enum modifies how users are selected. There are currently three
 supported options:
  * `all-users` to request all users who can approve
  * `random-users` to randomly select the number of users that are required
+ * `prefer-teams` to request users as teams if possible, otherwise behaves like `random-users`
 
 ```yaml
 options:
   request_review:
     enabled: true
-    mode: all-users|random-users
+    mode: all-users|random-users|prefer-teams
 ```
 
 The set of requested reviewers will not include the author of the Pull Request or
--- a/README.md
+++ b/README.md
@@ -235,9 +235,10 @@ options:
   request_review:
     # False by default
     enabled: true
-    # mode modifies how users are selected. `all-users` will request all users
+    # mode modifies how reviewers are selected. `all-users` will request all users
     # who are able to approve the pending rule. `random-users` selects a small
     # set of random users based on the required count of approvals.
+    # "prefer-teams" will request teams to review if possible.
     # defaults to 'random-users'
     mode: all-users|random-users
 
@@ -433,7 +434,7 @@ limitations. Please file an issue if this functionality is important to you.
 `policy-bot` can automatically request reviewers for all pending rules
 when Pull Requests are opened by setting the `request_review` option.
 
-The `mode` enum modifies how users are selected. There are currently three
+The `mode` enum modifies how reviewers are selected. There are currently three
 supported options:
  * `all-users` to request all users who can approve
  * `random-users` to randomly select the number of users that are required
--- a/README.md
+++ b/README.md
@@ -237,9 +237,8 @@ options:
     enabled: true
     # mode modifies how reviewers are selected. `all-users` will request all users
     # who are able to approve the pending rule. `random-users` selects a small
-    # set of random users based on the required count of approvals.
-    # "prefer-teams" will request teams to review if possible.
-    # defaults to 'random-users'
+    # set of random users based on the required count of approvals. `teams` will 
+    # request teams to review if possible. Defaults to 'random-users'.
     mode: all-users|random-users|teams
 
   # "methods" defines how users may express approval.
@@ -438,7 +437,8 @@ The `mode` enum modifies how reviewers are selected. There are currently three
 supported options:
  * `all-users` to request all users who can approve
  * `random-users` to randomly select the number of users that are required
- * `teams` to request users as teams if possible, otherwise behaves like `random-users`
+ * `teams` to request teams for review. Teams must be repository collaborators
+   with at least read access.
 
 ```yaml
 options:
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ options:
     # set of random users based on the required count of approvals.
     # "prefer-teams" will request teams to review if possible.
     # defaults to 'random-users'
-    mode: all-users|random-users
+    mode: all-users|random-users|teams
 
   # "methods" defines how users may express approval.
   methods:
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @AdamDeHovitz! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

README.md
```diff
@@ -235,9 +235,10 @@ options:
   request_review:
     # False by default
     enabled: true
-    # mode modifies how users are selected. `all-users` will request all users
+    # mode modifies how reviewers are selected. `all-users` will request all users
     # who are able to approve the pending rule. `random-users` selects a small
     # set of random users based on the required count of approvals.
+    # "prefer-teams" will request teams to review if possible.
     # defaults to 'random-users'
     mode: all-users|random-users
```
bluekeyes: Let's also add `prefer-teams` to this example list

README.md
```diff
@@ -433,16 +434,17 @@ limitations. Please file an issue if this functionality is important to you.
 `policy-bot` can automatically request reviewers for all pending rules
 when Pull Requests are opened by setting the `request_review` option.
 
-The `mode` enum modifies how users are selected. There are currently two
+The `mode` enum modifies how reviewers are selected. There are currently three
 supported options:
  * `all-users` to request all users who can approve
  * `random-users` to randomly select the number of users that are required
+ * `prefer-teams` to request users as teams if possible, otherwise behaves like `random-users`
```
bluekeyes: My first thought is that this should be `all-teams` or just `teams` without the fallback behavior, and we can expect people to use an appropriate mode in their policies. If feels weird that you might sometimes get teams and sometimes get users with the same mode.

GitHub also recently added native functionality to do individual user assignment when teams are requested.

policy/reviewer/reviewer.go
```diff
@@ -103,9 +103,9 @@ func selectOrgMembers(prctx pull.Context, allOrgs []string) ([]string, error) {
 	return allOrgsMembers, nil
 }
 
-func selectAdmins(prctx pull.Context) ([]string, error) {
+func selectAdmins(prctx pull.Context) (map[string][]string, error) {
```
bluekeyes: Kind of unrelated, but can we rename this `selectAdminTeamMembers` to reflect what it actually does?

policy/reviewer/reviewer.go
```diff
@@ -177,42 +195,54 @@ func SelectReviewers(ctx context.Context, prctx pull.Context, result common.Resu
...
 		}
 
-		var possibleReviewers []string
-		for u := range allUsers {
-			// Remove the author and any users who aren't collaborators
-			// since github will fail to assign _anyone_ if the request contains one of these
-			_, ok := collaboratorsToConsider[u]
-			if u != prctx.Author() && ok {
-				possibleReviewers = append(possibleReviewers, u)
+		allUsersKeys := make([]string, 0, len(allUsers))
```
bluekeyes: It took me a minute to realize this list has all the user names in it. Maybe call it `users` or `candidates` (a term we use in other parts of policy-bot) instead?

policy/reviewer/reviewer.go
```diff
@@ -177,42 +195,54 @@ func SelectReviewers(ctx context.Context, prctx pull.Context, result common.Resu
...
+		possibleReviewers := getPossibleReviewers(prctx, allUsersKeys, collaboratorsToConsider)
+		if child.GetMode() == common.RequestModePreferTeams {
+			var possibleTeams []string
+			teamMembersCount := 0
+			for team, users := range allTeamsWithUsers {
+				possibleReviewersInTeamCount := len(getPossibleReviewers(prctx, users, collaboratorsToConsider))
+				if possibleReviewersInTeamCount > 0 {
+					possibleTeams = append(possibleTeams, team)
+					teamMembersCount += possibleReviewersInTeamCount
+				}
```
bluekeyes: This is clever, but I think it will be more obvious to use `prctx.Teams()` (which returns all team collaborators and permissions) to build the set of possible reviewer teams.

That might remove the need to track eligible users by team, which could simplify the rest of this.

policy/reviewer/reviewer.go
```diff
@@ -177,42 +195,54 @@ func SelectReviewers(ctx context.Context, prctx pull.Context, result common.Resu
...
+			}
+			if len(possibleTeams) > 0 && teamMembersCount >= child.ReviewRequestRule.RequiredCount {
```
bluekeyes: I don't think it's necessary to compare the total count of team members to the required count. The existing code for user selection doesn't do this and I hope it will be obvious to people writing policies that they've requested more reviewers than exist for the teams in use!

policy/reviewer/reviewer.go
```diff
@@ -177,42 +195,54 @@ func SelectReviewers(ctx context.Context, prctx pull.Context, result common.Resu
...
+			if child.GetMode() == common.RequestModeAllUsers {
 				logger.Debug().Msgf("Found %d total reviewers after removing author and non-collaborators; requesting all", len(possibleReviewers))
 				usersToRequest = append(usersToRequest, possibleReviewers...)
-			case common.RequestModeRandomUsers:
-				logger.Debug().Msgf("Found %d total candidates for review after removing author and non-collaborators; randomly selecting %d", len(possibleReviewers), child.ReviewRequestRule.RequiredCount)
-				randomSelection := selectRandomUsers(child.ReviewRequestRule.RequiredCount, possibleReviewers, r)
-				usersToRequest = append(usersToRequest, randomSelection...)
-			default:
-				return nil, fmt.Errorf("unknown mode '%s' supplied", child.ReviewRequestRule.Mode)
+				continue
```
bluekeyes: I know a lot of this structure already existed, but I'm finding this use of `continue` confusing given the size of the body of this loop. Do you think we could restructure the old `switch` statement so that each review mode still has a dedicated case?

I think this is easier if we remove the fallback behavior for teams, but maybe a `fallthrough` statement would make that clear if you want to keep the fallback behavior. Inverting the `possibleReviewers` length check and having that be the only `continue` could help as well.

Alternately, maybe there's a helper method to extract that implements the mode switching.

policy/reviewer/reviewer.go
```diff
@@ -118,35 +118,53 @@ func selectAdmins(prctx pull.Context) ([]string, error) {
...
+		if u != prctx.Author() && ok {
+			possibleReviewers = append(possibleReviewers, u)
+		}
+	}
+	return possibleReviewers
+}
+
+func SelectReviewers(ctx context.Context, prctx pull.Context, result common.Result, r *rand.Rand) ([]string, []string, error) {
 	var usersToRequest []string
+	var teamsToRequest []string
```
bluekeyes: I think these need to be initialized as `[]string{}` so that the function always returns non-nil slices when using only one type of assignment. The GitHub API / `go-github` requires passing non-nil slices in the `github.ReviewersRequest` struct.

bluekeyes (commented):
Thanks for implementing this. I'm not convinced that the fallback to user assignment is necessary, so I'm interested in the practical cases where it would help. Otherwise, I had a few style/structure comments.

AdamDeHovitz:
@bluekeyes finally got around to responding, thanks for the review!
Something that I lacked context on was team name format. ctx.TeamMemberships returns `org/teamname` while ctx.Teams seems to return `teamname`. Do you know if it's safe to always cut the org name?

bluekeyes:
Yeah, it should be safe to cut or add to the team name as needed. The existing code has a few similar operations: [1](https://github.com/palantir/policy-bot/blob/develop/pull/github_membership.go#L71), [2](https://github.com/palantir/policy-bot/blob/develop/pull/github_membership.go#L95)

AdamDeHovitz:
Great, I think I responded to all of your feedback. Let me know if there's any other changes needed to get this merged!

bluekeyes (approved):

ajlake:
👍

PR #187, opened 2020-05-12
Enable asynchronous webhook processing

bluekeyes:
While policy-bot is generally pretty fast at responding to webhooks,
this avoids any possibility for timeouts for repositories that have
large policies that require many requests to evaluate.


ajlake:
👍

PR #188, opened 2020-05-12
Use unqualified names in team assignment

bluekeyes:
Because you can only assign teams from the organization, the API expects
to get unqualified team slugs as input.

cc @AdamDeHovitz


ajlake:
👍

PR #196, opened 2020-06-24
help make validation errors easier to read

jmcampanini:
This is an attempt to help clarify the error messages returned by the validation code. The text is returned by the validation endpoint, and can be difficult to read. The specific changes to error message are:

* reworded error for the root level `and` using `policy` instead of `subpolicy`
* added `(index=#)` to help identify which of the children rules are broken

With the following policy file:

```yaml
policy:
  approval:
    - rule1
    - or:
       - rule2
       - rule3
       - ruleUnknown

approval_rules:
  - name: rule1
  - name: rule2
  - name: rule3
```

The error would be:

```
failed to parse subpolicy for 'and': failed to parse subpolicy for 'or': policy references undefined rule 'ruleUnknown', allowed values: [rule1 rule2 rule3]

// when formatted
failed to parse subpolicy for 'and': // due to the implied root-level AND
failed to parse subpolicy for 'or':
policy references undefined rule 'ruleUnknown', allowed values: [rule1 rule2 rule3]
```

and now the new err is:

```
failed to parse policy (index=1): failed to parse subpolicy (index=2) for 'or': policy references undefined rule 'ruleUnknown', allowed values: [rule1 rule2 rule3]

// when formatted
failed to parse policy (index=1):
failed to parse subpolicy (index=2) for 'or':
policy references undefined rule 'ruleUnknown', allowed values: [rule1 rule2 rule3]
```

### DIFF
--- a/policy/approval/parse.go
+++ b/policy/approval/parse.go
@@ -85,10 +85,15 @@ func parsePolicyR(policy interface{}, rules map[string]*Rule, depth int) (common
 		}
 
 		var subrequirements []common.Evaluator
-		for _, subpolicy := range values {
+		for index, subpolicy := range values {
 			subreq, err := parsePolicyR(subpolicy, rules, depth+1)
 			if err != nil {
-				return nil, errors.WithMessage(err, fmt.Sprintf("failed to parse subpolicies for '%s'", op))
+				errMsg := fmt.Sprintf("failed to parse subpolicy (index=%d) for '%s'", index, op)
+				if depth == 0 {
+					errMsg = fmt.Sprintf("failed to parse policy (index=%d)", index)
+				}
+
+				return nil, errors.WithMessage(err, errMsg)
 			}
 			subrequirements = append(subrequirements, subreq)
 		}
--- a/policy/approval/parse_test.go
+++ b/policy/approval/parse_test.go
@@ -16,6 +16,7 @@ package approval
 
 import (
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


bluekeyes:
While the code change is fine, I don't think it addresses the issue that validation errors are hard to read. I think the original complaint was driven by three factors:

1. The error message does not appear first. Instead, you see the path to the rule that had an error, placing the actual error in the middle of the text.
2. The error includes all possible valid rules, which can be a large number and generates a lot of text if they have long names. The default slice printing does not clearly separate the slice elements.
3. The large error is a single line (or block when wrapped) making it hard to parse.

Assuming we want to keep the parsing logic mostly intact for now, here are some ideas for changes:

- Change the wrapping order so that the error appears first followed by the path through the policy. This would be similar to a stack trace in most languages.
- Do not print allowed rules if a rule reference is wrong. This is probably easier to fix by searching for the rules in your text editor when fixing the policy.
  - Alternately, only print up to N alternatives: `allowed values: [a, b, c,  ... and 10 more]`
  - Alternately, use a fuzzy match to print a suggestion: `ruuleX not found, did you mean 'ruleX'?`
- Change the slice formatting to quote each value and separate entries with commas
- Add newlines where it makes sense, perhaps after each level of the policy or in between possible values

We don't need all of these and there are probably other approaches that would work. But we should have hypothesis about why errors are hard to read and then make changes to test them.

johnwiseheart:
Hey folks,

I agree that generally people are used to reading errors like a stack trace in other languages - with the error message on top, and a snippet of where the error went wrong. For example, I've recently been working in groovy with snakeyaml, which when failing to parse yaml has the following format of error:

```
> java.lang.Exception: expected '<document start>', but found '<block mapping start>'
   in 'string', line 4, column 1:
      "Some key":
```

I'd think that something similar here would be useful:

```
policy references undefined rule 'ruleUnknown' at Line X, Column Y
```

As mentioned, I don't think providing the allowed values is that valuable for our use case, since theres hundreds of allowed values, and they just end up making the error message harder to read.

bluekeyes:
Closing this out due to age, but I think the problem still exists, so we may want to revisit.

PR #199, opened 2020-06-29
Use maximum page size when listing teams

bluekeyes:
This should reduce the total number of requests when operating in
organizations that have a lot of teams.


ajlake:
👍

PR #204, opened 2020-07-16
Add GitHub client request timeout, and upgrade go-githubapp and go-github

jmcampanini:
Add the ability to set and configure GitHub client request timeouts.

This PR also upgrades:

* go-githubapp to v0.4.0 - to get the option to set the timeout on the `ClientCreator`
* go-github to v32 (as a dependency of go-githubapp v0.4.0)


server/config.go
```diff
@@ -49,8 +50,9 @@ type CachingConfig struct {
 }
 
 type WorkerConfig struct {
-	Workers   int `yaml:"workers"`
-	QueueSize int `yaml:"queue_size"`
+	Workers       int           `yaml:"workers"`
+	QueueSize     int           `yaml:"queue_size"`
+	ClientTimeout time.Duration `yaml:"client_timeout"`
```
jmcampanini: considering the workers are the ones making the requests, i put the configuration in this block. open to having it in a different location and a different name

bluekeyes: I think here makes the most sense, followed the the generic "options" config.

What do you think about calling this `request_timeout` or `github_timeout`? I think it's kind of ambiguous what "client" refers to here.

server/server.go
```diff
@@ -90,6 +90,7 @@ func New(c *Config) (*Server, error) {
 	cc, err := githubapp.NewDefaultCachingClientCreator(
 		c.Github,
 		githubapp.WithClientUserAgent(userAgent),
+		githubapp.WithClientTimeout(c.Workers.ClientTimeout),
```
bluekeyes: If the property is not set, passing the zero value is equivalent to no timeout?

Since you can't turn off queued background processing, I think we should set a default timeout that users can increase or decrease, but not disable. Not sure what the value should be though, maybe 10s? 30s?

bluekeyes:
One more thing, we should add this (commented out) to the example configuration file

server/config.go
jmcampanini: keeping here and renaming to `github_timeout`

server/server.go
jmcampanini: correct:

```
	// A Timeout of zero means no timeout.
```

i like 10s as a start

bluekeyes (approved):

PR #211, opened 2020-07-28
Allow running with unchanged Docker image

sideeffffect:
similar in spirit to https://github.com/palantir/bulldozer/pull/189

This should make the Docker image runnable without any modifications, only with adding some environment variables.


vendor/github.com/palantir/go-baseapp/baseapp/config.go
```diff
@@ -34,6 +35,19 @@ type HTTPConfig struct {
 	ShutdownWaitTime *time.Duration `yaml:"shutdown_wait_time" json:"shutdownWaitTime"`
 }
 
+// SetValuesFromEnv sets values in the configuration from coresponding
+// environment variables, if they exist. The optional prefix is added to the
+// start of the environment variable names.
+func (c *HTTPConfig) SetValuesFromEnv(prefix string) {
```
sideeffffect, line 42: @jasei any idea why I'm getting [this](https://app.circleci.com/pipelines/github/palantir/policy-bot/514/workflows/21b037a5-7960-4c31-a89c-7f35dc392ccb/jobs/1519)?
```
server/config.go:70:11: c.Server.SetValuesFromEnv undefined (type baseapp.HTTPConfig has no field or method SetValuesFromEnv)
```

bluekeyes: I believe this is because you modified a vendored file from a library we depend on and part of the build process resets the vendored files to the correct upstream state.

If you want to add this method, you'll have to submit it to https://github.com/palantir/go-baseapp separately and then update that library version here.

bluekeyes: Now that the `go-baseapp` change merged, you should be able to use it here. After updating the version, be sure to run `./godelw mod` to vendor the new dependency files.

server/config.go
```diff
@@ -67,6 +67,7 @@ func ParseConfig(bytes []byte) (*Config, error) {
 	}
 
 	c.Options.FillDefaults()
+	c.Server.SetValuesFromEnv("")
```
bluekeyes: As I mentioned on the `go-baseapp` PR, I think we should use `POLICYBOT_` as the prefix for these values. The GitHub values below can stay as-is, since they already use `GITHUB_` as a prefix.

sideeffffect:
we have a green build :tada: 
/cc @jasei

bluekeyes (approved):

bluekeyes:
Thanks for adding this! I'll try to tag a release with these changes in the next few days.

sideeffffect:
hi @bluekeyes do you think you will get to releasing the new version anytime soon?

bluekeyes:
Sorry about that, last week was up busier than expected. [1.17.1](https://github.com/palantir/policy-bot/releases/tag/v1.17.1) is out now.

sideeffffect:
Thanks a bunch @bluekeyes! :pray:

PR #212, opened 2020-07-29
Use slug-based APIs for team operations

### LINKED ISSUE #200: Use slug-based APIs for team operations
On large organizations internally, we're seeing issues with rate limits evaluating PRs. Looking at requests suggests part of this is due to listing all teams in the organization to get the IDs for use in other API calls. On GitHub.com and in GitHub Enterprise 2.21+, the v3 API now has endpoints to get team information using the team slug, rather than the ID. This will allow Policy Bot to make requests only for the teams mentioned in the relevant policy rules, a significant savings if the org has hundreds of teams.

Note that after this change, Policy Bot will require GHE 2.21.1 or newer.

bluekeyes:
This reduces the total number of requests by eliminating calls to enumerate all teams, which were previously required to discover all team IDs and map them to names.

These APIs are only available on GitHub.com and in GitHub Enterprise 2.21 or newer, so policy-bot is no longer compatible with older versions of GitHub Enterprise.

I still need to fix up the tests and check to see if there are any other optimizations we should make to team operations.

Fixes #200.


asvoboda (approved):

bluekeyes:
Since this will cut off the number of GHE versions we support pretty dramatically, I intend to release it as `v1.18.0` after releasing a `v1.17.1` with some of the other performance improvements and changes that merged recently.

PR #214, opened 2020-08-04
Only process review requests when enabled

bluekeyes:
While reviewers were never assigned incorrectly, we previously made
unnecessary requests to GitHub to get collaborator and reviewer details
that were discarded when no enabled rules existed. Now, check for rules
that enable review requests first and only start reviewer assignment if
at least one exists.

This also cleans up some logging and other minor style things in code I
touched to fix the problem.


ajlake (approved):

asvoboda (approved):
Nice find!

PR #215, opened 2020-08-04
Use v3 API to list repository collaborators

bluekeyes:
We don't use any v4-only fields for this operation and using normal REST
requests lets us cache the responses in the global HTTP cache.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -269,11 +269,38 @@ func (ghc *GitHubContext) Reviews() ([]*Review, error) {
 
 func (ghc *GitHubContext) RepositoryCollaborators() (map[string]string, error) {
 	if ghc.collaborators == nil {
-		if err := ghc.loadCollaboratorData(); err != nil {
-			return nil, err
+		opts := github.ListCollaboratorsOptions{
+			ListOptions: github.ListOptions{
+				PerPage: 100,
+			},
 		}
+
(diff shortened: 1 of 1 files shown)


pull/github.go
```diff
@@ -269,13 +269,40 @@ func (ghc *GitHubContext) Reviews() ([]*Review, error) {
...
+			}
+			opts.Page = res.NextPage
 		}
+		ghc.collaborators = collaborators
 	}
 	return ghc.collaborators, nil
 }
 
+func coalescePermission(perms map[string]bool) string {
+	for _, p := range []string{"admin", "push", "pull"} {
```
asvoboda: While reviewing this, I think I discovered a separate bug for `WriteCollaborators` where we assume the permission for writing is `write` rather than `push`

https://github.com/palantir/policy-bot/blob/develop/policy/common/actor.go#L39

asvoboda (approved):

pull/github.go
bluekeyes: I think `write` is correct there, since it uses a different API: https://docs.github.com/en/rest/reference/repos#get-repository-permissions-for-a-user

I think the bug is actually in my conversion... the GraphQL object I replaced uses `read` and `write`: https://docs.github.com/en/graphql/reference/enums#repositorypermission

PR #220, opened 2020-08-14
Increase max rule nesting depth to 10

### LINKED ISSUE #168: Increase max nesting depth
I was wondering if it would be possible to increase the max nesting depth of ors/ands to 6-7? I have a complex use case that requires just a little bit more nesting than is currently supported :)

bluekeyes:
Fixes #168.


ajlake:
👍

Issue #230, opened 2020-09-08
Regression since 1.17.0 - policy does not recognize my github app name


NargiT:
Hello,

We are using GitHub Enterprise Server 2.20.5 so I cannot do any test on +1.17.1.

I did an upgrade from 1.16.3 to 1.17.0 and to 1.17.1 and now policy is not able to recognize the bot user and fail the status check.

Some context:
I use a Github App to create pull request to update my repository (same idea as renovate or dependabot).
The commit and the author have the following format:

```
Author:     mybot[bot] <mybot[bot]@users.noreply.github.domain.ch>
AuthorDate: Tue Sep 8 06:39:34 2020 +0000
Commit:     mybot[bot] <mybot[bot]@users.noreply.github.domain.ch>
CommitDate: Tue Sep 8 06:39:34 2020 +0000
```

```
policy:
  approval:
    - mybot upgrade

approval_rules:
  - name: mybot upgrade
    if:
      has_author_in:
        users: ['mybot[bot]']
      author_is_only_contributor: true
      targets_branch:
        pattern: ^master
    options:
      invalidate_on_push: true
```
The error

```
The pull request author "mybot" does not meet the required membership conditions
```

And if I modify the users list to `['mybot']` only I get another error

```
Commit 8a82a72953 was authored or committed by a different user
```

I also tried to remove the "bot" and only use mybot for commit and pull request creation, but end up with the previous error `Commit xxxx was authored or committed by a different user`

This is working very well with 1.16.3 but since 1.17.0 policy does not recognize my bot anymore.
Which field is used to match the user name ?

Looks like the issue comes from this changes https://github.com/palantir/policy-bot/compare/v1.16.3...v1.17.0

bluekeyes:
The first message

> The pull request author "mybot" does not meet the required membership conditions

uses the `user.login` field of the [pull request](https://docs.github.com/en/rest/reference/pulls#get-a-pull-request) or the `author.login` field of the [`PullRequest`](https://docs.github.com/en/graphql/reference/objects#pullrequest) GraphQL object. I think there actually is a bug here, as the GraphQL API returns logins in a different format and we didn't account for that when getting the pull request author. However, that bug was introduced back in 1.8.0.

The second message

> Commit 8a82a72953 was authored or committed by a different user

[...]

PR #231, opened 2020-09-08
Fix author value for bot-authored pull requests

bluekeyes:
The v3 API includes the '[bot]' suffix, while the v4 API does not. Depending on how the context was created, the object may have come from either API, so we need the helper function to standardize the value.

This might fix the bug in #230, but we'd have to backport to a 1.17.x release branch to test it out.


asvoboda (approved):

NargiT:
thanks !

PR #233, opened 2020-09-28
Load app name from GitHub dynamically

### LINKED ISSUE #207: Load app name from GitHub
There's no reason to include the `app_name` key in the server configuration, since the app can look this up dynamically from GitHub. Since policy-bot posts failed statuses when the name in the config doesn't match the name in GitHub, this option leads to confusion, for example #206, #107, and #72.

Using the `app_name` may also be unnecessary if we switch to IDs for auditing, per #109.

bluekeyes:
The 'options.app_name' server configuration value is now deprecated and ignored. The struct field is still defined to avoid breaking configuration when using UnmarshalStrict.

Fixes #207.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -15,6 +15,7 @@
 package server
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -91,7 +92,7 @@ func New(c *Config) (*Server, error) {
 		githubTimeout = 10 * time.Second
 	}
 
-	userAgent := fmt.Sprintf("%s/%s", c.Options.AppName, version.GetVersion())
+	userAgent := fmt.Sprintf("policy-bot/%s", version.GetVersion())
 	cc, err := githubapp.NewDefaultCachingClientCreator(
 		c.Github,
 		githubapp.WithClientUserAgent(userAgent),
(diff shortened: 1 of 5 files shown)


asvoboda (approved):

server/server.go
```diff
@@ -91,7 +92,7 @@ func New(c *Config) (*Server, error) {
 		githubTimeout = 10 * time.Second
 	}
 
-	userAgent := fmt.Sprintf("%s/%s", c.Options.AppName, version.GetVersion())
+	userAgent := fmt.Sprintf("policy-bot/%s", version.GetVersion())
```
bluekeyes, line 95: Using the app name here requires creating a separate app client and didn't seem worth it. I think `policy-bot` is an appropriate user agent in most cases, even if the app is registered with a different name.

ekosz:
Hi there. This is a great change and something we've been struggling to override in out k8s cluster (can't quite get the mounting of our config file working). Do you all have a plan release date in mind for this?

bluekeyes:
I hope to have a release with this by the end of the week, either as 1.19.2 or as part of a larger 1.20.0 release with some other changes I'm working on.

We also publish a `snapshot` tag to Docker Hub from the `develop` branch if you want to try this before the release is out.

PR #242, opened 2020-10-21
Run reviewer assignment on all evaluations

### LINKED ISSUE #152: Optimize event evaluation based on policy predicates
Recent predicate additions (`has_labels`, `has_successful_status`) require action on additional event types (label/unlabel, status). It would be nice to check the policy and only respond to these events when the policy contains rules that use these predicates. While we have to fetch the policy in all cases, we should be able to avoid the additional GitHub requests associated with validation.

### LINKED ISSUE #153: Assign reviewers when predicate rules change
Recent predicate additions (`has_labels`, `has_successful_status`) can enable rules well after a PR has opened in response to outside events. When these rules are enabled, we should re-run reviewer assignment to assign reviewers for these new rules, if enabled.

bluekeyes:
Running reviewer assignment more often will fix problems where new reviewers are needed due to things like label, status, or file changes. But it also adds risk for duplicate or extra assignments. We try to avoid that by:

- Only assigning missing reviewers
- Using random delays to avoid concurrent reviewer assignment on the same PR
- Minimizing the number of evaluations

But I expect it won't be perfect and will have some issues. Fixing this for real requires a coordination backend / shared storage.

The implementation has several components, split up by commits, but it might be better as separate PRs. I recommend reviewing by commit, but I'll try to summarize here:

1. Change random selection to use a RNG seeded with the pull request creation time. This is so we'll always select the same reviewers for a given PR when using random selection.
2. Add a function to get the currently assigned and dismissed reviewers 
3. Update reviewer request logic to only request reviews from missing reviewers (that have not been dismissed)
4. Add trigger conditions to predicates and evaluators so we only do work if the policy could have changed due to an event. This replaces the old `requestReviews` boolean used by handlers.

I still need to write some tests and try it out for real, but I wanted to put up the PR now to get feedback on the approach.

Fixes #153.
Fixes #152.


policy/approval/evaluate.go
```diff
@@ -28,6 +28,13 @@ type evaluator struct {
 	root common.Evaluator
 }
 
+func (eval *evaluator) Trigger() common.Trigger {
+	if eval.root != nil {
+		return eval.root.Trigger()
+	}
+	return common.TriggerCommit
```
bluekeyes:
```suggestion
	return common.TriggerStatic
```

If there's no root, there's nothing to evaluate.

policy/common/trigger.go
```diff
@@ -0,0 +1,90 @@
...
+	// TriggerStatic is a name for the empty trigger set and means the
+	// computation never needs updating.
+	TriggerStatic Trigger = 0
+
+	TriggerCommit Trigger = 1 << iota
+	TriggerComment
+	TriggerReview
+	TriggerLabel
+	TriggerStatus
+	TriggerPullRequest
```
bluekeyes, line 33: `TriggerPullRequest` is for things like editing the body and title or changing the target branch.

policy/common/trigger.go
```diff
@@ -0,0 +1,90 @@
...
+}{
+	{TriggerCommit, "Commit"},
+	{TriggerComment, "Comment"},
+	{TriggerReview, "Review"},
+	{TriggerLabel, "Label"},
+	{TriggerStatus, "Status"},
+	{TriggerPullRequest, "PullRequest"},
+}
+
+// Matches returns true if flapg contains any of the flags of this trigger.
```
bluekeyes:
```suggestion
// Matches returns true if flag contains any of the flags of this trigger.
```

server/handler/base.go
```diff
@@ -215,47 +223,73 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, re
...
-	requestedUsers, requestedTeams, err := reviewer.SelectReviewers(ctx, prctx, reqs, r)
-	if err != nil {
-		return errors.Wrap(err, "failed to select reviewers")
-	}
+	// This is a terrible strategy to avoid conflicts between closely spaced
+	// events assigning the same reviewers, but I expect it to work alright in
+	// practice and it avoids any kind of coordination backend:
+	//
+	// Wait a random amount of time to space out events then check for existing
+	// reviewers and apply the difference. The idea to order two competing
```
bluekeyes:
```suggestion
	// reviewers and apply the difference. The idea is to order two competing
```

server/handler/base.go
```diff
@@ -215,47 +223,73 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, re
...
+	// events such that one observes the applied reviewers of the other.
+	//
+	// Use the global random source instead of the per-PR source so that two
+	// events for the same PR don't wait for the same amount of time.
+	delay := time.Duration(rand.Intn(maxDelayMillis)) * time.Millisecond
+	logger.Debug().Msg("Waiting for %s to spread out reviewer processing")
```
bluekeyes:
```suggestion
	logger.Debug().Msgf("Waiting for %s to spread out reviewer processing", delay)
```

policy/common/trigger.go
```diff
@@ -0,0 +1,90 @@
...
+func (t Trigger) Matches(flags Trigger) bool {
+	return t&flags > 0
+}
+
+func (t Trigger) String() string {
```
asvoboda, line 62: Why do we need to implement the Stringer interface here? Just for debugging?

policy/common/trigger.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2020 Palantir Technologies, Inc.
```
asvoboda, line 1: Do you think it would be worthwhile to add some unit tests for `Trigger`? Or is it just useful that we test the downstream usage in the predicates?

pull/github.go
```diff
@@ -317,22 +324,74 @@ func coalescePermission(perms map[string]bool) string {
...
+			return nil, err
+		}
 	}
+	return ghc.reviewers, nil
+}
 
-	reviews, err := ghc.Reviews()
-	if err != nil {
-		return false, errors.Wrap(err, "Unable to list reviews")
+func (ghc *GitHubContext) loadRequestedReviewers() error {
```
asvoboda, line 336: In practice do you think its going to be worthwhile to avoid the v3 caching?

asvoboda (approved):
Very happy to see the review feature less full of wonk :)

policy/common/trigger.go
bluekeyes: Yeah, I include it in a few log messages and didn't want to deal with decoding integers back into the flags.

policy/common/trigger.go
bluekeyes: Yes, this is one of the changes that needs some tests.

pull/github.go
bluekeyes: Kind of hard to say, but the v4 API is the only one that supports filtering timeline events by type. I think combining these  queries into one request with filtering will be better than paging through multiple v3 requests with caching, but that's a guess.

bluekeyes:
I think one aspect I'll still need to consider is what happens if a person is requested, they review the PR, and then new commits are pushed. Right now, there won't be any review requests and we'll request reviews again. I think instead we'll probably need to _only_ do that if `invalidate_on_push` is true for the rules that request reviews and if `TriggerCommit` was present.

bluekeyes:
After testing, we don't need a special case for re-requesting reviews because we still only do reviewer assignment if the evaluation status is pending. Policies that invalidate approval on push will naturally return to the pending state, while policies that do not will remain approved. Any reviewers that are still requested will not be requested again, but anyone who has already reviewed will get another request.

I think this behavior is desirable because it improves notifications for PR updates but perhaps we'll get requests for an option to turn it off.

Otherwise, this seems functionally correct in the testing I've done so far, although I don't have a great way of stress testing the event deduplication strategy.

asvoboda (approved):

PR #243, opened 2020-10-21
Add from_branch predicate

### LINKED ISSUE #241: support predicate for source branch
It seems like it would be useful to have an analogous version of `targets_branch` for specifying the source branch. For my particular use case, I would like to have a policy where pull requests to the `master` branch can only come from the `develop` branch.

danny-wh:
Fixes #241.

This adds a new predicate called `source_branch` that's analogous to `targets_branch`, but for the PR's source branch. I'm very open to different names for this option, but I haven't come up with something I love.

I changed the existing `targets_branch` tests to also run the patterns against the source branch. I also added tests for making sure both branches are correctly set when creating the `GitHubContext` objects.

### DIFF
--- a/policy/approval/predicate.go
+++ b/policy/approval/predicate.go
@@ -28,7 +28,7 @@ type Predicates struct {
 	AuthorIsOnlyContributor *predicate.AuthorIsOnlyContributor `yaml:"author_is_only_contributor"`
 
 	TargetsBranch *predicate.TargetsBranch `yaml:"targets_branch"`
-	SourceBranch  *predicate.SourceBranch  `yaml:"source_branch"`
+	FromBranch    *predicate.FromBranch    `yaml:"from_branch"`
 
 	ModifiedLines *predicate.ModifiedLines `yaml:"modified_lines"`
 
--- a/policy/approval/predicate.go
+++ b/policy/approval/predicate.go
@@ -63,6 +63,10 @@ func (p *Predicates) Predicates() []predicate.Predicate {
 	if p.TargetsBranch != nil {
 		ps = append(ps, predicate.Predicate(p.TargetsBranch))
 	}
+	if p.FromBranch != nil {
+		ps = append(ps, predicate.Predicate(p.FromBranch))
+	}
+
 	if p.ModifiedLines != nil {
 		ps = append(ps, predicate.Predicate(p.ModifiedLines))
 	}
--- a/policy/approval/predicate.go
+++ b/policy/approval/predicate.go
@@ -28,6 +28,7 @@ type Predicates struct {
 	AuthorIsOnlyContributor *predicate.AuthorIsOnlyContributor `yaml:"author_is_only_contributor"`
 
 	TargetsBranch *predicate.TargetsBranch `yaml:"targets_branch"`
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @dbwhite831! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @dbwhite831! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

policy/approval/predicate.go
```diff
@@ -28,6 +28,7 @@ type Predicates struct {
 	AuthorIsOnlyContributor *predicate.AuthorIsOnlyContributor `yaml:"author_is_only_contributor"`
 
 	TargetsBranch *predicate.TargetsBranch `yaml:"targets_branch"`
+	SourceBranch  *predicate.SourceBranch  `yaml:"source_branch"`
```
bluekeyes: I think calling this `from_branch` would match the other names a bit better as a predicate: "if targets branch", "if has author in", "if from branch", etc

You'll also need to add an `if` statement around L65 to append this to the predicate list when it is set.

pull/github_test.go
```diff
@@ -396,6 +396,35 @@ func TestIsOrgMember(t *testing.T) {
...
+			Login: github.String("testorg2"),
+		},
+		Name: github.String("testrepofork"),
+	}
+
+	ctx := makeContext(t, rp, crossRepoPr)
+
+	base, head := ctx.Branches()
+	assert.Equal(t, "develop", base, "cross-repo base branch was not correctly set")
+	assert.Equal(t, "testorg2:test-branch", head, "cross-repo head branch was not correctly set")
```
bluekeyes, line 425: The documentation for this predicate should probably mention the fact that source branches for PRs from forks will have a `:` and that patterns will need to account for that.

Alternately, we could strip the organization from this name and then add a separate `from_fork` predicate for policies that want to differentiate forks. Right now, `from_branch` serves both purposes and a pattern like `:` would work as a "is this a fork" condition.

bluekeyes (commented):
Thanks for adding this. It would also be nice if you could update the README to add the new predicate to the example configuration.

policy/approval/predicate.go
danny-wh: Yeah, that works for me. I switched the option and predicate name over, but still used "source branch" to refer to the branch itself since sentences like "From branch did not match..." felt a little clumsy.

Also added the predicate to the list. Thanks for that.

pull/github_test.go
danny-wh: I left this together and added that info to the readme. Seemed like a decent solution to me. Let me know if you feel strongly about splitting them out and I can.

bluekeyes (approved):

PR #251, opened 2020-12-03
Add optional description field to approval rules

### LINKED ISSUE #250: Add an optional description to rules
Some more complicated rules might require a few sentences to describe to users what sort of approvals are required to satisfy the rule. It is possible to use a complicated rule name to describe this, but it's not nice for two reasons
1. The name gets repeated in the `.policy.approval` list. When the long name gets updated because someone wants to reword the approval requirements, they need to remember to make the same change to this list.
2. The UI looks strange with a long name as the large, bold text suggests it was probably designed around the assumption that the name would be short.

### Proposal

Add an optional `description` field to rules which can be used to store the detailed information for users about what is required to get the approval rule to pass. Since this field is optional, the change would be backwards compatible and have no effect for existing configurations of policy-bot.

As an example, consider the following config file:

```yaml
policy:
  approval:
  - Team 1
  - Team 2
  - Admin

approval_rules:
- name: Team 1
  description: Changes to config must be approved by your own team
  requires:
    count: 1
    teams:
    - myorg/team1
  if:
    changed_files:
      paths:
      - '^teams/team1/.*'
- name: Team 2
  description: Changes to config must be approved by your own team
  requires:
    count: 1
    teams:
    - myorg/team2
  if:
    changed_files:
      paths:
      - '^teams/team2/.*'
- name: Admin
  description: All other changes must be approved by an admin
  requires:
    count: 1
    teams:
    - myorg/admin
  if:
    changed_files:
      paths:
      - '.*'
      ignore:
      - '^teams/.*'
```

Which could look something like this:

image

robbiemcmichael:
This change adds an optional description field to approval rules. This description is only displayed when the field is non-empty and the rule has not been skipped.

Closes #250

### DIFF
--- a/server/templates/details.html.tmpl
+++ b/server/templates/details.html.tmpl
@@ -31,7 +31,7 @@
     <div class="status-banner {{$s}} flex flex-wrap items-center justify-between">
       <div>
         <h2 class="mb-1 text-lg font-bold">Status: {{$s | titlecase}}</h2>
-        <p>{{or .Result.Error .Result.Description}}</p>
+        <p>{{or .Result.Error .Result.StatusDescription}}</p>
       </div>
       <div class="p-2 rounded-md bg-white text-dark-gray1 text-sm text-shadow-none">
         <div class="toggle">
@@ -68,5 +68,5 @@
     <b class="font-bold">{{.Name}}</b>
     <span class="flex-none status-badge {{$s}}">{{$s | titlecase}}</span>
   </p>
-  <p class="text-dark-gray3 text-sm">{{or .Error .Description}}</p>
+  <p class="text-dark-gray3 text-sm">{{or .Error .StatusDescription}}</p>
 {{end}}
--- a/server/templates/details.html.tmpl
+++ b/server/templates/details.html.tmpl
@@ -68,5 +68,8 @@
     <b class="font-bold">{{.Name}}</b>
     <span class="flex-none status-badge {{$s}}">{{$s | titlecase}}</span>
   </p>
+  {{if (and .Description (ne $s "skipped"))}}
+  <p class="mb-2 text-dark-gray3 text-sm">{{.Description}}</p>
+  {{end}}
   <p class="text-dark-gray3 text-sm">{{or .Error .StatusDescription}}</p>
 {{end}}
--- a/config/policy-examples/complicated.yml
+++ b/config/policy-examples/complicated.yml
@@ -4,17 +4,19 @@
 policy:
   approval:
   - or:
-    - catskills can approve
-    - devtools can approve changes to staging
+    - catskills
+    - devtools
 
 approval_rules:
 
-- name: catskills can approve
+- name: catskills
+  description: catskills can approve
   requires:
     count: 1
     teams: ["palantir/catskills"]
 
-- name: devtools can approve changes to staging
+- name: devtools
+  description: devtools can approve changes to staging
   if:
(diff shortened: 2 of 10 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @robbiemcmichael! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

server/templates/details.html.tmpl
```diff
@@ -68,5 +68,8 @@
     <b class="font-bold">{{.Name}}</b>
     <span class="flex-none status-badge {{$s}}">{{$s | titlecase}}</span>
   </p>
-  <p class="text-dark-gray3 text-sm">{{or .Error .Description}}</p>
+  {{if (and .Description (ne .Status 0))}}
```
robbiemcmichael: I'm not sure if there's any way to reference the enum by name with Go templates, it's hard to understand that the 0 here means `StatusSkipped`.

server/templates/details.html.tmpl
```diff
@@ -68,5 +68,8 @@
...
+  <p class="text-dark-gray3 text-sm">{{.Description}}</p>
```
robbiemcmichael: We would need to add `margin-top: 0.5em` to `text-dark-gray3` to make it look like the screenshot in #250, otherwise this description and the status description below will be right up against each other.

I couldn't actually find the right CSS file in this repo to be able to make this change myself. I did find [server/assets/css/main.css](https://github.com/palantir/policy-bot/blob/develop/server/assets/css/main.css) but this doesn't seem to be the right file because when I try to use `server/assets` as the static files it really doesn't work well. The only way I could get the CSS to work correctly was to extract the static files out of the Docker image.

I think I must be missing something so I'd appreciate it if you could point me towards the right file.

server/templates/details.html.tmpl
bluekeyes: While I don't think you can reference the constants in the template, you could convert the `.Status` type to a string (e.g. the existing `$s` variable) and then compare the string representations, which might be clearer. The string representations are considered stable since we use them in other parts of the template.

```
{{if (and .Description (eq $s "skipped"))}}
```

server/templates/details.html.tmpl
bluekeyes: Sorry about this, looks like we didn't update the development instructions for working on the UI.

Policy Bot uses [Tailwind](https://tailwindcss.com/) as a CSS framework, so you'll want to add the `mt-2` class to this `p` tag to add the additional top margin.

To test this out locally:

1. Install NodeJS and Yarn if you don't have them already
2. Run `yarn install`
3. Run `yarn run build` to generate the CSS files
4. Uncomment or add the following block to your local server configuration file so it uses the generated CSS files:
   ```yaml
   files:
     static: build/static
     templates: server/templates
   ```

robbiemcmichael:
Thanks for the help @bluekeyes, I've updated this with your suggestions.

bluekeyes (approved):
Looks good, thanks for implementing this! We have some other changes pending release as well, so I'll see if I can make a new release by the end of the week.

dermyrddin:
@bluekeyes was this missed from adding to docs? :)

bluekeyes:
Oops, good catch - fixed in #381.

PR #255, opened 2021-01-04
add note about github apps as users

jmcampanini:
Add a note about referencing GitHub Apps as users with policy-bot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,9 @@ if:
       - "config/.*"
 
   # "has_author_in" is satisfied if the user who opened the pull request is in
-  # the users list or belongs to any of the listed organizations or teams.
+  # the users list or belongs to any of the listed organizations or teams. The
+  # `users` field can contain a GitHub Apps by appending `[bot]` to the end of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,7 +138,9 @@ if:
       - "config/.*"
 
   # "has_author_in" is satisfied if the user who opened the pull request is in
-  # the users list or belongs to any of the listed organizations or teams.
+  # the users list or belongs to any of the listed organizations or teams. The
```
jmcampanini, line 141: this could also be at a higher level, since it may apply in more cases. just adding it here because a user was asking

derekjobst (approved):

README.md
```diff
@@ -138,7 +138,9 @@ if:
...
+  # `users` field can contain a GitHub Apps by appending `[bot]` to the end of
```
```suggestion
  # `users` field can contain a GitHub App by appending `[bot]` to the end of
```

bluekeyes (approved):
This does apply anywhere that the `users` field appears, but just referencing it here for now is probably alright.

PR #256, opened 2021-01-12
Add support for enforcing regex rules on PR Title

### LINKED ISSUE #244: feature request: Check on the PR title
Have an approval and denial rule that checks the PR title against a regex.

Like this you can enforce only PRs with valid commit messages can be merged (e.g. https://www.conventionalcommits.org/en/v1.0.0/#summary)

Github hooks already trigger on renaming a PR.

derekjobst:
This PR includes changes which add functionality for disapproving pull requests which do not comply by defined title formatting requirements as mentioned in https://github.com/palantir/policy-bot/issues/244.

- A new predicate `title` is added for defining allowed (`not_match`) and disallowed (`match`) regex patterns on a pull request title.
- The `disapproval` policy is extended to allow predicates just as individual `approval_rules` do. However, whereas an approval rule may only allow approvals subject to passing predicates, the `disapproval` policy will only allow disapprovals subject to its own predicates all failing. Passing predicates on the `disapproval` policy will trigger a default disapproval, just as failing predicates on an approval rule will implicitly approve (pass).

An example of the new configuration:
```yaml
policy:
  disapproval:
    if:
      title: 
        not_matches:
          - "^(fix|feat|chore): (\\w| )+$"
          - "^(chore|docs|style|refactor|perf|test): (\\w| )+$"
          - "^feat\\(\\w+\\): (\\w| )+$"
          - "^BREAKING CHANGE: (\\w| )+$"
        matches:
          - "^BLOCKED"
```
This policy will disapprove PR's with a title that does not match one of the patterns in the `not_matches` list (conforming to the conventional commits spec) or if there are any matches in the `matches` pattern list, in this case anything that begins with the text `BLOCKED`.


policy/common/regexp.go
```diff
@@ -68,3 +68,12 @@ func (r *Regexp) UnmarshalJSON(data []byte) (err error) {
 	*r, err = NewRegexp(pattern)
 	return err
 }
+
+func AnyMatches(re []Regexp, s string) bool {
```
bluekeyes: I'm not convinced this needed to move to the `common` package, since it is still only used by predicate implementations. However, moving it to a file like `matches.go` in the `predicate` package probably makes sense now that it is used in more than one predicate.

policy/disapproval/disapprove.go
```diff
@@ -96,6 +98,33 @@ func (p *Policy) Evaluate(ctx context.Context, prctx pull.Context) (res common.R
...
+
+		if satisfied {
+			log.Debug().Msgf("disapproving, predicate of type %T was satisfied", p)
+			disapproved = true
+
+			if desc == "" {
+				desc = "One of the preconditions of this rule is satisfied"
+			}
+
+			msgs += desc + "\n"
```
bluekeyes: In most parts of policy-bot, evaluation stops as soon as we have the answer we need. Applying that here, I was expecting this to return immediately once a predicate was satisfied.

However, I can see some benefit in showing all the reasons for disapproval at once so people don't have to iterate as much on fixes, although I don't know in practice how often there will be (a) multiple disapproval predicates and (b) a PR that trips more than one of them.

If we do keep the multiple messages, my memory of the UI suggests it doesn't render new lines, so this will appear as a bunch of sentences. I think if we want to include all messages, it's worth rendering them in a good format, either as a list or with the line breaks.

policy/predicate/predicates.go
```diff
@@ -0,0 +1,85 @@
...
+	TargetsBranch *TargetsBranch `yaml:"targets_branch"`
+	FromBranch    *FromBranch    `yaml:"from_branch"`
+
+	ModifiedLines *ModifiedLines `yaml:"modified_lines"`
+
+	HasSuccessfulStatus *HasSuccessfulStatus `yaml:"has_successful_status"`
+
+	HasLabels *HasLabels `yaml:"has_labels"`
+
+	MatchesTitlePattern *MatchesTitlePattern `yaml:"title"`
```
bluekeyes: style: I suggest calling this `Title` so it matches the YAML key.

policy/predicate/title.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/rs/zerolog"
+
+	"github.com/palantir/policy-bot/policy/common"
+	"github.com/palantir/policy-bot/pull"
+)
+
+type MatchesTitlePattern struct {
+	TitleMatches    []common.Regexp `yaml:"matches"`
+	TitleNotMatches []common.Regexp `yaml:"not_matches"`
```
bluekeyes: style: these can be `Matches` and `NotMatches` since the "title" part is clear from the the type they are part of.

policy/predicate/title.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+var _ Predicate = MatchesTitlePattern{}
+
+func (pred MatchesTitlePattern) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+	title := prctx.Title()
+
+	log := zerolog.Ctx(ctx)
+	log.Debug().Msgf("PR title: %s", title)
```
bluekeyes: This seems like it won't help too much in practice, since it's pretty easy to check the title given the PR number if we need to debug something.

policy/predicate/title.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	return false, fmt.Sprintf("PR title: \"%s\" is satisfactory", title), nil
+}
+
+func (pred MatchesTitlePattern) Trigger() common.Trigger {
+	return common.TriggerPullRequest
+}
+Adds support for predicates in disapproval policy
+Adds MatchesTitlePattern predicate
```
bluekeyes:
```suggestion
```

policy/predicate/title.go
```diff
@@ -0,0 +1,55 @@
...
+var _ Predicate = MatchesTitlePattern{}
+
+func (pred MatchesTitlePattern) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+	title := prctx.Title()
+
+	log := zerolog.Ctx(ctx)
+	log.Debug().Msgf("PR title: %s", title)
+
+	if !common.AnyMatches(pred.TitleMatches, title) {
+		return true, fmt.Sprintf("PR title: \"%s\" doesn't match a required pattern", title), nil
```
bluekeyes: If you look at other predicate messages, they're usually either sentences or put the colon at the end. Maybe `Title doesn't match a required pattern` is enough? Similar to the log, I'm not sure how useful it is to include the title in the message.

pull/github.go
```diff
@@ -730,6 +735,7 @@ func backfillPushedAt(commits []*Commit, headSHA string) {
 // if adding new fields to this struct, modify Locator#toV4() as well
 type v4PullRequest struct {
 	Author    v4Actor
+	Title     string
```
bluekeyes, line 739: While I forgot to mention it in the comment here, you also need to modify `Locator.IsComplete` when adding new `v4PullRequest` fields.

bluekeyes (commented):
This looks like a good start. In addition to the tests you mentioned, we'll want to document this new predicate and the new disapproval feature in the README.

policy/common/regexp.go
derekjobst: Totally agree, done!

policy/predicate/title.go
derekjobst: Agreed. Perhaps the specific regex that matched could be helpful though?

policy/disapproval/disapprove.go
derekjobst: Fixed, and noted in issue: https://github.com/palantir/policy-bot/issues/258

policy/disapproval/disapprove.go
```diff
@@ -96,6 +98,26 @@ func (p *Policy) Evaluate(ctx context.Context, prctx pull.Context) (res common.R
...
 
+	for _, p := range p.Predicates.Predicates() {
+		satisfied, desc, err := p.Evaluate(ctx, prctx)
+
+		if err != nil {
+			res.Error = errors.Wrap(err, "failed to evaluate predicate")
+			return
+		}
+
+		if !satisfied {
```
bluekeyes: I think we want to disapprove when predicates are satisfied, rather than when they are not satisfied. See my comment on the documentation for more.

policy/predicate/matches.go
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package predicate
+
+import (
+	"github.com/palantir/policy-bot/policy/common"
+)
+
+func AnyMatches(re []common.Regexp, s string) bool {
```
bluekeyes: style: this is minor since this package is effectively internal, but this should probably be unexported (`anyMatches`)

policy/predicate/title_test.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func TestWithMixedRules(t *testing.T) {
+	p := &Title{
+		Matches: []common.Regexp{
+			common.NewCompiledRegexp(regexp.MustCompile("^[A-Z]*$")),
+		},
+		NotMatches: []common.Regexp{
+			common.NewCompiledRegexp(regexp.MustCompile("^[a-z]*$")),
+		},
```
bluekeyes: I think it would be good to have a test that more explicitly demonstrates the "any matches and no not matches" behavior. The empty string test kind of covers this, since it matches both test patterns, but an empty string is also a common special case, so it's not obvious which part is being tested.

pull/context.go
```diff
@@ -56,6 +56,9 @@ type Context interface {
 	// Number returns the number of the pull request.
 	Number() int
 
+	//Title returns the title of the pull request
```
bluekeyes:
```suggestion
	// Title returns the title of the pull request
```

pull/github.go
```diff
@@ -727,9 +733,10 @@ func backfillPushedAt(commits []*Commit, headSHA string) {
 	}
 }
 
-// if adding new fields to this struct, modify Locator#toV4() as well
+// if adding new fields to this struct, modify Locator#toV4() and Locator#IsComplete() as well
```
bluekeyes, line 736: Thanks for updating this!

server/templates/details.html.tmpl
```diff
@@ -72,4 +72,5 @@
   <p class="mb-2 text-dark-gray3 text-sm">{{.Description}}</p>
   {{end}}
   <p class="text-dark-gray3 text-sm">{{or .Error .StatusDescription}}</p>
+
```
bluekeyes: Left over from other changes?

```suggestion
```

README.md
```diff
@@ -325,19 +334,35 @@ Conjunctions can contain more conjunctions (up to a maximum depth of 5):
...
+policy has the following specification:
 
 ```yaml
 # "disapproval" is the top-level key in the policy block.
 disapproval:
+  # "if" specifies a set of predicates that must be true for the policy to apply.
+  # Otherwise, the policy will fail and the pull request will be diapproved.
+  #  
+  # This block, and every condition within it are optional. If the block does not
+  # exist, the rule applies to every pull request.
```
bluekeyes: This feels reversed to me. I read the example policy as "disapprove if title matches `patternMatches` and does not match`patternNotMatches`". But the implementation is actually "disapprove if title does not match `patternMatches` and matches `patternNotMatches`".

Do you have some examples with different predicates? Maybe the issue is using `if` as the name of this field rather than the behavior.

Additionally, I think the lines about the rule "applying" are misleading. But I tried to explain why and it was complicated by the previous issue about how the predicates act. Let's sort that out and then I'll see if my comment about the phrasing is still relevant.

README.md
```diff
@@ -325,19 +334,35 @@ Conjunctions can contain more conjunctions (up to a maximum depth of 5):
...
+  if:
+    # "title" is satisfied if the pull request title matches any one of the
+    # patterns within the "matches" list, and does not match all of the patterns
+    # within the "not_matches" list.
```
bluekeyes: I don't think we need to duplicate this predicate level documentation

README.md
```diff
@@ -325,19 +334,35 @@ Conjunctions can contain more conjunctions (up to a maximum depth of 5):
...
+    title:
+      matches:
+        - "regexPattern"
+      not_matches:
+        - "regexPattern"
+    # Any additional predicates from the approval rules section are also valid here
```
bluekeyes: I'd put this in either the documentation for the `if` block or before the first example predicate, rather than after.

policy/predicate/title_test.go
derekjobst: I think the updated test may address this, but I will take another look to be sure:
https://github.com/palantir/policy-bot/pull/256/files#diff-06aab9202668ad4ba029dbd445d710fb1daddc042cd3d6fd2cb1b1fe025e7dc7R97-R148

pull/github.go
derekjobst: For sure!

policy/predicate/title_test.go
derekjobst: I think this is addressed with the config:
```golang
NotMatches: []common.Regexp{
	common.NewCompiledRegexp(regexp.MustCompile("^(fix|feat|chore): (\\w| )+$")),
	common.NewCompiledRegexp(regexp.MustCompile("^BREAKING CHANGE: (\\w| )+$")),
},
Matches: []common.Regexp{
	common.NewCompiledRegexp(regexp.MustCompile("BLOCKED")),
},
```
And the spec:
```golang
{
	"matches pattern in both lists",
	true,
	&pulltest.Context{
		TitleValue: "BREAKING CHANGE: BLOCKED",
	},
},
```

README.md
derekjobst: Yes, you are right. Logic updated.

I've been thinking things through with the labels predicate. e.g., if the label "BLOCKED" is added, that predicate will evaluate true, and if it is in the disapproval policy the PR will be disapproved.

Updated this to:
```
  # "if" specifies a set of predicates which will cause disapproval if any are 
  # true
  #  
  # This block, and every condition within it are optional. If the block does 
  # not exist, a pull request is only disapproved if a user takes a disapproval 
  # action.
```

policy/disapproval/disapprove.go
```diff
@@ -96,6 +98,26 @@ func (p *Policy) Evaluate(ctx context.Context, prctx pull.Context) (res common.R
...
+	for _, p := range p.Predicates.Predicates() {
+		satisfied, desc, err := p.Evaluate(ctx, prctx)
+
+		if err != nil {
+			res.Error = errors.Wrap(err, "failed to evaluate predicate")
+			return
+		}
+
+		if satisfied {
+			log.Debug().Msgf("disapproving, predicate of type %T was not satisfied", p)
```
bluekeyes:
```suggestion
			log.Debug().Msgf("disapproving, predicate of type %T was satisfied", p)
```

policy/disapproval/disapprove.go
```diff
@@ -96,6 +98,26 @@ func (p *Policy) Evaluate(ctx context.Context, prctx pull.Context) (res common.R
...
+
+			res.Status = common.StatusDisapproved
+			res.StatusDescription = desc
+			if desc == "" {
+				res.StatusDescription = "The preconditions of this rule are not satisfied"
```
bluekeyes:
```suggestion
				res.StatusDescription = "A precondition of this rule was satisfied"
```

bluekeyes (approved):
Two minor message changes that I think were missed when fixing up the conditions, but otherwise this looks good to me.

derekjobst:
Fixed minor issue causing disapproval predicates from updating status checks on their own.

PR #257, opened 2021-01-12
Add support for handling review comments

### LINKED ISSUE #51: FR: Handle approvals embedded in GHE review comments
Sometimes a user will make a slightly weird, but otherwise correct review, and we'll see something like 

User1 reviewed with a comment:
```
 👍 
```

Policybot looks at the comment text and the result of the github approval, but not the _text_ of the github approval. I believe this is covered in the slightly different [reviews api](https://developer.github.com/v3/pulls/reviews/), so these types of "approvals" aren't counted.

jgiannuzzi:
This PR adds support for handling review comments like regular comments, thus allowing users to approve or disapprove without switching back to the "Conversation" tab of the pull request whilst reviewing the changes.

Fixes #51.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -519,7 +519,7 @@ func (ghc *GitHubContext) loadPagedData() error {
 				Reviews struct {
 					PageInfo v4PageInfo
 					Nodes    []v4PullRequestReview
-				} `graphql:"reviews(first: 100, after: $reviewCursor, states: [APPROVED, CHANGES_REQUESTED])"`
+				} `graphql:"reviews(first: 100, after: $reviewCursor, states: [APPROVED, CHANGES_REQUESTED, COMMENTED])"`
 			} `graphql:"pullRequest(number: $number)"`
 		} `graphql:"repository(owner: $owner, name: $name)"`
 	}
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @jgiannuzzi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

pull/github.go
```diff
@@ -548,7 +548,12 @@ func (ghc *GitHubContext) loadPagedData() error {
 		}
 
 		for _, r := range q.Repository.PullRequest.Reviews.Nodes {
-			reviews = append(reviews, r.ToReview())
+			switch r.State {
+			case "COMMENTED":
```
jgiannuzzi, line 552: Maybe this should be `case string(githubv4.PullRequestReviewStateCommented):` instead?

I did not find any instance of using the [`githubv4`](https://github.com/shurcooL/githubv4) module constants anywhere in the code, and thus decided to use these strings instead, but happy to change that if you'd like.

pull/github.go
```diff
@@ -548,7 +548,12 @@ func (ghc *GitHubContext) loadPagedData() error {
...
+				comments = append(comments, r.ToComment())
+			case "APPROVED", "CHANGES_REQUESTED":
```
jgiannuzzi, line 554: Same comment as above.

pull/github.go
bluekeyes: I think it's fine to keep these as strings for now

bluekeyes (approved):
Thanks for implementing this and updating the tests!

bluekeyes:
Hopefully we can include this in a release by the end of the week; there's one other feature currently in progress that I'd like to complete first. For now, you can test with the `snapshot` Docker tag if you want.

jgiannuzzi:
That would be awesome, thanks!

PR #261, opened 2021-01-21
Revert back to go-github v32

bluekeyes:
In v33, the library removed some preview media type headers that are still required in GitHub Enterprise, although they are no longer required on GitHub.com. Until there's a GHE release that drops the requirement for the preview media types, we'll need to stick with the older version.

Since we otherwise weren't taking advantage of new features or changes, this should be fine.

The Checks API headers removed in https://github.com/google/go-github/pull/1636 are the specific ones we care about.


IainSteers (approved):

asvoboda:
The header has graduated as part of GHE 3.0: https://docs.github.com/en/enterprise-server@3.0/rest/reference/checks#create-a-check-run--parameters, so we should be safe to update once again.

PR #265, opened 2021-02-03
Update status check on details page evaluation

derekjobst:
This PR includes changes which enhance the user experience when commits are missing in the `GitHubContext.loadPushedAt` method. This seeks to improve situations where commits cannot be found, leading to the status check resolving to an error:
```
Error evaluating policy defined by org/project ref=branch
```
### Changes
1. Loading the details page now posts updated status checks to GitHub.
    - This allows users to troubleshoot errors related to incomplete Github API responses, a likely cause for the previously mentioned errors.
    - The details page will no longer be more "up to date" than the GitHub status.
2. The error message/log for missing commits is updated with a troubleshooting tip and a list of missing commit SHAs. 
    - The error itself is expanded: 
`Error evaluating policy defined by org/project ref=branch. Missing: <list of commits>`

    - A new temporary error troubleshooting element is added to the UI: Screen Shot 2021-02-08 at 12 09 50 PM

Issue reports have indicated this troubleshooting method can be successful.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -706,7 +706,11 @@ func (ghc *GitHubContext) loadPushedAt(commits []*Commit) error {
 	}
 
 	if len(commitsBySHA) > 0 {
-		return errors.Errorf("%d commits were not found while loading pushed dates", len(commitsBySHA))
+		missingSHAs := make([]string, 0, len(commitsBySHA))
+		for sha := range commitsBySHA {
+			missingSHAs = append(missingSHAs, sha)
+		}
+		return errors.Errorf("%d commits were not found while loading pushed dates. Missing: %s", len(commitsBySHA), strings.Join(missingSHAs, ", "))
 	}
 	return nil
 }
--- a/pull/github.go
+++ b/pull/github.go
@@ -16,6 +16,7 @@ package pull
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 	"time"
@@ -44,6 +45,14 @@ type Locator struct {
 	Number int
 
 	Value *github.PullRequest
+}
+
+type TemporaryError struct{
+	error string
+}
+
+func (te *TemporaryError) Error() string {
+	return te.error
 }
 
 // IsComplete returns true if the locator contains a pull request object with
@@ -710,8 +719,10 @@ func (ghc *GitHubContext) loadPushedAt(commits []*Commit) error {
 		for sha := range commitsBySHA {
 			missingSHAs = append(missingSHAs, sha)
 		}
-		return errors.Errorf("%d commits were not found while loading pushed dates. Missing %s. This may be temporary. Wait 30 seconds, and refresh this page to re-evaluate.",
-			len(commitsBySHA), strings.Join(missingSHAs, ", "))
+
+		err := &TemporaryError{fmt.Sprintf("%d commits were not found while loading pushed dates. Missing %s.",
+			len(commitsBySHA), strings.Join(missingSHAs, ", "))}
+		return err
 	}
 	return nil
 }
--- a/pull/github.go
+++ b/pull/github.go
@@ -710,7 +710,8 @@ func (ghc *GitHubContext) loadPushedAt(commits []*Commit) error {
 		for sha := range commitsBySHA {
 			missingSHAs = append(missingSHAs, sha)
 		}
-		return errors.Errorf("%d commits were not found while loading pushed dates. Missing %s", len(commitsBySHA), strings.Join(missingSHAs, ", "))
+		return errors.Errorf("%d commits were not found while loading pushed dates. Missing %s. This may be temporary. Wait, and refresh this page to re-evaluate.",
+			len(commitsBySHA), strings.Join(missingSHAs, ", "))
 	}
 	return nil
 }
--- a/pull/github.go
+++ b/pull/github.go
@@ -710,7 +710,7 @@ func (ghc *GitHubContext) loadPushedAt(commits []*Commit) error {
 		for sha := range commitsBySHA {
 			missingSHAs = append(missingSHAs, sha)
 		}
-		return errors.Errorf("%d commits were not found while loading pushed dates. Missing %s. This may be temporary. Wait, and refresh this page to re-evaluate.",
+		return errors.Errorf("%d commits were not found while loading pushed dates. Missing %s. This may be temporary. Wait 30 seconds, and refresh this page to re-evaluate.",
 			len(commitsBySHA), strings.Join(missingSHAs, ", "))
 	}
 	return nil
--- a/pull/github.go
+++ b/pull/github.go
@@ -710,7 +710,7 @@ func (ghc *GitHubContext) loadPushedAt(commits []*Commit) error {
 		for sha := range commitsBySHA {
 			missingSHAs = append(missingSHAs, sha)
 		}
-		return errors.Errorf("%d commits were not found while loading pushed dates. Missing: %s", len(commitsBySHA), strings.Join(missingSHAs, ", "))
+		return errors.Errorf("%d commits were not found while loading pushed dates. Missing %s", len(commitsBySHA), strings.Join(missingSHAs, ", "))
 	}
 	return nil
 }
--- a/pull/github.go
+++ b/pull/github.go
@@ -617,7 +617,7 @@ func (ghc *GitHubContext) loadCommits() ([]*Commit, error) {
 	// In the second case, retrying after a delay can fix things, but the delay
 	// can be 15+ seconds in practice, so using the alternate API should
 	// improve latency at the cost of more API requests.
-	if head.PushedAt == nil {
+	//if head.PushedAt == nil {
 		log.Debug().
 			Bool("fork", ghc.pr.IsCrossRepository).
 			Msgf("failed to load pushed date via pull request, falling back to commit APIs")
@@ -625,7 +625,7 @@ func (ghc *GitHubContext) loadCommits() ([]*Commit, error) {
 		if err := ghc.loadPushedAt(commits); err != nil {
 			return nil, err
 		}
-	}
+	//}
 
 	if head.PushedAt == nil {
 		return nil, errors.Errorf("head commit %.10s is missing pushed date; this is probably a bug", ghc.pr.HeadRefOID)
@@ -670,49 +670,49 @@ func (ghc *GitHubContext) loadPushedAt(commits []*Commit) error {
 		commitsBySHA[c.SHA] = c
 	}
 
-	var q struct {
-		Repository struct {
-			Object struct {
-				Commit struct {
-					History struct {
-						PageInfo v4PageInfo
-						Nodes    []struct {
-							OID        string
-							PushedDate *time.Time
-						}
-					} `graphql:"history(first: 100, after: $cursor)"`
-				} `graphql:"... on Commit"`
-			} `graphql:"object(oid: $oid)"`
-		} `graphql:"repository(owner: $owner, name: $name)"`
-	}
-	qvars := map[string]interface{}{
-		"owner":  githubv4.String(ghc.pr.HeadRepository.Owner.Login),
-		"name":   githubv4.String(ghc.pr.HeadRepository.Name),
-		"oid":    githubv4.GitObjectID(ghc.pr.HeadRefOID),
-		"cursor": (*githubv4.String)(nil),
-	}
-
-	loaded := 0
-	for {
-		if err := ghc.v4client.Query(ghc.ctx, &q, qvars); err != nil {
-			return errors.Wrap(err, "failed to load commit pushed dates")
-		}
-		for _, n := range q.Repository.Object.Commit.History.Nodes {
-			if c, ok := commitsBySHA[n.OID]; ok {
-				c.PushedAt = n.PushedDate
-				delete(commitsBySHA, n.OID)
-			}
-		}
-
-		loaded += len(q.Repository.Object.Commit.History.Nodes)
-		if loaded > len(commits) {
-			break
-		}
-
-		if !q.Repository.Object.Commit.History.PageInfo.UpdateCursor(qvars, "cursor") {
-			break
-		}
-	}
+	//var q struct {
+	//	Repository struct {
+	//		Object struct {
+	//			Commit struct {
+	//				History struct {
+	//					PageInfo v4PageInfo
+	//					Nodes    []struct {
+	//						OID        string
+	//						PushedDate *time.Time
+	//					}
+	//				} `graphql:"history(first: 100, after: $cursor)"`
+	//			} `graphql:"... on Commit"`
+	//		} `graphql:"object(oid: $oid)"`
+	//	} `graphql:"repository(owner: $owner, name: $name)"`
+	//}
+	//qvars := map[string]interface{}{
+	//	"owner":  githubv4.String(ghc.pr.HeadRepository.Owner.Login),
+	//	"name":   githubv4.String(ghc.pr.HeadRepository.Name),
+	//	"oid":    githubv4.GitObjectID(ghc.pr.HeadRefOID),
+	//	"cursor": (*githubv4.String)(nil),
+	//}
+	//
+	//loaded := 0
+	//for {
+	//	if err := ghc.v4client.Query(ghc.ctx, &q, qvars); err != nil {
+	//		return errors.Wrap(err, "failed to load commit pushed dates")
+	//	}
+	//	for _, n := range q.Repository.Object.Commit.History.Nodes {
+	//		if c, ok := commitsBySHA[n.OID]; ok {
+	//			c.PushedAt = n.PushedDate
+	//			delete(commitsBySHA, n.OID)
+	//		}
+	//	}
+	//
+	//	loaded += len(q.Repository.Object.Commit.History.Nodes)
+	//	if loaded > len(commits) {
+	//		break
+	//	}
+	//
+	//	if !q.Repository.Object.Commit.History.PageInfo.UpdateCursor(qvars, "cursor") {
+	//		break
+	//	}
+	//}
 
 	if len(commitsBySHA) > 0 {
 		missingSHAs := make([]string, 0, len(commitsBySHA))
--- a/pull/github.go
+++ b/pull/github.go
@@ -47,7 +47,7 @@ type Locator struct {
 	Value *github.PullRequest
 }
 
-type TemporaryError struct{
+type TemporaryError struct {
 	error string
 }
 
--- a/pull/github.go
+++ b/pull/github.go
@@ -617,7 +617,7 @@ func (ghc *GitHubContext) loadCommits() ([]*Commit, error) {
 	// In the second case, retrying after a delay can fix things, but the delay
 	// can be 15+ seconds in practice, so using the alternate API should
 	// improve latency at the cost of more API requests.
-	//if head.PushedAt == nil {
+	if head.PushedAt == nil {
 		log.Debug().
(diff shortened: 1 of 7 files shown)


server/handler/base.go
```diff
@@ -185,7 +185,7 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
 
 	result := evaluator.Evaluate(ctx, prctx)
 	if result.Error != nil {
-		statusMessage := fmt.Sprintf("Error evaluating policy defined by %s", fetchedConfig)
+		statusMessage := fmt.Sprintf("Error evaluating policy defined by %s. This may be temporary. Create a PR comment to try again.", fetchedConfig)
```
bluekeyes: I'm not sure this will actually show up in a useful way in the GitHub UI... we'd have to see where the truncation happens. Originally, I was thinking this would be part of the error banner in the details view, rather than in the status message. That way, if someone clicked through to see why the check failed, there would be a tip on how to resolve it.

That also raises the point that the details view does a fresh evaluation but does not update status checks, so you can get into a situation where by the time you check why something failed, it's not failing anymore. Maybe having the details view update status is another way to help here?

server/handler/base.go
```diff
@@ -146,32 +146,43 @@ func (b *Base) Evaluate(ctx context.Context, installationID int64, trigger commo
...
-		return errors.WithMessage(err, fmt.Sprintf("failed to fetch policy: %s", fetchedConfig))
+		return errors.WithMessage(err, fmt.Sprintf("Failed to fetch configuration: %s", fetchedConfig))
 	}
 
-	return b.EvaluateFetchedConfig(ctx, prctx, client, fetchedConfig, trigger)
+	evaluator, err := b.ValidateFetchedConfig(ctx, prctx, client, fetchedConfig, trigger)
+
+	if evaluator == nil {
+		return err
+	}
```
derekjobst, line 168: Note: Logic for generating `err` is now slightly different. This should not affect functionality, but logs will be more detailed. Missing configs will now generate an additional log statement when their errors are handled by the caller. 

The specific changes are:
- When the config is **missing**: `nil` used to be returned, now an error with the config description (missing owner/repo/ref) is returned. 
  - Like before, the PR status is not updated so the missing config is effectively ignored here. The details page renders this error, which is why it is now returned.
- When the config is **invalid**: The config error is now wrapped with the config description (invalid owner/repo/ref).
  - The PR status message is unchanged. The enhanced error allows for more context on the details page.
- When the **evaluation** throws an error: That error is now wrapped with a message mentioning the policy could not be parsed.
  - Same impact as invalid.

server/handler/fetcher.go
```diff
@@ -58,11 +58,11 @@ func (fc FetchedConfig) String() string {
...
 	switch {
 	case fc.Missing():
-		return fmt.Sprintf("No policy found at ref=%s", fc.Ref)
+		return fmt.Sprintf("No policy found at %s", fc)
 	case fc.Invalid():
-		return fmt.Sprintf("Invalid configuration defined by ref=%s", fc.Ref)
+		return fmt.Sprintf("Invalid configuration defined by %s", fc)
 	}
-	return fmt.Sprintf("Valid policy found for ref=%s", fc.Ref)
+	return fmt.Sprintf("Valid policy found for %s", fc)
```
derekjobst, line 65: The `String` method is more detailed than just a `ref`, and we seem to be a bit inconsistent in using the two formats. I think in all cases owner/repo/ref would be preferable to just ref, correct?

pull/github.go
```diff
@@ -706,7 +706,12 @@ func (ghc *GitHubContext) loadPushedAt(commits []*Commit) error {
 	}
 
 	if len(commitsBySHA) > 0 {
-		return errors.Errorf("%d commits were not found while loading pushed dates", len(commitsBySHA))
+		missingSHAs := make([]string, 0, len(commitsBySHA))
+		for sha := range commitsBySHA {
+			missingSHAs = append(missingSHAs, sha)
+		}
+		return errors.Errorf("%d commits were not found while loading pushed dates. Missing %s. This may be temporary. Wait 30 seconds, and refresh this page to re-evaluate.",
```
bluekeyes: style:
```suggestion
		return errors.Errorf("%d commits were not found while loading pushed dates. Missing %s. This may be temporary. Wait 30 seconds then refresh this page to re-evaluate.",
```

I _think_ this will only appear on the details page (and in logs), so it's probably fine to include instructions like this here, but I think in an ideal implementation, this would mark the error as temporary in some way and only the frontend rendering code would add instructions about what to do for temporary errors.

This is an example of where having errors implement an interface or be a specific type could be helpful. On the other hand, retrying is usually a good strategy, so we could probably get 80% of the value by just adding a line in the UI that always suggests retrying on error.

server/handler/base.go
bluekeyes: > When the config is missing: nil used to be returned, now an error with the config description (missing owner/repo/ref) is returned.

I think we'll want to either leave this as a `nil` error or mark it in some way that can be ignored by things other than the detail view. If I'm following the code correctly, while this won't result in a failed status, it will still appear as an unexpected error in logs and metrics.

> When the evaluation throws an error: That error is now wrapped with a message mentioning the policy could not be parsed.

I don't see this behavior in the implementation, but I did want to mention that evaluation errors can only happen after successful parsing.

server/handler/base.go
```diff
@@ -146,32 +146,43 @@ func (b *Base) Evaluate(ctx context.Context, installationID int64, trigger commo
...
 	fetchedConfig, err := b.ConfigFetcher.ConfigForPR(ctx, prctx, client)
 	if err != nil {
-		return errors.WithMessage(err, fmt.Sprintf("failed to fetch policy: %s", fetchedConfig))
+		return errors.WithMessage(err, fmt.Sprintf("Failed to fetch configuration: %s", fetchedConfig))
 	}
 
-	return b.EvaluateFetchedConfig(ctx, prctx, client, fetchedConfig, trigger)
+	evaluator, err := b.ValidateFetchedConfig(ctx, prctx, client, fetchedConfig, trigger)
+
+	if evaluator == nil {
```
bluekeyes, line 161: This pattern reads a bit oddly to me. I think I'd prefer to see something like:

```go
if err != nil {
  return err
}
if evaluator == nil {
  return nil
}
```

I think that more clearly communicates that one of the "success" cases is when the function returns a nil evaluator.

server/handler/base.go
```diff
@@ -146,32 +146,43 @@ func (b *Base) Evaluate(ctx context.Context, installationID int64, trigger commo
...
+
+		return nil, errors.New(fetchedConfig.Description())
 	}
 
 	if fetchedConfig.Invalid() {
-		logger.Warn().Err(fetchedConfig.Error).Msgf("invalid policy: %s", fetchedConfig)
-		err := b.PostStatus(ctx, prctx, client, "error", fetchedConfig.Description())
-		return err
+		logger.Warn().Err(fetchedConfig.Error).Msg(fetchedConfig.Description())
+		b.PostStatus(ctx, prctx, client, "error", fetchedConfig.Description())
```
bluekeyes: You mentioned that we're no longer handling the error from `PostStatus`. I think we do need to do something here, but I agree that not returning the error makes it easier to return the other, more meaningful errors. 

Maybe we can change `PostStatus` to not return `error` and to always log the error instead? In general, there's not much we can do if we fail to post a status and there shouldn't be much harm in finishing out any remaining work, in the few cases where posting the status isn't the last step.

server/handler/base.go
```diff
@@ -204,14 +236,20 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
...
+
+func (b *Base) RequestReviewsForResult(ctx context.Context, prctx pull.Context, client *github.Client, result common.Result) error {
+	logger := zerolog.Ctx(ctx)
+
+	if prctx.IsDraft() {
+		return nil
 	}
 
-	if statusState == "pending" && !prctx.IsDraft() {
+	if result.Status == common.StatusPending {
```
bluekeyes: You could probably combine this with the new draft check and return early, unnesting the rest of this function:

```go
if prctx.IsDraft() || result.Status != common.StatusPending {
  return nil
}
```

server/handler/base.go
```diff
@@ -180,18 +191,39 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
...
+		b.PostStatus(ctx, prctx, client, "error", statusMessage)
+
+		return result, result.Error
+	}
+
+	if err := b.PostStatusForResult(ctx, prctx, client, result); err != nil {
+		return result, err
+	}
+
+	if err := b.RequestReviewsForResult(ctx, prctx, client, result); err != nil {
```
bluekeyes: I'm not sure we want to request reviewers when loading the details page. Unlike posting statuses, review request have a side-effects and the design relies to some extent on limiting the number of cases when the code runs.

There's also a random delay in the code when requesting reviewers to try to space out events. Since events are asynchronous this is fine, but in the context of the details page, it will add latency.

server/handler/details.go
```diff
@@ -121,31 +120,20 @@ func (h *Details) ServeHTTP(w http.ResponseWriter, r *http.Request) error {
...
-	}
-
-	evaluator, err := policy.ParsePolicy(config.Config)
-	if err != nil {
-		data.Error = errors.WithMessage(err, fmt.Sprintf("invalid policy at ref \"%s\"", config.Ref))
-		return h.render(w, data)
-	}
-
-	result := evaluator.Evaluate(ctx, prctx)
+	result, err := h.Base.EvaluateFetchedConfig(ctx, prctx, client, evaluator, config)
```
bluekeyes, line 141: Given the comment above about review requests, I think we should look at inlining `EvaluateFetchedConfig` here and accept the duplication so we can more easily drop parts we don't want. The introduction of `ValidateFetchedConfig` helps remove some of the existing duplication, so maybe that's enough.

server/handler/fetcher.go
bluekeyes: Yeah, that sounds fine. I suspect the original use of only `ref` was acceptable because this always appears in the context of a log message or UI page that is already annotated with information about the repository.

server/handler/base.go
derekjobst: Sounds good. I was a little unsure about adding this to details rendering, so the context here is helpful. Perhaps we want to add something to that UI note about how to trigger reviewer requests on failed builds? (i.e. add a comment or commit to the PR)

server/handler/details.go
derekjobst: I moved the success case status update logic back into `EvaluateFetchedConfig`, and it feels more like a self contained block of logic now. Errors are thrown when evaluation produces an error or validation of the error status has a problem. 

The callers of `EvaluateFetchedConfig` handle the err/result in their own ways, but don't seem to need more customization of what actually occurs in this method. Therefore, I think we can probably keep it as it's own method. Does that sound ok?

server/handler/base.go
derekjobst: Ok, sounds good.

pull/github.go
derekjobst: Ok, moved this to use a `TemporaryError`. I just added the troubleshooting note to the error message, but we could also add a separate UI element if you think that might be better. Perhaps a `data.ErrorTroubleshooting` passed to render?

server/handler/issue_comment.go
```diff
@@ -88,20 +88,33 @@ func (h *IssueComment) Handle(ctx context.Context, eventType, deliveryID string,
 	}
 
 	if fetchedConfig.Valid() {
-		tampered, err := h.detectAndLogTampering(ctx, prctx, client, event, fetchedConfig.Config)
-		if err != nil {
-			return errors.Wrap(err, "failed to detect tampering")
```
derekjobst, line 93: Hopefully this change does not reduce the effectiveness of this tampering check.

I believe this `err` clause would only catch the case where POST-ing the updated failure status check to github failed, since every other return provides `nil` for `err`. Therefore, I don't think this degrades the effectiveness of detectAndLogTampering. Let me know if I'm missing something. Perhaps there is a check somewhere on the policy-bot logs looking for "failed to detect tampering"?

If we want to keep this error, PostStatus can be changed back to returning errors, and they can be processed by the callers of PostStatus.

server/handler/base.go
derekjobst: Ok, sounds good. Changed it so that missing configs no longer return an error. The details view will now generate it's own error when `evaluator` is `nil`, whereas base/issue_comment callers don't care (since they just skip missing configs).

Sorry, I meant to say when `ParsePolicy` is called to get the evaluator and an error is produced we now wrap that error with a [message](https://github.com/palantir/policy-bot/pull/265/files#diff-24d63556b451fd44ed18a6ea7f625e95bc19c861c7988f1f81a6272fc18d0eb7R200) mentioning the policy could not be parsed (line 200 of this file)

server/handler/base.go
```diff
@@ -98,17 +100,18 @@ func (b *Base) PostStatus(ctx context.Context, prctx pull.Context, client *githu
 	}
 
 	if err := b.postGitHubRepoStatus(ctx, client, owner, repo, sha, status); err != nil {
-		return err
+		logger.Err(errors.Wrap(err, "unable to post repo status"))
```
bluekeyes: Zerolog works by attaching fields to events and then sending the event, meaning you always need a final method call (`Msg`, `Msgf`, or `Send`) to actually write the log. For something like this, I'd probably do something like:

```go
logger.Err(err).Msgf("Failed to post status for %s", sha)
```

While wrapping the error adds a stack trace, putting the additional message on the error instead of in the log isn't that useful since we don't pass this error up to anything else.

server/handler/details.go
```diff
@@ -121,31 +120,33 @@ func (h *Details) ServeHTTP(w http.ResponseWriter, r *http.Request) error {
...
+	if err != nil {
+		data.Error = err
 		return h.render(w, data)
 	}
 
-	evaluator, err := policy.ParsePolicy(config.Config)
-	if err != nil {
-		data.Error = errors.WithMessage(err, fmt.Sprintf("invalid policy at ref \"%s\"", config.Ref))
+	if evaluator == nil {
+		data.Error = errors.Errorf("Unable to evaluate: %s", config.Description())
```
bluekeyes, line 137: From your other comment, `evalutator` is nil when the policy is not defined, right? Maybe this message should be something like `No policy defined for %s`

server/handler/details.go
```diff
@@ -121,31 +120,33 @@ func (h *Details) ServeHTTP(w http.ResponseWriter, r *http.Request) error {
...
 	}
 
-	result := evaluator.Evaluate(ctx, prctx)
+	result, err := h.Base.EvaluateFetchedConfig(ctx, prctx, client, evaluator, config)
 	data.Result = &result
 
+	if err != nil {
+		if _, ok := err.(*pull.TemporaryError); ok {
+			err = errors.WithMessage(err, "This error may be temporary. Wait 30 seconds and refresh this page to retry")
+		}
```
bluekeyes: This will read in the UI as something like:

```
This error may be temporary. Wait 30 seconds and refresh this page to retry: failed to do a thing: unable to get the stuff: unexpected status doing my job.
```

While a bit more work, I think it would be nice to add an `IsTemporaryError` field to `data` and then add this help message as part of the UI rendering. That way, it could appear after the error, which I think is more natural.

This also raises a good related point about error wrapping. The message is meant to be additional context that appears as a prefix of the message already in the error. These can get pretty long so it's worth thinking about when additional context is actually valuable and then phrasing that context to succinct and direct.

It's unfortunately pretty common to end up with errors that read like "failed to X: unable to X: could not X: the actual reason X failed", where the context doesn't really add much.

server/handler/issue_comment.go
bluekeyes: Any log monitoring should use the audit event, which I believe will always happen. This change sounds fine to me.

bluekeyes (commented):
Nice, this is looking good. I had a few more comments, but they're relatively minor.

server/handler/base.go
bluekeyes: Let's leave it as-is and see if there's any user confusion should Policy Bot fail to request reviews. For instance, people might just find an appropriate chat room and ask for a review or tag some reviewers directly on the PR.

server/handler/base.go
derekjobst: Ok, sounds good. Totally missed the Msg/Msgf/Send part of this, thanks for catching it!

server/handler/details.go
derekjobst: The evaluator is `nil` without an error when config is missing or the trigger does not match. You're right, when `TriggerAll` is passed in that should only be the case for missing configs. Adding a specific formatted error like that seems a little tightly bound, although it is a lot simpler. Right now the error should resolve to:

```Unable to evaluate: No policy found at <owner>/<repo> ref:<ref>.```

For example, if a status trigger was introduced without being added to `TriggerAll`, this hardcoded error might lead the user down the wrong debugging path. That might be a very specific case though — what do you think?

Perhaps the error ```Unable to evaluate nil evaluator: ...``` would be a better message to add here.

The reason this error needs to be created here is that it is ignored elsewhere. That is a little weird, and perhaps a better way of going about this would be to emit an error for missing and non-triggered evaluations, then catch and suppress that error in the callers which prefer to ignore them. This would be more verbose, but much more explicit. I could make the errors `IgnoreableErrors`'s in the same way we added `TemporaryError`'s elsewhere. Maybe that is overkill — what do you think?

Comment (bluekeyes) [reply server/handler/details.go:137]: The original message is fine as-is; I didn't realize that the `config.Description()` included the information about the policy being missing so I thought we were confusing invalid policies with missing policies.

I don't like the idea of a generally ignorable error. If we needed this, I think we'd want a specific error code/type that indicates the failure condition. But as I mentioned, I don't think we actually need to change anything.

File: server/handler/details.go (hunk 13 above)
Comment (derekjobst) [reply server/handler/details.go]: Sounds good, added it as an element in the details template.

File: server/handler/details.go (hunk 12 above)
Comment (derekjobst) [reply server/handler/details.go:137]: Ok, sounds good. Thanks!

File: server/handler/base.go [hunk 14]
```diff
@@ -98,17 +100,28 @@ func (b *Base) PostStatus(ctx context.Context, prctx pull.Context, client *githu
 	}
 
 	if err := b.postGitHubRepoStatus(ctx, client, owner, repo, sha, status); err != nil {
-		return err
+		logger.Err(errors.WithStack(err)).
+			Str("owner", owner).
+			Str("repo", repo).
```
bluekeyes: I think these fields should be on the logger already as a result of using the logger from the context.

server/handler/details.go
```diff
@@ -109,43 +108,46 @@ func (h *Details) ServeHTTP(w http.ResponseWriter, r *http.Request) error {
...
+	if err != nil {
+		data.Error = err
 		return h.render(w, data)
 	}
 
-	evaluator, err := policy.ParsePolicy(config.Config)
-	if err != nil {
-		data.Error = errors.WithMessage(err, fmt.Sprintf("invalid policy at ref \"%s\"", config.Ref))
+	if evaluator == nil {
+		data.Error = errors.Errorf("Unable to evaluate nil evaluator: %s", config.Description())
```
bluekeyes:
```suggestion
		// config.Description() includes a good message for missing policies
		data.Error = errors.New(config.Description())
```

`nil evaluator` is a pretty low-level implementation detail for a UI message and it sounds like the existing description is good enough to tell people what the problem is. If you do want a prefix, I think the one you had before was better ("Unable to evaluate").

bluekeyes (approved):
Two minor style comments, but looks good to go otherwise.

Issue #269, opened 2021-02-15
Prevent Details Page from updating status on merged PRs


derekjobst:
The details page will now update status checks when it is loaded (#265), but this could cause confusion if checks are updated on a closed PR.

1. Verify that the details page will attempt to update status checks on a merged PR and that those checks will update.
2. Disable status check updates from the detail page of a merged PR

PR #270, opened 2021-02-16
Avoid updating state on closed PRs

derekjobst:
Addresses #269

### DIFF
--- a/pull/context.go
+++ b/pull/context.go
@@ -65,8 +65,11 @@ type Context interface {
 	// CreatedAt returns the time when the pull request was created.
 	CreatedAt() time.Time
 
-	// State returns the state of the pull request (open or closed)
-	State() string
+	// IsOpen returns true when the state of the pull request is "open"
+	IsOpen() bool
+
+	// IsClosed returns true when the state of the pull request is "closed"
+	IsClosed() bool
 
 	// HeadSHA returns the SHA of the head commit of the pull request.
 	HeadSHA() string
--- a/pull/context.go
+++ b/pull/context.go
@@ -65,6 +65,9 @@ type Context interface {
 	// CreatedAt returns the time when the pull request was created.
 	CreatedAt() time.Time
 
+	// State returns the state of the pull request (open or closed)
+	State() string
+
 	// HeadSHA returns the SHA of the head commit of the pull request.
 	HeadSHA() string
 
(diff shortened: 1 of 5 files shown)


pull/context.go
```diff
@@ -65,6 +65,9 @@ type Context interface {
 	// CreatedAt returns the time when the pull request was created.
 	CreatedAt() time.Time
 
+	// State returns the state of the pull request (open or closed)
+	State() string
```
bluekeyes: Since this is either `open` or `closed` is there an advantage to returning it as a string instead of implementing it as an `IsClosed()` or `IsOpen()` method?

recepsirin (approved):

server/handler/base.go
```diff
@@ -88,6 +88,11 @@ func (b *Base) PostStatus(ctx context.Context, prctx pull.Context, client *githu
 	sha := prctx.HeadSHA()
 	base, _ := prctx.Branches()
 
+	if !prctx.IsOpen() {
+		logger.Info().Msg("Skipping status update because PR state is not open")
+		return
+	}
+
```
derekjobst, line 95: `!IsOpen` seemed better than `IsClosed` since there could be additional options added in the future. We are specifically looking to skip in all cases except when status is `open`, rather than just when it is `closed`.

pull/context.go
```diff
@@ -65,6 +65,12 @@ type Context interface {
 	// CreatedAt returns the time when the pull request was created.
 	CreatedAt() time.Time
 
+	// IsOpen returns true when the state of the pull request is "open"
+	IsOpen() bool
+
+	// IsClosed returns true when the state of the pull request is "closed"
+	IsClosed() bool
+
```
derekjobst, line 73: Perhaps `IsClosed` is not necessary yet, but it felt right to add since `open` and `closed` are the two possible values (that I am aware of, right now).

pull/context.go
derekjobst: Returning a `bool` seems like a better option. Changed.

I believe the values are limited to `open` and `closed`, although I thought I saw somewhere that `merged` was also an option. After looking again, [this](https://docs.github.com/en/rest/reference/pulls#update-a-pull-request) is the best enumeration of state values I've been able to find and it's just the two.

server/handler/base.go
```diff
@@ -88,6 +88,11 @@ func (b *Base) PostStatus(ctx context.Context, prctx pull.Context, client *githu
 	sha := prctx.HeadSHA()
 	base, _ := prctx.Branches()
 
+	if !prctx.IsOpen() {
```
derekjobst, line 91: `!IsOpen` seemed better than `IsClosed` since there could be additional options added in the future. We are specifically looking to skip in all cases except when status is `open`, rather than just when it is `closed`.

bluekeyes (approved):

PR #273, opened 2021-03-02
Add note to README about double quoted/plain YAML string escape characters

derekjobst:
Due to the way yaml is decoded, `\` characters must be escaped when used in strings with [double quotes](https://yaml.org/spec/1.2/spec.html#id2787109), but should not be escaped in [plain style](https://yaml.org/spec/1.2/spec.html#id2788859) strings without quotes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,23 +61,6 @@ Consider the following example, which allows changes to certain paths without
 review, but all other changes require review from the `palantir/devtools`.
 Any member of the `palantir` organization can also disapprove changes.
 
-#### Notes on YAML Syntax
-The YAML language specification supports flow scalars (basic values like strings
-and numbers) in three formats: 
-[single-quoted](https://yaml.org/spec/1.2/spec.html#id2788097), 
-[double-quoted](https://yaml.org/spec/1.2/spec.html#id2787109), and 
-[plain](https://yaml.org/spec/1.2/spec.html#id2788859). Each support different 
-escape characters, which can cause confusion when used for regex strings 
-(which often contain the `\\` character).
-
-- Single Quoted: `'` is used as an escape character. Backslash characters do not need to be escaped.
-  e.g. `'^BREAKING CHANGE: (\w| )+$'`
-- Double Quoted: `\` is used as an escape character. Backslash characters must 
-  be escaped with a preceding `\`.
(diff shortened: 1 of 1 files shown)


asvoboda (approved):

bluekeyes:
While including the note here works, many other predicates also use regular expressions. Maybe it would be beneficial to instead add a section like `Notes on YAML Syntax` to the README as a sub-heading of the `policy.yml Specification` section?

That would give space to be more expansive and add other notes in the future. For example, I just checked and single quoted strings are also valid and don't need escaping.

derekjobst:
That's a great idea. I'll add the section now. I'd be worried users might miss it when looking up a single regex based rule in the examples — what do you think of adding a note to `comment_patterns` and `changed_files` paths that repeat the difference and reference this new section?

bluekeyes:
The inline comment format isn't great for adding lots of information about each property and I'd be worried about duplication, but I think a line similar to `See "Notes on YAML Syntax" for tips to correctly escape regular expressions` above each relevant property would be fine.

README.md
```diff
@@ -61,6 +61,23 @@ Consider the following example, which allows changes to certain paths without
 review, but all other changes require review from the `palantir/devtools`.
 Any member of the `palantir` organization can also disapprove changes.
 
+#### Notes on YAML Syntax
```
bluekeyes: I'd place this after the sample YAML block and before the `Remote Policy Configuration` section. Right now, the previous text says "Consider the following example" but then this section appears before the example.

README.md
```diff
@@ -61,6 +61,23 @@ Consider the following example, which allows changes to certain paths without
...
+The YAML language specification supports flow scalars (basic values like strings
+and numbers) in three formats: 
+[single-quoted](https://yaml.org/spec/1.2/spec.html#id2788097), 
+[double-quoted](https://yaml.org/spec/1.2/spec.html#id2787109), and 
+[plain](https://yaml.org/spec/1.2/spec.html#id2788859). Each support different 
+escape characters, which can cause confusion when used for regex strings 
+(which often contain the \\ character).
```
```suggestion
(which often contain the `\\` character).
```

README.md
derekjobst: Oops, good catch

bluekeyes (approved):

PR #275, opened 2021-03-26
Clarify PR Title Matching Error

rfc1149:
When viewing this check in the github UI, it's not obvious what "Title" is referring to. Adding "PR" to clarify that it's referring to the title of the PR itself that needs to be updated.


asvoboda (approved):
Thanks for the clarification!

PR #279, opened 2021-04-02
Use Actions for CI and publishing

bluekeyes:


bluekeyes:
No idea if the release and publishing stuff will work, as it's kind of hard to test on a branch without making an actual release. Any ideas for testing this beyond careful review and then making point-releases-of-shame if it doesn't work?

.github/workflows/build.yml
```diff
@@ -0,0 +1,141 @@
...
+      if: ${{ startsWith(github.ref, 'refs/tags/') }}
+      env:
+        UPLOAD_URL: ${{ steps.find-release.outputs.upload_url }}
+      run: |
+        dist_file=$(./godelw artifacts dist)
+        curl -sS --fail -X POST \
+          -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+          -H "Content-Type: application/gzip" \
+          "${UPLOAD_URL}?name=$(basename ${dist_file})" \
+          --data-binary @${dist_file}
```
bluekeyes: I feel kind of bad just doing this with `curl` but the semi-official action to upload release assets is archived/deprecated and most community ones I found seem to also want to create the release. I guess a lot of people want to push a tag and have that automate the release creation instead of creating the release as a way to create the tag.

Another option is to write a new publisher plugin for Godel that can do this and the previous step. That's probably the best long-term option, but I wasn't ready to write that.

I know @jamestoyer likes `goreleaser`, but it looked like it also handles a bunch of the build process that we currently use Godel for and also works on the "push a tag to create a release" model.

jmcampanini:
this looks pretty dang cool! i defer to @jamestoyer on what to use. i can see a godel plugin being useful here.

bluekeyes:
I expect the Godel plugin will exist soon as part of moving the Godel ecosystem itself to publish to GitHub Releases, so we might just be ahead of the curve there.

.github/workflows/build.yml
```diff
@@ -0,0 +1,141 @@
...
+    - name: Find release
+      if: ${{ startsWith(github.ref, 'refs/tags/') }}
+      id: find-release
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        curl -sS --fail -H "Authorization: Bearer ${GITHUB_TOKEN}" "${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/releases/tags/${GITHUB_REF#refs/tags/}" \
+          | jq .upload_url \
+          | sed 's/{.*}$//' \
+          | xargs -0 printf "::set-output name=upload_url::%s\n"
```
derekjobst: This feature of github actions is pretty cool! [Docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs) if anyone else wants to read more.

.github/workflows/build.yml
derekjobst: `goreleaser` looks nice. Autorelease Bot also follows this workflow of Create Release/Tag → Run CI to add artifacts/dists to the release, right?

derekjobst (approved):
Excited for this!

.github/workflows/build.yml
bluekeyes: Yeah, Autorelease does create a tag before the release and then the tag creation triggers CI, but most (all?) of these projects don't actually attach artifacts to the release, but publish to other systems.

I was mainly flagging the workflow differences for people creating releases. Our current process is:

1. Publish a GitHub Release with release notes and a tag that does not exist yet
2. GitHub creates the tag
3. The tag triggers CI
4. CI publishes/upload artifacts

It seems like many of the open-source tools I found assume this process:

1. Locally create a tag and push it to GitHub
2. The tag triggers CI/automation
3. The automation creates a GitHub Release for the the existing tag
4. CI publishes/uploads artifacts

nmiyake:
Will likely discuss in our next sync, but we've actually already rolled out OSS publishing for the godel ecosystem using GitHub actions and have excavated it out across OSS. Here's an example: https://github.com/palantir/godel-generate-plugin/blob/master/.github/workflows/publish-godel-artifacts.yml

.github/workflows/build.yml
jamestoyer: A lot of OSS tooling will also be fine with an existing release. They will just update that. Generally, I much prefer using the `release` event to trigger the actual building of a release artifact

asvoboda:
> Will likely discuss in our next sync, but we've actually already rolled out OSS publishing for the godel ecosystem using GitHub actions and have excavated it out across OSS. Here's an example: https://github.com/palantir/godel-generate-plugin/blob/master/.github/workflows/publish-godel-artifacts.yml

In the case of policy-bot, we really only publish a container and don't rely on the resulting tarball at all in prod. However, we recognize this is useful to some of the folks externally who use policy bot. This PR is intended to replace all of our circleci usage with actions, not just the publishing aspect.

asvoboda (approved):
Lets try it out!

bluekeyes:
Ok, I added the secrets by hand and updated the workflow to use `release` events, so I think I'm ready to flip the required status checks and try this out.

jamestoyer (approved):

asvoboda:
🎉  👍

PR #285, opened 2021-04-16
Add support for signature verification predicates

IainSteers:
This PR introduces support for policy-bot to verify the presence of git commit signatures in a pull request.

- New predicate `has_valid_signatures`, this predicate should be used to test if commits are signed and their signatures valid.
- New predicate `has_valid_signatures_by`, commits are required to have valid signatures and additionally the signatures must be authenticated as one of the valid users/team/org members listed.
- New predicate `has_valid_signatures_by_key`, commits are required to have valid signatures and additionally the signatures must be attributed to one of the key_ids listed.

```yaml
approval_rules:

- name: commits are signed
  if:
    has_valid_signatures: true

- name: commits are signed by excavator
  if:
    has_valid_signatures_by:
      users: ["svc-excavator-bot"]
      
- name: commits are signed by known excavator key
  if:
    has_valid_signatures_by_keys:
      key_ids: ["3AA5C34371567BD2"]

```


policy/predicate/signature.go
```diff
@@ -0,0 +1,92 @@
...
+
+	signers := make(map[string]struct{})
+
+	for _, c := range commits {
+		if c.Signature == nil {
+			return false, fmt.Sprintf("Commit %.10s has no signature", c.SHA), nil
+		}
+		if !c.Signature.IsValid {
+			return false, fmt.Sprintf("Commit %.10s has an invalid signaturer", c.SHA), nil
+		}
```
asvoboda, line 75: This could probably be a helper func with the same code in `HasValidSignatures`

policy/predicate/signature.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return false, "", errors.Wrap(err, "failed to get commits")
+	}
+
+	for _, c := range commits {
+		if c.Signature == nil {
+			return false, fmt.Sprintf("Commit %.10s has no signature", c.SHA), nil
+		}
+		if !c.Signature.IsValid {
+			return false, fmt.Sprintf("Commit %.10s has an invalid signaturer due to %s", c.SHA, c.Signature.State), nil
```
asvoboda: Does state encode any helpful debugging info?

pull/context.go
```diff
@@ -200,3 +204,13 @@ type Reviewer struct {
...
 }
+
+type Signature struct {
+	Email             string
+	IsValid           bool
+	Payload           string
+	Signature         string
+	Signer            string
+	State             string
+	WasSignedByGitHub bool
```
asvoboda: Are there other parts of https://docs.github.com/en/graphql/reference/objects#gpgsignature that we want to expose? Maybe `KeyID`?

pull/github.go
```diff
@@ -860,6 +860,7 @@ type v4Commit struct {
 			OID string
 		}
 	} `graphql:"parents(first: 3)"`
+	Signature *v4GitSignature
```
asvoboda, line 863: Should we be looking for a https://docs.github.com/en/graphql/reference/objects#gpgsignature or a https://docs.github.com/en/graphql/reference/objects#smimesignature ? It seems like we're using the later?

IainSteers: I was going off the generic interface that they both implement https://docs.github.com/en/graphql/reference/interfaces#gitsignature

Perhaps it makes more sense to key off of the implemented type and grab the extra fields

IainSteers: I refactored to inline fragment the implementation objects and expose it that way

policy/predicate/signature.go
IainSteers: It's exposed via the GitHub ui (clicking on the unverified button), but I figured it would provide useful context in policy-bot given how some people skip straight to the details page.

policy/predicate/signature.go
IainSteers: Done

README.md
```diff
@@ -244,6 +244,25 @@ if:
...
+  # or belong to any of the listed organizations or teams.
+  has_valid_signatures_by:
+    users: ["user1", "user2", ...]
+    organizations: ["org1", "org2", ...]
+    teams: ["org1/team1", "org2/team2", ...]
+
+  # "has_valid_signatures_by_keys" is satisfied if the commits in the pull request
+  # all have git commit signatures that have been verified by GitHub, and
+  # the authenticated signatures are attributed to a key in the list of key-ids.
+  has_Valid_signatures_by_keys:
```
asvoboda:
```suggestion
  has_valid_signatures_by_keys:
```

README.md
```diff
@@ -244,6 +244,25 @@ if:
...
       - "^(docs|style|chore): (\\w| )+$"
 
+  # "has_valid_signatures" is satisfied if the commits in the pull request
+  # all have git commit signatures that have been verified by GitHub
+  has_valid_signatures: true
+
+  # "has_valid_signatures_by" is satisfied if the commits in the pull request
+  # all have git commit signatures that have been verified by GitHub, and
+  # the authenticated signatures are attributed to a user in the users list 
+  # or belong to any of the listed organizations or teams.
```
```suggestion
  # or belong to a user in any of the listed organizations or teams.
```

README.md
```diff
@@ -244,6 +244,25 @@ if:
     not_matches:
       - "^(docs|style|chore): (\\w| )+$"
 
+  # "has_valid_signatures" is satisfied if the commits in the pull request
+  # all have git commit signatures that have been verified by GitHub
+  has_valid_signatures: true
```
bluekeyes, line 249: If this is `false`, what should that mean? Right now, the the code allows any PR when this is `false`. I think it might be better if it actually inverted the condition: at least one commit is missing a signature or has an invalid signature.

policy/predicate/signature.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (pred HasValidSignatures) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+	commits, err := prctx.Commits()
+	if err != nil {
+		return false, "", errors.Wrap(err, "failed to get commits")
+	}
+
+	if !pred {
+		return true, "", nil
+	}
```
bluekeyes: See my note above about what it means when this predicate is false. If we do keep the current implementation, let's do this check before getting the commits.

policy/predicate/signature.go
```diff
@@ -0,0 +1,150 @@
...
+
+	keys := make(map[string]struct{})
+
+	for _, c := range commits {
+		valid, desc := hasValidSignature(ctx, c)
+		if !valid {
+			return false, desc, nil
+		}
+		// Only GPG signatures are valid for this predicate
```
bluekeyes, line 115: Let's mention this in the docs for this predicate as well, something like "are attributed to a GPG key with an ID in the list."

policy/predicate/signature.go
```diff
@@ -0,0 +1,150 @@
...
+		default:
+			return false, fmt.Sprintf("Commit %.10s signature is not a GPG signature", c.SHA), nil
+		}
+	}
+
+	for key := range keys {
+		isValidKey := false
+		for _, acceptedKey := range pred.KeyIDs {
+			if key == acceptedKey {
+				isValidKey = true
```
bluekeyes, line 128:
```suggestion
				isValidKey = true
				break
```

Although I could also see converting `pred.KeyIDs` to a map or moving this inner loop check to a helper.

pull/context.go
```diff
@@ -200,3 +204,67 @@ type Reviewer struct {
...
+
+type Signature interface {
+	GetEmail() string
+	GetIsValid() bool
+	GetPayload() string
+	GetSignature() string
+	GetSigner() string
+	GetState() string
+	GetSignedByGitHub() bool
+}
```
bluekeyes: I'm not sold on the use of an interface to represent the different signature types here, especially since the only difference is the addition of the ID field in the GPG case. It stands out in comparison to the rest of the `pull.Context` return types, which are all simple concrete structs.

Using a pointer to a concrete `Signature` in `Commit` also makes it clearer that the value may not be present.

What do you think about using a concrete type with an optional `KeyID` field and maybe a `Type` field (similar to the `Type` field on `Reviewer`, especially since you already defined a type type for use with GraphQL)? Since you already have functions to convert the GitHub API types into the `pull.Context` types, I don't think it would change the code much.

pull/github.go
```diff
@@ -933,3 +940,70 @@ func isNotFound(err error) bool {
 	}
 	return false
 }
+
+type SignatureType string
+
+const (
+	SignatureGpg     SignatureType = "GpgSignature"
+	SignatureSmime   SignatureType = "SmimeSignature"
+	SignatureUnknown SignatureType = "UnknownSignature"
```
bluekeyes: Is this type used anywhere?

bluekeyes (commented):
I'm interested to see how these predicates end up being used. I can see uses as disapproval conditions and in combination with other conditions, but I'm curious if they will be used on their own. Maybe commit signatures could replace the forgeable (without additional pre-receive hooks) author and committer checks to some extent?

pull/github.go
IainSteers: No, I thought it might be necessary at one point but forgot to remove it. Done

pull/context.go
IainSteers: That's fair. I mostly went with the interface out of habit rather than a particular need.

You're right it didn't change the code much and arguably given then types involved it feels neater now.

policy/predicate/signature.go
IainSteers: 👍

README.md
IainSteers: I changed the flow so that it properly inverts the condition. It looks to work pretty similarly to the AuthorIsOnlyContributor logic now.

pull/context.go
```diff
@@ -200,3 +204,15 @@ type Reviewer struct {
...
+	Type              SignatureType
+	Email             string
+	IsValid           bool
+	KeyID             string
+	Payload           string
+	Signature         string
+	Signer            string
+	State             string
+	WasSignedByGitHub bool
+}
```
bluekeyes: Historically, I've tried to only expose fields that were actually used and then add additional fields as needed. It looks like `WasSignedByGitHub`, `Email`, `Payload`, `Signature`, and maybe others are not used. What do you think about removing those for now?

bluekeyes (approved):
One non-blocking comment on minimizing unused fields, but otherwise this looks good to me.

PR #297, opened 2021-05-11
Fix conflict between direct and inherited perms

bluekeyes:
Previously, if the aggregate permission for a collaborator was higher than a permission granted directly, the direct permission was discarded and the user would not be selected for reviews. This was particularly bad for org admins who are also members of a team with lower permissions.

In the revised model, return both the aggregate permission and additional lower permissions that are granted directly so that they may still be used for reviewer assignment. ~~However, for the approval to actually count, the rule must specify the minimum permission and all higher permissions that selected users may have.~~

~~For example, if a team has `maintain` permission and a member of the team is an org admin, that member's approval will only count if the rule requires both `admin` and `maintain` permission.~~

~~This is making me think that the `permissions` key should be interpreted as a minimum permission for approval, but as an exact list of permissions for requests. That way, you can list `maintain` and both maintainers and admins can approve but if a review is requested, it will only request people who explicitly and directly have `maintain` permission. While it sounds more complicated, I think the actual result will be less confusing - exact permissions don't make a lot of sense for approval.~~

Implemented the minimum permission approval in #298.

I still need to add some tests for the new behavior.

### DIFF
--- a/pull/context.go
+++ b/pull/context.go
@@ -220,10 +220,23 @@ type Reviewer struct {
 }
 
 type Collaborator struct {
-	Name       string
+	Name        string
+	Permissions []CollaboratorPermission
+}
+
+func (c *Collaborator) HasExactPermission(p Permission, viaRepo bool) bool {
+	for _, cp := range c.Permissions {
+		if cp.ViaRepo == viaRepo && cp.Permission == p {
+			return true
+		}
+	}
+	return false
+}
+
+type CollaboratorPermission struct {
 	Permission Permission
 
 	// True if Permission is granted by a direct or team association with the
 	// repository. If false, the permisssion is granted by the organization.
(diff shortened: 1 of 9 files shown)


pull/context.go
```diff
@@ -220,10 +220,23 @@ type Reviewer struct {
 }
 
 type Collaborator struct {
-	Name       string
+	Name        string
+	Permissions []CollaboratorPermission
+}
+
+func (c *Collaborator) HasExactPermission(p Permission, viaRepo bool) bool {
```
asvoboda: Is this used anywhere?

asvoboda (approved):

pull/context.go
bluekeyes: Good catch, I thought I was going to use this then ended up doing something else

PR #298, opened 2021-05-11
Use minimum permission for approval checks

bluekeyes:
Change the 'permissions' field to accept any approval from a user with
at least the minimum permission specified in the list. This means that
for approval purposes, there is never a reason to specify more than one
permissions.

For reviewer assignment, it may be useful to specify more than one
permission to expand the set of users who are eligible for selection.
For example, this would allow selecting reviews from both admin and
maintainer teams.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,8 +345,11 @@ requires:
 
   # A user must have at least the minimum permission in this list for their
   # approval to count for this rule. Valid permissions are "admin", "maintain",
-  # "write", "triage", and "read". Specifying more than one permission is
-  # useful to control reviewer assignment.
+  # "write", "triage", and "read".
+  #
+  # Specifying more than one permission is only useful to control which users
+  # or teams are selected for review requests. See the documentation on review
+  # requests for details.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -343,10 +343,11 @@ requires:
...
   teams: ["org1/team1", "org2/team2"]
 
-  # A user must have a permission in this list for their approval to count for
-  # this rule. Valid permissions are "admin", "maintain", "write", "triage",
-  # and "read".
-  permissions: ["admin", "maintain", "write"]
+  # A user must have at least the minimum permission in this list for their
+  # approval to count for this rule. Valid permissions are "admin", "maintain",
+  # "write", "triage", and "read". Specifying more than one permission is
+  # useful to control reviewer assignment.
```
derekjobst: It might be helpful to clarify how specifying more than one permission affects reviewer assignment? Perhaps mentioning that there are more details below.

derekjobst (approved):

PR #300, opened 2021-05-14
Parse comparison expressions when loading policies

### LINKED ISSUE #289: Policy validation should check for invalid comparison operators
I accidentally bricked our pbot config by using `>=` (I know, should have double checked the docs).  If the policy validation API checked for that, it would be very helpful.

bluekeyes:
Expressions are functionally equivalent to before, but are now validated
as part of the validation endpoint. One behavior change is that we now
allow any number of spaces between the operator and the number.

Fixes #289.

### DIFF
--- a/policy/predicate/file.go
+++ b/policy/predicate/file.go
@@ -15,6 +15,7 @@
 package predicate
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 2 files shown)


policy/predicate/file.go
```diff
@@ -96,35 +95,79 @@ type ModifiedLines struct {
...
+
+func (exp *ComparisonExpr) UnmarshalText(text []byte) error {
+	if len(text) == 0 {
+		*exp = ComparisonExpr{}
+		return nil
+	}
+
+	var i int
+	var op CompareOp
+	switch text[i] {
```
asvoboda, line 152: Should we attempt to trim leading whitespace?

asvoboda (approved):

derekjobst (approved):

PR #302, opened 2021-05-14
Fix re-request loop for user reviewers

### LINKED ISSUE #263: Stop re-requesting reviews when all-users mode is used
We've encountered the following issues on a GitHub repository today:
- We have a PR that modifies a single file.
- That file is configured to require approval from a GitHub team, and the requesting mode is set to `all-users`.
- policy-bot requests the review from all the users of that GitHub team, as expected.
- A member reviews the PR (in this case the PR is rejected asking for changes).
- policy-bot immediately re-requests a review from the member that reviewed.

This triggers more email notifications than are necessary, but more importantly the state the PR is left in looks like it needs to be reviewed (i.e. action pending on the reviewing team) when it actually isn't.

bluekeyes:
If a user was requested for review, but their review didn't change the
status of the policy for some reason, policy-bot could immediately
re-request them as a reviewer. This was most likely to happen if a user
left a "request changes" review on a repository that does not have a
disapproval policy.

To fix this, look at all reviews on the current commit and exclude their
authors from the list of requests. If the pull request is updated in the
future, the users become eligible for requesting again. Because we
already list reviews for basic policy evaluation, this shouldn't add any
new requests.

Fixes #263.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -982,6 +982,9 @@ type v4PullRequestReview struct {
 	State       string
 	Body        string
 	SubmittedAt time.Time
+	Commit      struct {
+		OID string
+	}
 }
 
 func (r *v4PullRequestReview) ToReview() *Review {
@@ -990,6 +993,7 @@ func (r *v4PullRequestReview) ToReview() *Review {
 		Author:    r.Author.GetV3Login(),
 		State:     ReviewState(strings.ToLower(r.State)),
 		Body:      r.Body,
+		SHA:       r.Commit.OID,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pull/github.go
```diff
@@ -982,6 +982,9 @@ type v4PullRequestReview struct {
 	State       string
 	Body        string
 	SubmittedAt time.Time
+	Commit      struct {
+		OID string
```
asvoboda, line 987: Is `OID` `Object ID`? It feels a bit funky that the graphql api doesn't just call this `sha`.

asvoboda (approved):

derekjobst (approved):

bluekeyes:
Hmm, I think I'll need to update this to also account for "commented" reviews. Right now, those get grouped with comments rather than with reviews, so their users aren't counted as reviewers.

PR #307, opened 2021-05-19
Skip reviewer assignment on comment/review events

bluekeyes:
With the current set of predicates, these events can never change the
set of requested reviewers or trigger a re-request so there's no reason
to do the work of computing and assigning reviewers.

This is an alternate solution to the re-request loops that I fixed by
including head commit reviews in the set of reviewers. I think the other
approach is more correct, but this doesn't hurt.

The main downside is that if we add predicates that use comments or
reviews in the future, it will be easy to forget to change this trigger
condition and it might be a while before someone reports unexpected
behavior with the new predicate and reviewer assignment.

See also #306 and #302.

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -171,7 +171,7 @@ func (b *Base) Evaluate(ctx context.Context, installationID int64, trigger commo
 		return err
 	}
 
-	return b.RequestReviewsForResult(ctx, prctx, client, result)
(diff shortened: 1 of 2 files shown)


asvoboda (approved):
I think the benefit of reducing the number of Github API calls is a decent tradeoff now for a possible future bug.

PR #321, opened 2021-07-30
Add support for shared organization policies

bluekeyes:
Refactor configuration loading to use the `go-githubapp/appconfig` package. This mostly adds support for shared organization policies (in the `.github` repository by default), but should also make error messages more accurate.

It also changes the policy link in the details view to point to the resolved policy, instead of the local policy file.

I'll need to rebase this after #320 merges, but I didn't want to combine the `go-github` upgrade with the functionality changes. Also needs https://github.com/palantir/go-githubapp/pull/94 to merge to maintain current behavior with empty configuration files.

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -33,7 +33,10 @@ import (
 )
 
 const (
-	DefaultPolicyPath         = ".policy.yml"
+	DefaultPolicyPath       = ".policy.yml"
+	DefaultSharedRepository = ".github"
+	DefaultSharedPolicyPath = "policy.yml"
+
 	DefaultStatusCheckContext = "policy-bot"
 
 	LogKeyGitHubSHA = "github_sha"
@@ -52,6 +55,9 @@ type Base struct {
 
 type PullEvaluationOptions struct {
 	PolicyPath string `yaml:"policy_path"`
+
+	SharedRepository string `yaml:"shared_repository"`
+	SharedPolicyPath string `yaml:"shared_policy_path"`
 
 	// StatusCheckContext will be used to create the status context. It will be used in the following
 	// pattern: <StatusCheckContext>: <Base Branch Name>
@@ -72,6 +78,12 @@ type PullEvaluationOptions struct {
 func (p *PullEvaluationOptions) FillDefaults() {
 	if p.PolicyPath == "" {
 		p.PolicyPath = DefaultPolicyPath
+	}
+	if p.SharedRepository == "" {
+		p.SharedRepository = DefaultSharedRepository
+	}
+	if p.SharedPolicyPath == "" {
+		p.SharedPolicyPath = DefaultSharedPolicyPath
 	}
 
 	if p.StatusCheckContext == "" {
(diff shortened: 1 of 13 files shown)


server/handler/base.go
```diff
@@ -151,17 +163,11 @@ func (b *Base) Evaluate(ctx context.Context, installationID int64, trigger commo
 		return err
 	}
 
-	fetchedConfig, err := b.ConfigFetcher.ConfigForPR(ctx, prctx, client)
-	if err != nil {
-		return errors.WithMessage(err, fmt.Sprintf("Failed to fetch configuration: %s", fetchedConfig))
-	}
-
+	fetchedConfig := b.ConfigFetcher.ConfigForPR(ctx, prctx, client)
```
asvoboda, line 166: just checking: we don't return an error anymore here because if one exists, its embedded in the `fetchedConfig` and is then evaluated below in `ValidateFetchedConfig` with better errors?

server/handler/base.go
```diff
@@ -174,32 +180,36 @@ func (b *Base) Evaluate(ctx context.Context, installationID int64, trigger commo
...
-	if fetchedConfig.Missing() {
-		logger.Debug().Msg(fetchedConfig.Description())
+	switch {
+	case fc.LoadError != nil:
+		msg := fmt.Sprintf("Error loading policy from %s", fc.Source)
+		logger.Warn().Err(fc.LoadError).Msg(msg)
 
-		return nil, nil
-	}
+		b.PostStatus(ctx, prctx, client, "error", msg)
```
asvoboda, line 191: nit (since it seems like `PostStatus` is showing up in this PR): should we have something like `PostErrorStatus` or a constant so that callers don't need to pass around magic `"error"` strings?

server/handler/fetcher.go
```diff
@@ -16,194 +16,49 @@ package handler
...
-	PolicyPath string
+	Loader *appconfig.Loader
 }
 
-// ConfigForPR fetches the policy configuration for a PR. It returns an error
-// only if the existence of the policy could not be determined. If the policy
-// does not exist or is invalid, the returned error is nil and the appropriate
-// fields are set on the FetchedConfig.
-func (cf *ConfigFetcher) ConfigForPR(ctx context.Context, prctx pull.Context, client *github.Client) (FetchedConfig, error) {
+func (cf *ConfigFetcher) ConfigForPR(ctx context.Context, prctx pull.Context, client *github.Client) FetchedConfig {
```
asvoboda, line 40: this is so much nicer!

asvoboda (approved):

server/handler/base.go
bluekeyes: Yes, previously, loading errors were handled by failing and not posting anything on the pull request. Only parsing errors resulted in a status check. In the new model, both types of errors are handled the same way.

server/handler/base.go
bluekeyes: This is an existing function that is called in slightly different places now. It's also called with `"pending"` and `"success"` (so the possible check statuses in GitHub). We could make defined functions for each status, but I think I'd leave that for a separate PR.

PR #329, opened 2021-08-27
Update validate endpoint to check local and remote policies

asvoboda:
Currently, policy-bot fails to validate remote policies. Instead, policy bot's validate endpoint should accept a remote config as valid. I noticed this once trying to use the new org default policies in #321 .

Perhaps we should also check that the remote policy is also valid?

### DIFF
--- a/server/handler/validate.go
+++ b/server/handler/validate.go
@@ -15,10 +15,12 @@
 package handler
 
 import (
+	"fmt"
 	"io/ioutil"
 	"net/http"
 
 	"github.com/palantir/go-baseapp/baseapp"
+	"github.com/palantir/go-githubapp/appconfig"
 	"github.com/palantir/policy-bot/policy"
 	"github.com/palantir/policy-bot/version"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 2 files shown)


server/handler/validate.go
```diff
@@ -45,22 +47,40 @@ func Validate() http.Handler {
...
-			baseapp.WriteJSON(w, http.StatusUnprocessableEntity, &check)
+		if validLocal || validRemote {
+			check.Message = "Policy file is valid"
+			baseapp.WriteJSON(w, http.StatusOK, &check)
 			return
 		}
 
-		check.Message = "Policy file is valid"
-		baseapp.WriteJSON(w, http.StatusOK, &check)
+		check.Message = fmt.Sprintf("Policy is invalid and neither local nor remote. '%s' OR '%s'.", localStrErr, remoteStrErr)
```
asvoboda: Should we only return the local string error? This line atm would definitely make #196 harder

server/handler/validate.go
```diff
@@ -45,22 +47,40 @@ func Validate() http.Handler {
...
+	if _, err := policy.ParsePolicy(&policyConfig); err != nil {
+		return false, err.Error()
+	}
+	return true, ""
+}
+
+func isValidRemotePolicy(requestPolicy []byte) (bool, string) {
+	var remoteConfig appconfig.RemoteRef
+
+	if err := yaml.UnmarshalStrict(requestPolicy, &remoteConfig); err != nil {
```
bluekeyes: Might want to use `appconfig.YAMLRemoteRefParser` so that the logic is exactly the same as what's used for real. The catch is that most invalid configurations won't produce an error, they'll just return `nil` indicating it is not a remote policy. Maybe that's enough?

server/handler/validate.go
bluekeyes: Yeah, given these are already hard to read, I'd like to not make it harder. Maybe this can be simplified by following the same logic as the config loader (hinted at below)? In other words, if you get an error parsing a remote policy, we know it is supposed to be remote and that's the only message to return. If it doesn't appear to be a valid remote, assume it's supposed to be local and validate it that way.

server/handler/validate.go
```diff
@@ -45,22 +47,40 @@ func Validate() http.Handler {
...
+		if validLocal || validRemote {
+			check.Message = "Policy file is valid"
+			baseapp.WriteJSON(w, http.StatusOK, &check)
 			return
 		}
 
-		check.Message = "Policy file is valid"
-		baseapp.WriteJSON(w, http.StatusOK, &check)
+		check.Message = fmt.Sprintf("Policy is invalid and neither local nor remote. '%s' OR '%s'.", localStrErr, remoteStrErr)
+		baseapp.WriteJSON(w, http.StatusBadRequest, &check)
```
bluekeyes: This is going to drop a status code variant compared to the previous version. Not sure how important that is, but I can imagine callers checking for specific codes.

bluekeyes (commented):
I'm not sure about validating the remote policy. On one hand, if we don't, it's an easy way to break the policy in a repository. But on the other, actually doing the validation might be complicated and suddenly requires an endpoint that used to require no API requests or authentication to make authenticated requests.

jamestoyer:
I agree with @bluekeyes that testing out the remote policy will make things hard and I'm not sure of it's worth. To me, it makes sense to validate that the remote policy configuration in the local repo is syntactically correct. You could then argue the remote policy should be validated in the repo it's defined in

server/handler/validate.go
jamestoyer: I think `http.StatusUnprocessableEntity` makes most sense here, so lets go with that

asvoboda:
👍 thanks for picking this up.

IainSteers (approved):

PR #338, opened 2021-10-27
Add GitHub request cache size metric

IainSteers:
Currently policy=bot leverages a lrucache to cache github responses.

Previously, we don't measure the approximate memory size in-use by the cache. This should resolve that and allow us to set the cache max size accordingly..

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -32,6 +32,7 @@ import (
 	"github.com/palantir/go-githubapp/appconfig"
 	"github.com/palantir/go-githubapp/githubapp"
 	"github.com/palantir/go-githubapp/oauth2"
+	"github.com/palantir/policy-bot/metrics"
 	"github.com/palantir/policy-bot/server/handler"
 	"github.com/palantir/policy-bot/version"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -98,7 +106,7 @@ func New(c *Config) (*Server, error) {
 		githubapp.WithClientUserAgent(userAgent),
 		githubapp.WithClientTimeout(githubTimeout),
 		githubapp.WithClientCaching(true, func() httpcache.Cache {
-			return lrucache.New(maxSize, 0)
+			return githubCache
```
bluekeyes, line 109: This is actually a significant change that I don't think will work. While we didn't document this well in `go-githubapp`, it's important that this is a factory function that returns a new cache instance for each client. `httpcache` only uses the method and URL as the cache key, so in general we need separate instances to make sure each client has its own view of the world, respecting the different permissions between clients.

We _might_ be able to prove that a single global cache is safe for policy-bot (e.g. by showing that all requests have IDs or names in the URL and that the forced re-validation will surface permission error), but it feels like setting ourselves up for subtle bugs later on.

bluekeyes (commented):
I agree this would be nice to have, but I think it will be challenging to implement here. Given the cache-per-client requirement (see comment), I don't think we can register these with the metrics registry without leaking memory.

That makes me think this needs to be part of `go-githubapp` instead. Maybe:

1. Check if `httpcache.Cache` implementations also implement a `Size() int64` method
2. If yes and client metrics are enabled, record the cache size with each request or every N requests. This could probably be a gauge tagged with information about the client (like the rate limit metrics) or a histogram aggregating cache information for all clients.

Because we already emit metrics about cached vs uncached requests from `go-githubapp` I think this would fit in nicely there.

server/server.go
IainSteers: Ah yes, you're correct.

I missed that its a factory function, looking at githubapp a bit deeper though I see how its used.

Yes, I agree about being part of the githubapp client metrics.

I'll take another look at this tomorrow and see if there's a cleaner approach inside of githubapp

PR #349, opened 2021-12-08
fix: PullRequestState returns uppercase enums

gcampbell12:
Not sure when/if this changed but the `IsOpen()` and `IsClosed()` functions are currently returning false even if the pull request is actually open or closed respectively because the enums returned by the graphql API return the state as uppercase (https://docs.github.com/en/graphql/reference/enums#pullrequeststate), this should keep things backwards compatible by always converting the state to lowercase before comparing it.


asvoboda:
👍  Thanks for the fix @gcampbell12 . Can you send me an internal PR so I can double check the logs?

PR #357, opened 2021-12-22
feat: ignore edited comments with new option

### LINKED ISSUE #356: feat: ignore edited comments with new option
Firstly, thanks for maintaining such an amazing tool! We hope to get many miles out of it.

Our engineering org has strict governance and compliance rules related to code review / approval. One part of our code review process is ensuring the integrity of an approving PR comment. As a result, edited comments need to be treated with utmost scrutiny. We would love to use github reviews entirely, but still have somewhat of a comment driven PR workflow and there doesn’t seem to be a way to easily determine if an edited approval comment is genuine. In our own tooling, we’ve taken the approach of ignoring edited comments when evaluating approval rules. As we lean into policy-bot, we thought this might be a useful feature in case using the `github_reviews: true` option isn’t sufficient for all use cases. Example policy here:

```yaml
approval_rules:
- name: domain review
  requires:
    count: 1
    teams:
    - "Betterment/engineering"
    permissions:
    - write
  options:
    invalidate_on_push: true
    ignore_edited_comments: true
    methods:
      github_review: false
      comment_patterns:
      - '.*?domain.*?lgtm'
- name: platform review
  requires:
    count: 1
    teams:
    - "Betterment/platform-reviewers"
    permissions:
    - write
   options:
    invalidate_on_push: true
    ignore_edited_comments: true
    methods:
      github_review: false
      comment_patterns:
      -  '.*?platform.*?lgtm'
```

Lastly, I'd be happy to open a pull request if you think this is something worthwhile. Thanks and happy holidays!

devinburnette:
Closes https://github.com/palantir/policy-bot/issues/356

As discussed in the attached issue, this feature would allow a new boolean option `ignore_edited_comments:` that when set to `true` would ignore both issue comments and review comments that have been edited when evaluating the approval rules. This helps guard against potential situations where a user could edit a comment to gain approval outside of the intended approval rules construct.

I went with adding the `updated_at` field we get back from the graphql response because that makes it easy to determine quickly if a comment has been edited irrespective of whom the comment was edited by (which was also an intentional choice).

I went with putting it in the approval part of the code because it seemed like that was where it best fit naturally given we already have access to the rule options and the candidates and this is the place we seem to already be doing other types of candidate filtering. This would have been harder to implement over in the server handlers as we don't have access to the options yet. Let me know what you think of this approach, open to other ideas, but this seemed the easiest to implement.

### DIFF
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -41,8 +41,9 @@ type Options struct {
 	AllowContributor bool `yaml:"allow_contributor"`
 	InvalidateOnPush bool `yaml:"invalidate_on_push"`
 
-	IgnoreUpdateMerges bool          `yaml:"ignore_update_merges"`
-	IgnoreCommitsBy    common.Actors `yaml:"ignore_commits_by"`
+	IgnoreEditedComments bool          `yaml:"ignore_edited_comments"`
+	IgnoreUpdateMerges   bool          `yaml:"ignore_update_merges"`
+	IgnoreCommitsBy      common.Actors `yaml:"ignore_commits_by"`
 
 	RequestReview RequestReview `yaml:"request_review"`
 
@@ -248,13 +249,54 @@ func (r *Rule) IsApproved(ctx context.Context, prctx pull.Context) (bool, string
 }
 
 func (r *Rule) filteredCandidates(ctx context.Context, prctx pull.Context) ([]*common.Candidate, error) {
-	log := zerolog.Ctx(ctx)
-
 	candidates, err := r.Options.GetMethods().Candidates(ctx, prctx)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get approval candidates")
 	}
+
 	sort.Stable(common.CandidatesByCreationTime(candidates))
+
+	if r.Options.IgnoreEditedComments {
+		candidates, err = r.filterEditedCandidates(ctx, prctx, candidates)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if r.Options.InvalidateOnPush {
+		candidates, err = r.filterInvalidCandidates(ctx, prctx, candidates)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return candidates, nil
+}
+
+func (r *Rule) filterEditedCandidates(ctx context.Context, prctx pull.Context, candidates []*common.Candidate) ([]*common.Candidate, error) {
+	log := zerolog.Ctx(ctx)
+
+	if !r.Options.IgnoreEditedComments {
+		return candidates, nil
+	}
+
+	var allowedCandidates []*common.Candidate
+	for _, candidate := range candidates {
+		if r.Options.IgnoreEditedComments {
+			if candidate.UpdatedAt == candidate.CreatedAt {
+				allowedCandidates = append(allowedCandidates, candidate)
+			}
+		}
+	}
+
+	log.Debug().Msgf("discarded %d candidates with edited comments",
+		len(candidates)-len(allowedCandidates))
+
+	return allowedCandidates, nil
+}
+
+func (r *Rule) filterInvalidCandidates(ctx context.Context, prctx pull.Context, candidates []*common.Candidate) ([]*common.Candidate, error) {
+	log := zerolog.Ctx(ctx)
 
 	if !r.Options.InvalidateOnPush {
 		return candidates, nil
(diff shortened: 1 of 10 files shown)


policy/approval/approve.go
```diff
@@ -224,6 +225,13 @@ func (r *Rule) IsApproved(ctx context.Context, prctx pull.Context) (bool, string
 			continue
 		}
 
+		if r.Options.IgnoreEditedComments {
+			if c.UpdatedAt != c.CreatedAt {
+				log.Debug().Str("user", c.User).Msg("ignoring approval because comment was edited")
+				continue
+			}
+		}
```
bluekeyes: While I believe this will work here, I think it might be more correct to perform this filtering as part of the `filteredCandidates` function. That will make sure they are ignored before almost any other processing happens and logically, if we're ignoring these comments, they should never appear as "candidates" for approval.

bluekeyes (commented):
Thanks for the PR! Overall, this looks good, but I had one suggestion about where to place the filtering logic.

policy/approval/approve.go
```diff
@@ -275,16 +276,23 @@ func (r *Rule) filteredCandidates(ctx context.Context, prctx pull.Context) ([]*c
...
+		if r.Options.InvalidateOnPush {
+			if candidate.CreatedAt.Before(*last.PushedAt) {
+				log.Debug().Msgf("discarded candidate because it was invalidated by push of %s at %s",
+					last.SHA,
+					last.PushedAt.Format(time.RFC3339))
+				continue
+			}
 		}
+		allowedCandidates = append(allowedCandidates, candidate)
```
devinburnette, line 293: @bluekeyes will this work? or were you thinking something different?

devinburnette: @bluekeyes like this? or were you thinking something different?

bluekeyes: Yeah, this is the right direction, but I think it needs a few more adjustments.

policy/approval/approve.go
```diff
@@ -256,7 +257,7 @@ func (r *Rule) filteredCandidates(ctx context.Context, prctx pull.Context) ([]*c
 	}
 	sort.Stable(common.CandidatesByCreationTime(candidates))
 
-	if !r.Options.InvalidateOnPush {
+	if !r.Options.IgnoreEditedComments && !r.Options.InvalidateOnPush {
```
bluekeyes: I think we need to treat these as two separate conditions: the code on L264-275 is only relevant for `InvalidateOnPush`, so ideally we only run that when the option is enabled. There's also a situation (L269) where we could return early without applying `IgnoreEditedComments` filtering.

It might make sense to extract each filter into a new helper function so that we can directly update `candidates`, something like:

```go
candidates, err := r.Options.GetMethods().Candidates(ctx, prctx)
// ...

if r.Options.IgnoreEditedComments {
  candidates, err = r.filterEditedCandidates(ctx, prctx, candidates)
  if err != nil {
    return nil, err
  }
}
if r.Options.InvalidateOnPush {
  candidates, err = r.filterInvalidCandidates(ctx, prctx, candidates)
  if err != nil {
    return nil, err
  }
}

return candidates
```

devinburnette: i see, i missed L269, will give this approach a shot.

devinburnette: i guess i was just trying to avoid looping over the candidates multiple times

policy/approval/approve.go
```diff
@@ -248,14 +249,55 @@ func (r *Rule) IsApproved(ctx context.Context, prctx pull.Context) (bool, string
...
+
+	log.Debug().Msgf("discarded %d candidates with edited comments",
+		len(candidates)-len(allowedCandidates))
+
+	return allowedCandidates, nil
+}
+
+func (r *Rule) filterInvalidCandidates(ctx context.Context, prctx pull.Context, candidates []*common.Candidate) ([]*common.Candidate, error) {
+	log := zerolog.Ctx(ctx)
+
```
devinburnette, line 300: @bluekeyes updated with the suggested approach here.

bluekeyes (approved):
Thanks for the updates - this looks good to me! I think iterating over candidates multiple times is fine given the expected low number and the relative cost compared to all the API requests that have to be made.

I'll try to test this out in our environment and make a release by the end of the week. If you'd like to test in your environment before that and you're using Docker, you can pull the `snapshot` tag from Docker Hub once this merges and the build completes.

PR #361, opened 2022-01-07
feat: add comment matching to github review approval with new option

### LINKED ISSUE #359: feat: add comment matching to github review approval
Hello, it's me again. I have another feature idea that could help us with our use cases.

We love github's built in PR reviews and would love to be able to use policy-bot with github reviews in our PR workflow, but we currently have to rely on comments to distinguish the various types of reviewers that may be members of the same groups. For governance/compliance reasons, we require certain team members to receive multiple reviews for both content and style, whereas for other team members we require only one review based on content.

As you can imagine, it becomes very difficult to delineate which review approval counts for which rule. For example in our org, anyone in engineering can perform a domain review, but a platform expert must perform a platform review. Of course our platform reviewers are also members of engineering, so we need a way to explicitly detect which reviewer is satisfying which rule and so the below example policy didn't work for us.

```yaml
approval_rules:
- name: domain review
  requires:
    count: 1
    teams:
    - "Betterment/engineering"
    permissions:
    - write
  options:
    invalidate_on_push: true
    methods:
      github_review: true
- name: platform review
  requires:
    count: 1
    teams:
    - "Betterment/platform-reviewers"
    permissions:
    - write
   options:
    invalidate_on_push: true
    methods:
      github_review: true
```

I originally thought that some combination of one of the `comments:` options and `github_review: true` would do it, but that doesn't appear to be the case.

We could use comments alone, but there are so many more other features you get with github reviews, like requesting review, integrating codeowners, built in approval dismissal functionality, etc. Leaning into Github Reviews as the canonical source of approval but with the added flexibility of human-enabled opt-in behavior seems like the best fit for us, and we believe this can benefit other teams as well who need to support more nuanced approval rules.

So the feature would be adding comment matching ability for github review body comments in order to determine approval and not just the review approval itself. An example of an accompanying policy file with this change would look like:

```yaml
approval_rules:
- name: domain review
  requires:
    count: 1
    teams:
    - "Betterment/engineering"
    permissions:
    - write
  options:
    invalidate_on_push: true
    methods:
      github_review: true
      github_review_comment_patterns:
      - "domain lgtm"
- name: platform review
  requires:
    count: 1
    teams:
    - "Betterment/platform-reviewers"
    permissions:
    - write
   options:
    invalidate_on_push: true
    methods:
      github_review: true
      github_review_comment_patterns:
      -  "platform lgtm"
```

If you think the above sounds acceptable, I don't think it should be too difficult to implement. As always, I'm happy to take a stab at a PR.

devinburnette:
Closes https://github.com/palantir/policy-bot/issues/359

As discussed in the attached issue, this feature would support new method options for `github_review_comments:` and `github_review_comment_patterns:` creating parity with issue comment approval for github review approval comment bodies. This would help in use cases where github review approval alone isn't sufficient.

I went with nesting the logic in the existing reviews for loop but first checking to confirm the review state is approved. That would make this feature work if `github_review` is provided in the config and set to `true`, or not provided in the config defaulting to `true`. As written, this feature doesn't work if `github_review: false` explicitly in the config. That doesn't seem surprising behavior to me, but let me know if there's an approach that makes better sense to accomplish the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,8 +347,14 @@ options:
     # See the Notes on YAML Syntax section of this README for more information.
     comment_patterns:
       - "^Signed-off by \\s+$"
-    # If true, GitHub reviews can be used for approval. Default is true.
+    # If true, GitHub reviews can be used for approval. All GitHub review approvals
+    # will be accepted as approval candidates. Default is true.
     github_review: true
+    # Just like the "comment_patterns" option, but for GitHub reviews. Only GitHub
+    # review approvals matching the included patterns will be accepted as
+    # approval candidates. Defaults to an empty list.
+    github_review_comment_patterns:
+      - '\b(?i)domain\s*lgtm\b'
 
 # "requires" specifies the approval requirements for the rule. If the block
 # does not exist, the rule is automatically approved.
--- a/policy/common/methods.go
+++ b/policy/common/methods.go
@@ -23,9 +23,10 @@ import (
 )
 
 type Methods struct {
-	Comments        []string `yaml:"comments,omitempty"`
-	CommentPatterns []Regexp `yaml:"comment_patterns,omitempty"`
-	GithubReview    bool     `yaml:"github_review,omitempty"`
+	Comments                    []string `yaml:"comments,omitempty"`
+	CommentPatterns             []Regexp `yaml:"comment_patterns,omitempty"`
+	GithubReview                bool     `yaml:"github_review,omitempty"`
+	GithubReviewCommentPatterns []Regexp `yaml:"github_review_comment_patterns,omitempty"`
 
 	// If GithubReview is true, GithubReviewState is the state a review must
 	// have to be considered a candidated. It is currently excluded from
@@ -71,7 +72,7 @@ func (m *Methods) Candidates(ctx context.Context, prctx pull.Context) ([]*Candid
 		}
 	}
 
-	if m.GithubReview {
+	if m.GithubReview || len(m.GithubReviewCommentPatterns) > 0 {
 		reviews, err := prctx.Reviews()
 		if err != nil {
 			return nil, err
@@ -79,11 +80,21 @@ func (m *Methods) Candidates(ctx context.Context, prctx pull.Context) ([]*Candid
 
 		for _, r := range reviews {
 			if r.State == m.GithubReviewState {
-				candidates = append(candidates, &Candidate{
-					User:      r.Author,
-					CreatedAt: r.CreatedAt,
-					UpdatedAt: r.UpdatedAt,
-				})
+				if len(m.GithubReviewCommentPatterns) > 0 {
+					if m.githubReviewCommentMatches(r.Body) {
+						candidates = append(candidates, &Candidate{
+							User:      r.Author,
+							CreatedAt: r.CreatedAt,
+							UpdatedAt: r.UpdatedAt,
+						})
+					}
+				} else {
+					candidates = append(candidates, &Candidate{
+						User:      r.Author,
+						CreatedAt: r.CreatedAt,
+						UpdatedAt: r.UpdatedAt,
+					})
+				}
 			}
 		}
 	}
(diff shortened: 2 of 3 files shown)


policy/common/methods.go
```diff
@@ -23,9 +23,11 @@ import (
 )
 
 type Methods struct {
-	Comments        []string `yaml:"comments,omitempty"`
-	CommentPatterns []Regexp `yaml:"comment_patterns,omitempty"`
-	GithubReview    bool     `yaml:"github_review,omitempty"`
+	Comments                    []string `yaml:"comments,omitempty"`
+	CommentPatterns             []Regexp `yaml:"comment_patterns,omitempty"`
+	GithubReview                bool     `yaml:"github_review,omitempty"`
+	GithubReviewComments        []string `yaml:"github_review_comments,omitempty"`
```
bluekeyes: Let's remove this option and only support `GitHubReviewCommentPatterns`.

If I were to implement Policy Bot again (or make a v2 configuration format), I think I'd only support `CommentPatterns`, since it's a superset of the plain `Comments` field. While we haven't formally deprecated `Comments`, I consider it deprecated.

policy/common/methods.go
```diff
@@ -119,5 +131,15 @@ func (m *Methods) CommentMatches(commentBody string) bool {
...
+	for _, comment := range m.GithubReviewComments {
+		if strings.Contains(commentBody, comment) {
+			return true
+		}
+	}
+	for _, pattern := range m.GithubReviewCommentPatterns {
+		if pattern.Matches(commentBody) {
+			return true
+		}
+	}
```
bluekeyes: I believe this will have to be in a new method so that review comment patterns are not matched against regular comment patterns and vice versa.

policy/common/methods.go
```diff
@@ -79,11 +81,21 @@ func (m *Methods) Candidates(ctx context.Context, prctx pull.Context) ([]*Candid
 
 		for _, r := range reviews {
 			if r.State == m.GithubReviewState {
-				candidates = append(candidates, &Candidate{
-					User:      r.Author,
-					CreatedAt: r.CreatedAt,
-					UpdatedAt: r.UpdatedAt,
-				})
+				if len(m.GithubReviewComments) > 0 || len(m.GithubReviewCommentPatterns) > 0 {
```
bluekeyes: Because of the way method defaults work, I don't think this will work as expected (or at least as we discussed in the issue.) We only use the defaults if the `methods` block is completely absent. If the user sets any fields, no defaults are used. That means that with the current code, users will have to set both `github_reviews` and `github_review_comment_patterns` to have things work.

We may want to change the condition on L76 to `m.GitubReview || len(m.GithubReviewCommandPatterns) > 0` so that setting either field will enable review processing - unless we actually want to require that both fields are set.

bluekeyes (commented):
Thanks for adding new tests along with this!

policy/common/methods.go
devinburnette: ahh ok, i think i misunderstood how the defaults were applied. Lemme look into fixing this.

policy/common/methods.go
devinburnette: makes sense, fine by me.

README.md
```diff
@@ -347,8 +347,14 @@ options:
...
       - "^Signed-off by \\s+$"
-    # If true, GitHub reviews can be used for approval. Default is true.
+    # If true, GitHub reviews can be used for approval. All GitHub review approvals
+    # will be accepted as approval candidates. Default is true.
     github_review: true
+    # Just like the "comment_patterns" option, but for GitHub reviews. Only GitHub
+    # review approvals matching the included patterns will be accepted as
+    # approval candidates. Defaults to an empty list.
+    github_review_comment_patterns:
+      - '\b(?i)domain\s*lgtm\b'
```
devinburnette, line 357: @bluekeyes ok! I think this is all set. I've also updated the readme to reflect the use case more accurately.

bluekeyes (approved):
Looks good, thanks!

PR #364, opened 2022-01-14
support env vars for options in server config

### LINKED ISSUE #363: Support Env Variables for Options in Server Config
Hello again!

We're looking at setting up a shared repository that isn't the default .github repository. I see the option for it in the server config to change where policy bot looks for a shared policy config, but I don't see that as being a currently supported Env variable. We could probably just as easy deploy a config file with this in it, but so far this is the only thing and Env variables are a lot easier for us to manage.

Would you accept a PR to make the server config parser set options from Env before filling the defaults?

devinburnette:
Closes https://github.com/palantir/policy-bot/issues/363

As discussed in the attached issue, I've made an attempt at setting the policy path and shared options from env vars if they exist. If env vars are not set, then it should just fall back to the defaults. Let me know if there's a more clever way of accomplishing the same. I also updated the example config to reflect the possibilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -569,12 +569,6 @@ ignored. If this happens, you will need to reapprove the pull request.
 
 This feature has [security implications](#update-merge-conflicts).
 
-#### Private Repositories
-
-`policy-bot` works with private repositories, but currently does not support
-pull requests from private _forks_ of private repositories due to GitHub API
-limitations. Please file an issue if this functionality is important to you.
-
 #### Automatically Requesting Reviewers
 
 `policy-bot` can automatically request reviewers for all pending rules
--- a/README.md
+++ b/README.md
@@ -126,7 +126,8 @@ define a `remote` key. Only 1 level of remote configuration is supported by desi
 
 ```yaml
 # The remote repository to read the policy file from. This is required, and must
-# be in the form of "org/repo-name". Must be a public repository.
+# be in the form of "org/repo-name". The policy bot github app must have read
+# access to this repository.
 remote: org/repo-name
 
 # The path to the policy config file in the remote repository. If none is
@@ -323,6 +324,7 @@ options:
   request_review:
     # False by default
     enabled: true
+
     # mode modifies how reviewers are selected. `all-users` will request all users
     # who are able to approve the pending rule. `random-users` selects a small
     # set of random users based on the required count of approvals. `teams` will 
@@ -340,6 +342,7 @@ options:
     comments:
       - ":+1:"
       - "👍"
+
     # If a comment matches a regular expression in this list, it counts as
     # approval. Defaults to an empty list.
     #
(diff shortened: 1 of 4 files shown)


server/config.go
```diff
@@ -65,8 +65,22 @@ func ParseConfig(bytes []byte) (*Config, error) {
...
+	}
+	if v, ok := os.LookupEnv("POLICYBOT_SHARED_REPOSITORY"); ok {
+		c.Options.SharedRepository = v
+	}
+	if v, ok := os.LookupEnv("POLICYBOT_SHARED_POLICY_PATH"); ok {
+		c.Options.SharedPolicyPath = v
+	}
+	if v, ok := os.LookupEnv("POLICYBOT_STATUS_CHECK_CONTEXT"); ok {
+		c.Options.StatusCheckContext = v
+	}
```
devinburnette: This would work right? @bluekeyes

server/handler/base.go
```diff
@@ -91,6 +92,14 @@ func (p *PullEvaluationOptions) FillDefaults() {
 	}
 }
 
+func (p *PullEvaluationOptions) SetValuesFromEnv(prefix string) {
+	setStringFromEnv("POLICY_PATH", prefix, &p.PolicyPath)
+	setStringFromEnv("SHARED_REPOSITORY", prefix, &p.SharedRepository)
+	setStringFromEnv("SHARED_POLICY_PATH", prefix, &p.SharedPolicyPath)
+	setStringFromEnv("STATUS_CHECK_CONTEXT", prefix, &p.StatusCheckContext)
+	p.fillDefaults()
+}
```
devinburnette, line 101: i think maybe this is the better way mirroring your pattern from go-baseapp and go-githubapp, but i felt like these options weren't suited to live in those codebases. so i put them here instead. 

only thing i felt weird about was setting the defaults here, probably fine, but let me know what you think

README.md
```diff
@@ -126,7 +126,8 @@ define a `remote` key. Only 1 level of remote configuration is supported by desi
 
 ```yaml
 # The remote repository to read the policy file from. This is required, and must
-# be in the form of "org/repo-name". Must be a public repository.
+# be in the form of "org/repo-name". The policy bot github app must have read
+# access to this repository.
```
devinburnette, line 130: i added this because it was little misleading

README.md
```diff
@@ -564,12 +569,6 @@ ignored. If this happens, you will need to reapprove the pull request.
 
 This feature has [security implications](#update-merge-conflicts).
 
-#### Private Repositories
-
-`policy-bot` works with private repositories, but currently does not support
-pull requests from private _forks_ of private repositories due to GitHub API
-limitations. Please file an issue if this functionality is important to you.
-
```
devinburnette, line 572: removing this, because I don't think it's true anymore. private forks of private repos seem to work fine on my end.

README.md
bluekeyes: Thanks, that's probably a bit clearer... hopefully it is implied that apps have read access to public repositories.

README.md
bluekeyes: Interesting, this was originally added due to [this fallback condition](https://github.com/palantir/policy-bot/blob/develop/pull/github.go#L801-L809) when loading commit pushed dates. Maybe GitHub improved the GraphQL API so that pushed dates are returned from forks now and we don't need the fallback.

server/handler/base.go
bluekeyes: I agree it's a bit weird, but I think it's fine since this is only used in one place.

bluekeyes (approved):
Looks good, thanks for updating the docs along with this.

PR #365, opened 2022-01-14
fix: triggers should include all comments

devinburnette:
Noticed in my testing of https://github.com/palantir/policy-bot/pull/361 the code responsible for triggering evaluation was incomplete. Rules for methods that were more recently added weren't being evaluated until a user clicked on the details page in the github status link instead of at the time of the webhook event. This attempts to solve that.

Also added some tests for the Trigger method so that we can catch it in the future.


bluekeyes (approved):
Good catch, and thank you for increasing the test coverage!

PR #370, opened 2022-01-22
add unconditional successful repo status check on installation

### LINKED ISSUE #369: feat: update required status checks for repository
We noticed while setting up policy bot for the first time on a repository that the `policy-bot: main` status check context was not available to select in the github branch protection settings because it was not in the list of "status checks in the last week for this repository" search box. Apparently, github will only let you populate this check with statuses that have already posted to the repository. This creates an additional step of having to let policy bot run once against a pull request without the required check. However, I did notice that you can add a new status check context via github's API.

I noticed there's a webhook that fires when installing policy bot on a repo with the event type `installation_repositories` that policy bot appears to be automatically subscribed when ran as as a github app. What if policy bot could just upsert the required status check branch protection rule automatically upon installation on a repository?

A couple things came to mind when noodling about this feature. The first being some people might not want their PRs to be totally blocked by policy bot in the case they need to get a hot fix out immediately without approval. The second being that I believe a feature like this would require the github app to have additional repository level permissions to write the branch protection rule. I could see a policy bot server config option allowing folks to opt-in to allowing policy bot to do this on their behalf. That way it's kind of like two layers of "are you sure?" Once in the server config, and once by the act of setting the permission on the github app.

Although the feature would be useful for some, I understand the risks associated with giving the app basically administrator access on repos. Just wanted to discuss here and see if this had been thought about before and what the general consensus was for an idea like this.

devinburnette:
Closes https://github.com/palantir/policy-bot/issues/369

As discussed in the attached issue, this feature makes it easier to onboard new repositories with policy-bot. By posting an unconditional successful status check on the repo at installation allows the policy-bot status check context to be selectable in the branch protection settings without first making a commit.

Let me know if this looks okay. Happy to take a different approach if you think there's a better way of accomplishing the same.


server/handler/installation_repositories.go
```diff
@@ -0,0 +1,90 @@
...
+		return err
+	}
+
+	switch event.GetAction() {
+	case "added":
+		for _, repo := range event.RepositoriesAdded {
+			h.postRepoInstallationStatus(ctx, client, repo)
+		}
+	default:
+		return nil
```
bluekeyes: As a style thing, I'd omit the default case since there's no special logic here

server/handler/installation_repositories.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return nil
+}
+
+func (h *InstallationRepositories) postRepoInstallationStatus(ctx context.Context, client *github.Client, r *github.Repository) {
+	logger := zerolog.Ctx(ctx)
+
+	repoFullName := strings.Split(r.GetFullName(), "/")
+	owner, repo := repoFullName[0], repoFullName[1]
+	repository, _, err := client.Repositories.Get(ctx, owner, repo)
```
bluekeyes, line 67: Might be worth leaving a comment here that this is required because the repository included in the installation event is a partial object that is missing fields we need

server/handler/installation_repositories.go
```diff
@@ -0,0 +1,90 @@
...
+	branch, _, err := client.Repositories.GetBranch(ctx, owner, repo, defaultBranch, false)
+	if err != nil {
+		return
+	}
+
+	head := branch.GetCommit().GetSHA()
+
+	contextWithBranch := fmt.Sprintf("%s: %s", h.PullOpts.StatusCheckContext, defaultBranch)
+	state := "success"
+	message := "Policy-Bot successfully installed."
```
bluekeyes:
```suggestion
	message := fmt.Sprintf("%s successfully installed.", h.AppName)
```

bluekeyes (commented):
Thanks for implementing this. Hopefully it won't be too expensive when you install Policy Bot on an organization with hundreds of repositories for the first time.

server/handler/installation_repositories.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2018 Palantir Technologies, Inc.
```
derekjobst:
```suggestion
// Copyright 2022 Palantir Technologies, Inc.
```

server/handler/installation_repositories.go
```diff
@@ -0,0 +1,90 @@
...
+		return
+	}
+
+	defaultBranch := repository.GetDefaultBranch()
+	branch, _, err := client.Repositories.GetBranch(ctx, owner, repo, defaultBranch, false)
+	if err != nil {
+		return
+	}
+
+	head := branch.GetCommit().GetSHA()
```
derekjobst, line 78: Has this been tested on new repositories without any commits? I'd assume `GetBranch` above will error since there is no commit, although perhaps it will just return a `nil` commit.

devinburnette: the behavior on a new repository without a branch is that it will return early because there's no branch and i don't think you can push a branch without commits, or at least I've never seen it.

devinburnette:
@bluekeyes @derekjobst any idea of when there will be another official release?

bluekeyes (approved):

bluekeyes:
Thanks for contributing this and apologies for the delay in reviewing it again after your updates. For the next release, I'll try to test internally and tag a new release early next week.

PR #372, opened 2022-01-26
Add PR information to error object

jmcampanini:
#### Currently
When there is an error during the `Handle` logic, the logged error does not contain the PR information since the logger writing the `error` object doesn't have those field configured. This means some errors will only have the GH delivery_id and type, but not the pull request that created it.

#### This PR
Wraps the `error` object with the PR locator information: `<owner>/<repo>/<PR>` as a way to reference back in the logs what PR caused this error.


derekjobst (approved):

jmcampanini:
it seems that there is another error handler that returns more useful information, so this PR isnt necessary

PR #386, opened 2022-02-16
Use Go version from the .palantir/go-version file

bluekeyes:
Will make automatic version updates actually work and avoids having to
set the version in multiple places.


asvoboda (approved):
Much nicer!

PR #407, opened 2022-04-08
Adding a details tab to policy bot rules

yiweny:
This PR is for adding an expendable details tab to policy bot. 
Screen Shot 2022-04-17 at 3 23 30 PM

Here is an example of the UI. User can click on the little marker to expand and collapse the details of the rule. This can help user to easily understand why a rule is selected, as well as providing them a link to the users that can approve their PR.

The way policy bot works currently is that it receives a webhook from Github, and then finds the policy written in the repo. For each rule, it determines if it is skipped or selected based on the predicates. It also determines the status of each rule based on the approval/disapproval review the PR received. 

Our change adds a struct called `PredicateInfo`. This struct will be produced while the code is evaluating predicate and passed to `Result` to get rendered in html. It has two string fields and some pointer fields. 

```
type PredicateInfo struct {
	Name        string
	Type        PredicateInfoType
	Description string

	ContributorInfo *ContributorInfo
	FileInfo        *FileInfo
	PatternInfo     *PatternInfo
	CommitInfo      *CommitInfo
	StatusInfo      *StatusInfo
}
```

Because each predicate requires a different set of data to display as details, we store those in a separate struct and the `PredicateInfo` struct stores the pointer to that struct. The Type field indicates which pointer is not null/used and the Name field is used to generate frontend text, indicating what data is related to the Predicate. 

There are 15 possible predicates. In the original code, the predicates are grouped by files. Therefore, for each group of predicates, we have a specific struct to store the information of that group. 

For example, there are 4 predicates in the `predicate/author.go` file. We have the `ContributorInfo` struct that stores the data needed by the predicates in `author.go`. Specifically, each struct needs two types of data, first is that defines the condition to qualify/disqualify a predicate, and the second is the context of the PR used for determining the (dis)qualifications. In `ContributorInfo`, `Organizations`, `Teams` and `Users` are data used for qualification purpose and `Author`, `Contributors` are the actual values of the context of the PR. 

```none
type ContributorInfo struct {
	RequiredMembership Actors
	ActualAuthor       string
	ActualContributors []string
}
```
| Predicate Name | Fields Needed for Qualification | Fields Needed for Context |
| :--------------: | :--------------: | :----------: |
| `HasAuthorIn` | Organizations, Teams, Users | Author|
|`OnlyHasContributorsIn`| Organizations, Teams, Users| Contributors|
|`HasContributorIn`| Organizations, Teams, Users| Contributors|
|`AuthorIsOnlyContributor`| | Author,Contributors|

If a field is not needed, then we would not assign values to it, and by default in go, it would be `nil`.

The context field would store minimal needed data for (dis)qualification. For some predicates, when translate into logic, can be expressed as either `for all` or `there exists`. If the predicate is in the form of for all, e.g. `OnlyHasContributorsIn`, the original code has short circuiting so that while looping over all contributors, if the code sees an example of contributor that disqualifies the predicate, then it will return right away. In this case, the Contributors field in `ContributorInfo` would only store the specific contributor that doesn't qualify the predicate. Otherwise, if all contributors satisfy a condition, then the Contributors field would store all the contributors. The same short circuiting scheme applies to `there exists`, if there exists a value that qualifies a predicate, then we would simply return that value, ignoring other potential candidates. 

To display the reviewer information, we move the `Requires` struct from approval/disapproval to common package. So now both approval and disapproval uses the same `Requires` struct. This way we can easily pass the `Requires` struct to `Result`. Previous, the `Requires` struct in disapproval does not have a count field. Now we are just setting that number to 1 to keep the same behavior. 

For the front end part, we added two templates, `result-predicate-details` and `result-requires-details`. They are for displaying the predicate information and review information. 

`getPredicateInfo` and `getPredicateRequirement` in the `frontend.go` file are used to render the text displayed.

`getPredicateInfo` is used to obtain the context information related to each predicate, specifically, the value we store in the name field. And `getPredicateRequirement` is for obtaining the values for qualification/disqualification. For example, in `HasAuthorIn` predicate, it returns the organizations, teams and users. However, this function does not cover all the predicates. For boolean predicates like `HasSuccessfulStatus` or `AuthorIsOnlyContributor`, the explanatory requirement is not needed because the name of the predicate is already explanatory.

We also added the `PredicateInfo` to each predicate test. There are two issues worth thinking here. 

An issue is that using maps as dictionaries in the code would produce non-deterministic behaviors when testing in go. For example, in the `committersInListButAuthorsAreNot` test for the `OnlyHasContributorsIn` predicate, both `ttest1` and `ttest2` can be returned as result(because of short circuiting only one of the two will be returned). Therefore, my solution is to use the `Subset` function in the `testify` package. However, this cause us to not test equality in the other cases where there is no short circuiting. I do not know what is the best approach to this but two ways I can think of is 1) simply change the code to not use non-deterministic data structures like maps. 2) remove the test of which the result is non-deterministic. 

Another issue is related to the 'labels do not exist' test in `label_test.go`. I am not sure the point of testing it. I assume when labels do not exist then there is an error, so we can just return `nil` for the `PredicateInfo`. Are we just testing a case of error?


bluekeyes:
> An issue is that using maps as dictionaries in the code would produce non-deterministic behaviors when testing in go. [...] I do not know what is the best approach to this but two ways I can think of is 1) simply change the code to not use non-deterministic data structures like maps. 2) remove the test of which the result is non-deterministic.

Now that it matters, I suggest extracting the keys from the map, sorting them, and then iterating through the sorted list in the predicate implementation. We generally still need to use a `map` because we need a data structure that functions like a set and using `map` is the idiomatic way to do this in Go. Sorting the keys as a list is a good way to make the results deterministic.

> Another issue is related to the 'labels do not exist' test in label_test.go. I am not sure the point of testing it. I assume when labels do not exist then there is an error, so we can just return nil for the PredicateInfo. Are we just testing a case of error?

It looks like this was added to test the difference between a `nil` value and a `[]string{}` value. For most uses, these are equivalent in Go, so I don't think this test is adding much value. It should be safe to delete.

policy/common/predicateinfo.go
```diff
@@ -0,0 +1,78 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+type PredicateInfo struct {
+	Name string
+	Type string
```
bluekeyes: I think there's some confusion over the use of the `Type` field. When building one of these "union" structs, the field that acts as the type switch should map directly to the sub-fields of the struct. In this case, that means `Type` should contain one of `contributor`, `branch`, `file`, `label`, `commit`, `status`, or `title`.

To help clarify this, you can define a custom type instead of using `string`:

```go
type PredicateInfoType string

const (
  ContributorType PredicateInfoType = "contributor"
  BranchType      PredicateInfoType = "branch"
  // ...
)
```

The `Name` field then contains the name (or type) of the predicate, like `HasAuthorIn`. But as noted later, we might not need the `Name` field or it might be more appropriate in a different type.

policy/common/predicateinfo.go
```diff
@@ -0,0 +1,78 @@
...
+	LabelInfo       *LabelInfo
+	CommitInfo      *CommitInfo
+	StatusInfo      *StatusInfo
+	TitleInfo       *TitleInfo
+}
+
+type ContributorInfo struct {
+	Organizations []string
+	Teams         []string
+	Users         []string
```
bluekeyes: Both this and `CommitInfo` have fields that are very similar to `common.Actors` and in most cases are populated by copying from a `common.Actors` struct. Would it make sense to embed that struct here instead of copying the fields?

policy/common/predicateinfo.go
```diff
@@ -0,0 +1,78 @@
...
+	Author        string
+	Contributors  []string
```
bluekeyes: It's also not always clear which fields of a `___Info` struct represent the actual PR values and which ones are the predicate values. Maybe there's a naming convention we could use to clarify? You kind of do this in some other types, like `LabelInfo`, which as `RequiredLabels` and `PRLabels`. Maybe `Required` and `Actual` as field prefixes?

policy/approval/approve_test.go
```diff
@@ -29,6 +29,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type Requires = common.Requires
```
bluekeyes: Good use of a type alias! If we actually move the `Requires` type (see other comments), I suggest putting this alias in `approve.go` so it applies to the whole package instead of only to the tests.

policy/common/result.go
```diff
@@ -59,12 +59,20 @@ type ReviewRequestRule struct {
 	Mode RequestMode
 }
 
+type Requires struct {
+	Count int `yaml:"count"`
+
+	Actors `yaml:",inline"`
+}
```
bluekeyes: Does this type need to move to `common`? From what I can see, it looks like the `Count` property is never used, so perhaps the `Result` struct could use `Actors` directly instead?

policy/common/result.go
```diff
@@ -59,12 +59,20 @@ type ReviewRequestRule struct {
...
+
 type Result struct {
 	Name              string
 	Description       string
 	StatusDescription string
```
bluekeyes, line 65: `StatusDescription` has a similar purpose to `PredicatesInfo` and `Requires`. It might make sense to try combining these field in some way.

policy/predicate/predicate.go
```diff
@@ -26,5 +26,5 @@ type Predicate interface {
 
 	// Evaluate determines if the predicate is satisfied. It also returns an
 	// optional string providing details about the evaluation result.
-	Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error)
+	Evaluate(ctx context.Context, prctx pull.Context) (bool, string, *common.PredicateInfo, error)
```
bluekeyes: It's unfortunate that this now returns 4 values. I think it would be best if we could combine the existing string description with the new `PredicateInfo` value. Maybe make it a field of the struct or add a function to render the struct into a text description similar to how we render it in to HTML?

policy/common/predicateinfo.go
```diff
@@ -0,0 +1,78 @@
...
+
+type BranchInfo struct {
+	Patterns []string
+	Branch   string
+}
+
+type FileInfo struct {
+	Paths              []string
+	IgnorePaths        []string
+	ChangedFiles       []string
```
bluekeyes: It would be nice if there was a way to further generalize these structs to reduce the number of different types to manage. For example, these three fields of `FileInfo`, `LabelInfo`, `StatusInfo`, and `TitleInfo` all implement a similar concept:

```go
type PatternInfo struct {
  IncludePatterns []string
  ExcludePatterns []string
  ActualValues    []string
}
```

Some of those fields will be empty or only contain one value, but conceptually, all of these predicates have a list of required values/patterns, a (optional) list of values/patterns to ignore/exclude/fail, and then one or more actual values matched against the patterns.

Add in a `ValueName` field or something and when rendering, you can do:

```
Predicate {{.Name}} is satisfied because the {{.PatternInfo.ValueName}}:
{{range .PatternInfo.ActualValues}}* {{.}}
{{end}}
matched:
{{range .PatternInfo.IncludePatterns}}* {{.}}
{{end}}
```

which renders as:

```
Predicate OnlyChangedFiles is satisfied because the files:

* src/test/FooTest.java
* src/test/BarTest.java

matched:

* ^src/test/.*$
```

You might not even need the lists since there's usually a single `(pattern, value)` combination that passes or fails the rule.

policy/predicate/file.go
```diff
@@ -184,16 +241,50 @@ func (pred *ModifiedLines) Evaluate(ctx context.Context, prctx pull.Context) (bo
...
 
-	for expr, v := range map[ComparisonExpr]int64{
-		pred.Additions: additions,
-		pred.Deletions: deletions,
-		pred.Total:     additions + deletions,
-	} {
-		if !expr.IsEmpty() && expr.Evaluate(v) {
-			return true, "", nil
+	if !pred.Additions.IsEmpty() && pred.Additions.Evaluate(additions) {
+		res, err := pred.Additions.MarshalText()
```
bluekeyes: It's unfortunate that you need to check the error here, especially since it should always be `nil` in practice. I think it would make sense to:

1. Add a `String` method to `ComparisonExpr`. This will basically be the same implementation as `MarshalText`, but it can use `?` or `_` or `<unknown>` if the operator is not supported instead of returning an error.
2. Modify `MarshalText` to call `String` (or possibly remove it completely)
3. Use `String` in this function to remove all the error handling

server/handler/frontend.go
```diff
@@ -55,6 +61,164 @@ func LoadTemplates(c *FilesConfig, basePath string) (templatetree.HTMLTree, erro
...
+			if len(result.Requires.Organizations) != 0 || len(result.Requires.Teams) != 0 || len(result.Requires.Users) != 0 {
+				return true
+			}
+			return false
+		},
+		"getRequires": func(result *common.Result) map[string][]Membership {
+			membershipInfo := make(map[string][]Membership)
+			for _, org := range result.Requires.Organizations {
+				if orgs, ok := membershipInfo["Organizations"]; ok {
+					membershipInfo["Organizations"] = append(orgs, Membership{Name: org, Link: "https://www.github.com/orgs/" + org})
```
bluekeyes: As mentioned in Slack, the links here need to use the server configuration to set the correct GitHub URL. You can get the base web URL from `config.Github.WebURL`. We'll also need to make sure the generated links behave correctly for both github.com and GitHub Enterprise.

server/handler/frontend.go
```diff
@@ -55,6 +61,164 @@ func LoadTemplates(c *FilesConfig, basePath string) (templatetree.HTMLTree, erro
 
 			return r
 		},
+		"getPredicateInfo": func(predicateInfo *common.PredicateInfo) []string {
```
bluekeyes: Because these new functions are quite complicated, I think they should be defined as top-level named functions rather than defined anonymously as part of this map.

server/handler/frontend.go
```diff
@@ -55,6 +61,164 @@ func LoadTemplates(c *FilesConfig, basePath string) (templatetree.HTMLTree, erro
...
+			case "Status":
+				return predicateInfo.StatusInfo.Status
+			case "Title":
+				return []string{predicateInfo.TitleInfo.PRTitle}
+			default:
+				return []string{}
+			}
+		},
+		"getPredicateRequirement": func(predicateInfo *common.PredicateInfo) map[string][]string {
+			switch predicateInfo.Type {
```
bluekeyes: It's odd that this switches on `Type`, while the other function switches on `Name`. We should aim to only have a single field that requires a `switch` like this to make it easier to correctly add new predicates.

bluekeyes (changes requested):
This looks like a good start! In this initial review, I left some higher-level comments about the overall structure and design, which I'm happy to talk more about when we meet tomorrow. I think we'll also want to look at the visual design, but that's hard to communicate here, so I'll try to make a mockup or drawing.

Later, I'll do a second review looking more at details and style things, but I didn't want to comment on code that might be deleted if the structure changes. But if you prefer, I can do that now so you have everything at once.

policy/common/result.go
yiweny: Yeah it makes sense. I will switch to using `Actors` for `Requires`

policy/common/predicateinfo.go
yiweny: I am using the `PatternInfo` you suggested for labels, branches and title, but not for files. The files have an `IgnorePath` field which indicates the fields to ignore. This piece of information cannot be included in `PatternInfo` unless we add a new field specifically for it. Also, currently the way we check if a file matches a pattern is use a function called `anyMatches()`, this function returns a boolean. So if we want to use a single `(pattern, value)` combination, we will need to change that function.

policy/common/predicateinfo.go
yiweny: I fixed the issue by putting the predicate name in the name field and having the following values for type field 
```
const (
	ContributorType PredicateInfoType = "contributor"
	FileType        PredicateInfoType = "file"
	CommitType      PredicateInfoType = "commit"
	StatusType      PredicateInfoType = "status"
	PatternType     PredicateInfoType = "pattern"
)
```

policy/common/predicateinfo.go
yiweny: I used `common.Actors` for structs that need membership information

policy/common/predicateinfo.go
yiweny: I fixed by adding `Required` and `Actual` to some structs as field prefixes to make the naming more clear.

policy/approval/approve_test.go
yiweny: I am removing this because we are no longer using the same `Requires` struct for approval and disapproval. So there is no need to put `Requires` in common package.

policy/common/result.go
yiweny: It is hard to combine to existing structs because `PredicatesInfo` is `[]*PredicateInfo` and `Requires` is `Actors`. On the other hand, I think creating another struct to combine all three is kind of unnecessary. So I hold my opinion on this one and thinks it is fine the way it is now.

policy/predicate/predicate.go
yiweny: I put `desc` as a field in `PredicateInfo`, so now we are returning 3 again.

policy/predicate/file.go
yiweny: Fixed this by creating a function called `MarshalTextToString` that runs the steps you mentioned above.

server/handler/frontend.go
yiweny: fixed based on our slack conversation

server/handler/frontend.go
yiweny: fixed by putting long functions at the end of the file.

server/handler/frontend.go
yiweny: now all the functions related to rendering predicate info switch on `Type`

policy/approval/approve.go
```diff
@@ -103,9 +103,17 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
 	res.Name = r.Name
 	res.Description = r.Description
 	res.Status = common.StatusSkipped
+	res.Requires = common.Actors{
+		Organizations: r.Requires.Organizations,
+		Teams:         r.Requires.Teams,
+		Users:         r.Requires.Users,
+	}
```
bluekeyes:
```suggestion
	res.Requires = r.Requires.Actors
```

Embedded fields in Go are implicitly named after their type, so can use this to copy the whole field in one operation.

policy/approval/approve.go
```diff
@@ -103,9 +103,17 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
...
+
+	var predicatesInfo []*common.PredicateInfo
```
bluekeyes: style: `predicateInfos`

policy/approval/approve.go
```diff
@@ -103,9 +103,17 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
...
 
 	for _, p := range r.Predicates.Predicates() {
-		satisfied, desc, err := p.Evaluate(ctx, prctx)
+		satisfied, pPredicateInfo, err := p.Evaluate(ctx, prctx)
```
bluekeyes: style: rename `pPredicateInfo` to `info`

policy/common/result.go
bluekeyes: Since this type is now unused, let's delete it from this package.

policy/disapproval/disapprove.go
```diff
@@ -104,10 +104,17 @@ func (p *Policy) Evaluate(ctx context.Context, prctx pull.Context) (res common.R
 
 	res.Name = "disapproval"
 	res.Status = common.StatusSkipped
+	res.Requires = common.Actors{
+		Organizations: p.Requires.Organizations,
+		Teams:         p.Requires.Teams,
+		Users:         p.Requires.Users,
+	}
```
bluekeyes:
```suggestion
	res.Requires = p.Requires.Actors
```

policy/disapproval/disapprove.go
```diff
@@ -104,10 +104,17 @@ func (p *Policy) Evaluate(ctx context.Context, prctx pull.Context) (res common.R
...
 
-	for _, p := range p.Predicates.Predicates() {
-		satisfied, desc, err := p.Evaluate(ctx, prctx)
+	var predicatesInfo []*common.PredicateInfo
```
bluekeyes: style: `predicateInfos`

policy/disapproval/disapprove.go
```diff
@@ -104,10 +104,17 @@ func (p *Policy) Evaluate(ctx context.Context, prctx pull.Context) (res common.R
...
 
+	for _, p := range p.Predicates.Predicates() {
+		satisfied, pPredicateInfo, err := p.Evaluate(ctx, prctx)
```
bluekeyes: style: rename `pPredicateInfo` to `info`

policy/approval/approve.go
```diff
@@ -103,9 +103,17 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
...
+		predicatesInfo = append(predicatesInfo, pPredicateInfo)
```
bluekeyes: Because `pPredicateInfo` may be `nil` (especially when `err` is not `nil`), it may be better to check for this or to do this `append` after checking the error.

policy/predicate/predicate.go
```diff
@@ -26,5 +26,5 @@ type Predicate interface {
 
 	// Evaluate determines if the predicate is satisfied. It also returns an
 	// optional string providing details about the evaluation result.
-	Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error)
+	Evaluate(ctx context.Context, prctx pull.Context) (bool, *common.PredicateInfo, error)
```
bluekeyes: Because this can return a `nil` info struct, we'll want to account for that in any place that calls `Evaluate`. The other option would be to make this a value and return the zero value when there's an error, similar to what we used to do for strings. I think either can work.

We should also update the comment on this function, as it no longer returns a string.

server/assets/css/main.css
```diff
@@ -177,6 +177,14 @@ a:hover {
   @apply text-center;
 }
 
+details summary {
+  cursor: pointer;
+}
+
+details summary > * {
+  display: inline-block;
+}
```
bluekeyes: These should go further up in the file, probably around L16. Rules like this that modify default behavior for plain HTML tags are part of the "base" styles rather than part of a component (if they were attached to class names, then they'd function more like a component.)

server/handler/frontend.go
```diff
@@ -35,7 +36,17 @@ type FilesConfig struct {
...
 
-func LoadTemplates(c *FilesConfig, basePath string) (templatetree.HTMLTree, error) {
+type Membership struct {
+	Name string
+	Link string
+}
+
+type PredicateData struct {
+	PredicateDataName string
+	PredicateValues   []string
```
bluekeyes: style: I'd probably call this `Name` and `Values`, the "predicate" part is clear from the containing type

server/templates/details.html.tmpl
```diff
@@ -56,6 +56,27 @@
 <li data-status="{{$s}}">
   <div class="bg-white p-2 shadow-sm max-w-lg status-stripe {{$s}}">
     {{template "result-details" .}}
+    {{if (or .PredicatesInfo (isRequiresNotEmpty .Requires))}}
+    <div class="pt-2">
+    <div class="bg-light-gray3">
+    <details>
```
bluekeyes: I'm not sure either of these `<div>` tags are needed. It should be possible to add two classes to a single `div` (possibly using margin instead of padding) or add the styles directly to the `details` element.

server/templates/details.html.tmpl
```diff
@@ -76,3 +97,74 @@
   {{end}}
   <p class="text-dark-gray3 text-sm">{{or .Error .StatusDescription}}</p>
 {{end}}
+
+{{define "result-predicates-details"}}
+  {{ $s := (or (and .Error "error") (.Status | print)) }}
+  {{if ne $s "skipped"}}
+    <b class="font-bold text-sm">This Rule is Selected Due to</b>
```
bluekeyes:
```suggestion
    <b class="font-bold text-sm">This rule is selected because</b>
```

server/templates/details.html.tmpl
```diff
@@ -76,3 +97,74 @@
...
+  {{else}}
+    <b class="font-bold text-sm">This Rule is Skipped Due to</b>
```
bluekeyes:
```suggestion
    <b class="font-bold text-sm">This rule is skipped because</b>
```

server/templates/details.html.tmpl
```diff
@@ -76,3 +97,74 @@
...
+            </li>
+          {{end}}
+        </ul>
+    </li>
+  {{end}}
+  </ul>
+{{end}}
+
+{{define "result-requires-details"}}
+  <b class="font-bold text-sm">This Rule Requires Review from:</b>
```
bluekeyes:
```suggestion
  <b class="font-bold text-sm">This rule requires approval from</b>
```

server/templates/details.html.tmpl
```diff
@@ -76,3 +97,74 @@
...
+      <ul class="list-disc list-inside">
+      {{range $value}}
+        <li class="ml-4">
+          <a href={{.Link}}>{{.Name}}</a>
+        </li>
+      {{end}}
+      </ul>
+    </li>
+  {{end}}
+  </ul>
```
bluekeyes: Instead of nested unordered lists, I think this is a good place for a description list, which represents a key-value mapping:

```html
<dl>
  {{range $key, $value := getRequires . }}
  <dt>{{$key}}</dt>
  <dd>
    <ul class="list-disc list-inside">
    {{range $value}}
    <li class="ml-4"><a href="{{.Link}}">{{.Name}}</a></li>
    {{end}}
    </ul>
  </dd>
  {{end}}
</dl>
```

server/templates/details.html.tmpl
```diff
@@ -76,3 +97,74 @@
   {{end}}
   <p class="text-dark-gray3 text-sm">{{or .Error .StatusDescription}}</p>
 {{end}}
+
+{{define "result-predicates-details"}}
+  {{ $s := (or (and .Error "error") (.Status | print)) }}
+  {{if ne $s "skipped"}}
```
bluekeyes: style: I would invert most of these conditions to be `if eq $s "skipped"` as I think that's a bit more natural than using the negative condition everywhere.

server/templates/details.html.tmpl
```diff
@@ -76,3 +97,74 @@
...
+{{end}}
+
+{{define "result-predicates-info"}}
+  {{ $s := (or (and .Error "error") (.Status | print)) }}
+  <ul class="list-decimal list-inside">
+  {{range .PredicatesInfo}}
+    {{$predicateData := getPredicateInfo .}}
+    <li class="pt-2 text-sm">
+        {{if ne $s "skipped"}}
+          Predicate <span class="bg-light-gray1 font-bold">{{.Name}}</span> is satisfied because the {{$predicateData.PredicateDataName}} :
```
bluekeyes: (1/?) This will require some additional changes (which I'll try to outline in other comments, so I apologize if the order is confusing), but I'd like to see if we can make this section more conversational:

```html
The {{.ValuesPhrase}}:
<ul>
{{range .Values}}<li>{{.}}</li>{{end}}
</ul>
{{if eq $s "skipped"}}do not{{else}} {{.ConditionPhrase}}
{{if .ConditionMap}}
<dl>
{{range $k, $v := .ConditionMap}}
  {{if $v}}
  <dt>{{$k}}</dt>
  <dd><ul>{{range $v}}<li>{{.}}</li>{{end}}</ul></dd>
  {{end}}
{{end}}
</dl>
{{else}}
<ul>{{range .ConditionValues}}<li>{{.}}</li>{{end}}</ul>
{{end}}
```

This would render something like (in actual HTML):

```
This rule is selected because

  1. The contributors:

     * bluekeyes
     * yiweny

     meet the membership requirements:

       Teams:
         * org/team1
         * org/team2

       Organizations:
         * other-org 
```

The same format should work for other rules too:

```
This rule is skipped because

1. The modified lines:

     * +3 / -16 (total 19)

   do not match the line conditions:

     * added > 100
```

The best outcome is that you can read this from top to bottom and it sounds mostly like a real sentence.

server/handler/frontend.go
```diff
@@ -73,3 +90,149 @@ func Static(prefix string, c *FilesConfig) http.Handler {
 
 	return http.StripPrefix(prefix, http.FileServer(http.Dir(dir)))
 }
+
+func getPredicateInfo(predicateInfo *common.PredicateInfo) PredicateData {
```
bluekeyes: (2/?) Looking at this now, I'm wondering if we can remove this function (and the `getPredicateRequirement` function.) Between the two, we're taking a `PredicateInfo` struct and turning it into:

* A string
* A list of string values
* A map from string to a list of strings

To do that, we've defined 5 different sub-structs of `PredicateInfo` and implemented two `switch` statements. What would happen if we passed this directly from the predicates and could use `PredicateInfo` directly in the rendering?

I think that would lead to a struct like this:

```go
type PredicateInfo struct {
  // Describes the values, used as "the $ValuesPhrase"; must be plural
  ValuesPhrase string
  Values []string

  // Describes the condition, used as "$ConditionPhrase" or "does not $ConditionPhrase"
  ConditionPhrase string
  // If non-empty, use the map, otherwise, use the regular list
  ConditionMap map[string][]string
  ConditionValues []string
}
```

Most of the logic that is part of these functions would move to the predicates themselves so they can directly fill out these fields instead of filling out intermediate types that are converted by another level.

server/handler/frontend.go
```diff
@@ -73,3 +90,149 @@ func Static(prefix string, c *FilesConfig) http.Handler {
...
+		}
+		return PredicateData{"commits", commits}
+	case common.StatusType:
+		return PredicateData{"status", predicateInfo.StatusInfo.Status}
+	default:
+		return PredicateData{"", []string{}}
+	}
+}
+
+func getPredicateRequirement(predicateInfo *common.PredicateInfo) map[string][]string {
```
bluekeyes: If we do keep both of these functions, can we combine them? `PredicateData` could get a `Requirement` field that is a `map[string][]string`. That would mean we only need to implement a single `switch` that handles everything.

server/handler/frontend.go
```diff
@@ -73,3 +90,149 @@ func Static(prefix string, c *FilesConfig) http.Handler {
...
+		return true
+	}
+	return false
+}
+
+func getRequires(result *common.Result, githubURL string) map[string][]Membership {
+	membershipInfo := make(map[string][]Membership)
+	for _, org := range result.Requires.Organizations {
+		if orgs, ok := membershipInfo["Organizations"]; ok {
+			membershipInfo["Organizations"] = append(orgs, Membership{Name: org, Link: githubURL + "/orgs/" + org})
```
bluekeyes: Let's add `/people` to the end of this link so it goes to the member listing, similar to what you did for teams.

server/handler/frontend.go
```diff
@@ -73,3 +90,149 @@ func Static(prefix string, c *FilesConfig) http.Handler {
...
+		} else {
+			membershipInfo["Organizations"] = []Membership{{Name: org, Link: githubURL + "/orgs/" + org}}
```
bluekeyes: style: because maps return zero values for missing keys and `append` correctly handles the zero value for slices, you don't need this `if`/`else` condition. You can append directly to the map and reduce duplication:

```go
membershipInfo["Organizations"] = append(membershipInfo["Organizations"], Membership{
  Name: org, 
  Link: "",
})
```

server/handler/frontend.go
```diff
@@ -73,3 +90,149 @@ func Static(prefix string, c *FilesConfig) http.Handler {
...
+		teamName := strings.Split(team, "/")
+		if teams, ok := membershipInfo["Teams"]; ok {
+			membershipInfo["Teams"] = append(teams, Membership{Name: team, Link: githubURL + "/orgs/" + teamName[0] + "/teams/" + teamName[1] + "/members"})
+		} else {
+			membershipInfo["Teams"] = []Membership{{Name: team, Link: githubURL + "/orgs/" + teamName[0] + "/teams/" + teamName[1] + "/members"}}
+		}
+	}
+	for _, user := range result.Requires.Users {
+		if users, ok := membershipInfo["Users"]; ok {
+			membershipInfo["Users"] = append(users, Membership{Name: user, Link: githubURL + user})
```
bluekeyes: The previous uses of `githubURL` assume it does not have a trailing slash, but this one assumes it does. We should be consistent, probably by calling `strings.TrimSuffix(githubURL, "/")` upfront and then always adding our own slashes.

policy/common/predicateinfo.go
```diff
@@ -0,0 +1,73 @@
...
+	FileType        PredicateInfoType = "file"
+	CommitType      PredicateInfoType = "commit"
+	StatusType      PredicateInfoType = "status"
+	PatternType     PredicateInfoType = "pattern"
+)
+
+type PredicateInfo struct {
+	Name        string
+	Type        PredicateInfoType
+	Description string
```
bluekeyes: (3/?) Thanks for adding the description here. In addition to removing the need for all the sub-types, one other advantage of my proposed new `PredicateInfo` struct is that there's a clearer path for rendering the description automatically. For example:

```go
type PredicateInfo struct {
  Satisfied bool

  // Describes the values, used as "the $ValuesPhrase"
  ValuesPhrase string
  Values []string

  // Describes the condition, used as "$ConditionPhrase" or "does not $ConditionPhrase"
  ConditionPhrase string
  // If non-empty, use the map, otherwise, use the regular list
  ConditionMap map[string][]string
  ConditionValues []string
}

func (pi PredicateInfo) String() string {
  if pi.Satisfied {
    return fmt.Sprintf("The %s %s", pi.ValuePhrase, pi.ConditionPhrase)
  }
  return fmt.Sprintf("The %s does not %s", pi.ValuePhrase, pi.ConditionPhrase)
}
```

Adding the `Satisfied` field could even change the `Predicate.Evaluate` signature from 3 to 2 return values. We could then call this `PredicateResult` instead of `PredicateInfo`.

policy/predicate/author.go
```diff
@@ -29,17 +30,30 @@ type HasAuthorIn struct {
...
+			Teams:         pred.Teams,
+			Users:         pred.Users,
+		},
+		ActualAuthor: author,
+	}
+	predicateInfo := common.PredicateInfo{
+		Type:            common.ContributorType,
+		Name:            "HasAuthorIn",
+		ContributorInfo: &contributorInfo,
+	}
```
bluekeyes: (4/?) To give an example of how the proposed `PredicateInfo` struct might work in a predicate:

```go
info := PredicateInfo{
  ValuesPhrase: "authors",
  Values: []string{author},
  ConditionPhrase: "meet the required membership conditions",
  ConditionsMap: map[string][]string{
    "Organizations": pred.Organizations,
    "Teams": pred.Teams,
    "Users": pred.Users,
  },
}
```

policy/predicate/author.go
```diff
@@ -127,26 +184,39 @@ type AuthorIsOnlyContributor bool
...
+
+	var contributorInfo common.ContributorInfo
+
+	contributorInfo.ActualAuthor = author
+
+	predicateInfo := common.PredicateInfo{
+		Type:            common.ContributorType,
+		Name:            "AuthorIsOnlyContributor",
+		ContributorInfo: &contributorInfo,
+	}
```
bluekeyes: (5/?) Here's another example:

```go
info := PredicateInfo{
  ValuesPhrase: "commits",
  Values: []string{fmt.Sprintf("%.10s", c.SHA}, // or the full list of commit SHAs
  ConditionPhrase: "meet the authorship requirement",
  ConditionValues: []string{
    fmt.Sprintf("authored and committed by %s", author),
  },
}
```

policy/predicate/file.go
```diff
@@ -140,6 +192,14 @@ func (exp ComparisonExpr) MarshalText() ([]byte, error) {
 	return []byte(fmt.Sprintf("%s %d", op, exp.Value)), nil
 }
 
+func (exp ComparisonExpr) MarshalTextToString() string {
```
bluekeyes: This should be called `String` to match Go conventions (implement the `Stringer` interface)

policy/predicate/file.go
```diff
@@ -140,6 +192,14 @@ func (exp ComparisonExpr) MarshalText() ([]byte, error) {
...
+	res, err := exp.MarshalText()
+	if err != nil {
+		return "Unknown"
```
bluekeyes:
```suggestion
		return fmt.Sprintf("?? (op:%d) %d"), exp.Op, exp.Value) 
```

This will give at least a bit of information about what's wrong.

policy/predicate/file.go
```diff
@@ -184,16 +252,39 @@ func (pred *ModifiedLines) Evaluate(ctx context.Context, prctx pull.Context) (bo
...
+	fileInfo.DeletedLines = deletions
+	res, err = pred.Total.MarshalText()
+	if err != nil {
+		return false, nil, errors.Wrap(err, "failed to marshal text for total limit")
 	}
-	return false, fmt.Sprintf("modification of (+%d, -%d) does not match any conditions", additions, deletions), nil
+	fileInfo.TotalLimit = string(res[:])
+	fileInfo.TotalModifiedLines = additions + deletions
+	predicateInfo.Description = fmt.Sprintf("modification of (+%d, -%d) does not match any conditions", additions, deletions)
+	return false, &predicateInfo, nil
```
bluekeyes: (6/?) One more example

```go
info := PredicateInfo{
  ValuesPhrase: "file modifications",
  Values: []string{fmt.Sprintf("+%d / -%d (total %d)", additions, deletions, additions+deletions)},
  ConditionPhrase: "meet the modification conditions",
  ConditionValues: []string{ // filter this to only contain the non-empty conditions
    fmt.Sprintf("added %s", pred.Additions),
    fmt.Sprintf("deleted %s", pred.Deletions),
    fmt.Sprintf("total %s", pred.Total),
  }
}
```

bluekeyes (changes requested):
Thanks for the updates! Looking through this again, I had another idea for how to further simplify the `PredicateInfo` concept, hopefully making this easier to understand and extend in the future. Let me know if it make sense to you or if you have any disagreements or questions. I think it can work in all cases, but I haven't actually tried every single predicate.

That's really the only design comment I have, although it is a big one. Look for the comments that start with `(1/?)` for discussion on that topic. Everything else is relatively minor comments about code style.

I appreciate you iterating with me on this so far!

server/templates/details.html.tmpl
yiweny: This might not be well-suited to for the disapproval case which also uses this template. It's kind of weird to say a disapproval rule requires approval from

policy/predicate/predicate.go
yiweny: don't think it's going to return `nil` anymore now that we are putting the Satisfied bool into it though.

policy/approval/approve.go
yiweny: Since now we are putting the `satisfied` as a bool field to `PredicateResult`, the pointer would not be `nil` anymore given our logic. So I assume we don't need to check this anymore?

server/handler/frontend.go
yiweny: Removed this because of our new struct. We no longer need to render tex in the `frontend.go` file

server/templates/details.html.tmpl
```diff
@@ -76,3 +95,62 @@
...
+  {{end}}
+  {{template "result-predicates-info" .}}
+{{end}}
+
+{{define "result-predicates-info"}}
+  {{ $s := (or (and .Error "error") (.Status | print)) }}
+  <ul class="list-decimal list-inside py-2">
+  {{range .PredicateResults}}
+  <li>
+  The {{.ValuePhrase}} :
```
bluekeyes:
```suggestion
  The {{.ValuePhrase}}:
```

server/templates/details.html.tmpl
bluekeyes: Did you try using the descriptions lists here as well, instead of the `p` tag with the key?

server/handler/frontend.go
```diff
@@ -73,3 +82,26 @@ func Static(prefix string, c *FilesConfig) http.Handler {
...
+	for _, org := range result.Requires.Organizations {
+		membershipInfo["Organizations"] = append(membershipInfo["Organizations"], Membership{Name: org, Link: githubURL + "/orgs/" + org + "/people"})
+	}
+	for _, team := range result.Requires.Teams {
+		teamName := strings.Split(team, "/")
+		membershipInfo["Teams"] = append(membershipInfo["Teams"], Membership{Name: team, Link: githubURL + "/orgs/" + teamName[0] + "/teams/" + teamName[1] + "/members"})
+
+	}
+	for _, user := range result.Requires.Users {
+		membershipInfo["Users"] = append(membershipInfo["Users"], Membership{Name: user, Link: githubURL + user})
```
bluekeyes:
```suggestion
		membershipInfo["Users"] = append(membershipInfo["Users"], Membership{Name: user, Link: githubURL + "/" + user})
```

server/handler/frontend.go
```diff
@@ -55,6 +60,10 @@ func LoadTemplates(c *FilesConfig, basePath string) (templatetree.HTMLTree, erro
 
 			return r
 		},
+		"isRequiresNotEmpty": isRequiresNotEmpty,
```
bluekeyes: style: I might call this `hasRequires`

policy/predicate/author.go
```diff
@@ -29,17 +30,28 @@ type HasAuthorIn struct {
...
-
 	if !result {
 		desc = fmt.Sprintf("The pull request author %q does not meet the required membership conditions", author)
 	}
 
-	return result, desc, err
+	predicateResult := common.PredicateResult{
+		Satisfied:       result,
+		Description:     desc,
+		ValuePhrase:     "Author",
```
bluekeyes:
```suggestion
		ValuePhrase:     "authors",
```

I know there's only one author, but this will fit better with the list and the rest of the phrasing.

policy/predicate/author.go
```diff
@@ -52,10 +64,22 @@ type OnlyHasContributorsIn struct {
 
 var _ Predicate = &OnlyHasContributorsIn{}
 
-func (pred *OnlyHasContributorsIn) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred *OnlyHasContributorsIn) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	commits, err := prctx.Commits()
+
+	predicateResult := common.PredicateResult{
+		Satisfied:       false,
+		ValuePhrase:     "Contributors",
```
bluekeyes:
```suggestion
		ValuePhrase:     "contributors",
```

policy/predicate/author.go
```diff
@@ -52,10 +64,22 @@ type OnlyHasContributorsIn struct {
...
+		ConditionPhrase: "meet the required membership conditions that contributors are only in",
```
bluekeyes:
```suggestion
		ConditionPhrase: "all meet the membership conditions",
```

policy/predicate/author_test.go
```diff
@@ -304,8 +408,7 @@ func TestAuthorIsOnlyContributor(t *testing.T) {
 
 	runAuthorTests(t, p, []AuthorTestCase{
 		{
-			"authorIsOnlyContributor",
-			true,
+			"AuthorIsOnlyContributor",
```
bluekeyes: Unintentional test name change?

policy/predicate/author_test.go
```diff
@@ -438,24 +565,36 @@ func TestAuthorIsNotOnlyContributor(t *testing.T) {
...
 type AuthorTestCase struct {
-	Name     string
-	Expected bool
-	Context  pull.Context
+	Name                    string
+	Context                 pull.Context
+	ExpectedPredicateResult *common.PredicateResult
+}
+
+func isContributorInfoEqual() {
```
bluekeyes: I think this function is unused

policy/predicate/branch_test.go
```diff
@@ -103,16 +135,22 @@ func runBranchesTestCase(t *testing.T, regex string, cases []branchesTestCase) {
...
 
 		t.Run(tc.name+" targets_branch", func(t *testing.T) {
-			ok, _, err := targetsPredicate.Evaluate(ctx, targetsContext)
+			predicateResult, err := targetsPredicate.Evaluate(ctx, targetsContext)
 			if assert.NoError(t, err, "targets_branch predicate evaluation failed") {
-				assert.Equal(t, tc.expected, ok, "targets_branch predicate was not correct")
+				assert.Equal(t, tc.ExpectedPredicateResult.Satisfied, predicateResult.Satisfied, "predicate was not correct")
+				assert.Equal(t, tc.ExpectedPredicateResult.Values, predicateResult.Values, "values were not correct")
+				assert.Equal(t, tc.ExpectedPredicateResult.ConditionsMap, predicateResult.ConditionsMap, "conditions were not correct")
+				assert.Equal(t, tc.ExpectedPredicateResult.ConditionValues, predicateResult.ConditionValues, "conditions were not correct")
```
bluekeyes: Since these assertions are the same between all tests, it would be nice to create a helper function for it to minimize duplication. You can put it in a `predicate_test.go` file (or technically, any file that ends with `_test.go`) and it will be available to all other tests in this package.

```go
func assertPredicateResult(t *testing.T, expected, actual *common.PredicateResult) {
  // ...
}
```

server/templates/details.html.tmpl
```diff
@@ -76,3 +95,62 @@
...
+  <ul class="list-star list-inside ml-2 mb-2">
+    {{range .Values}}<li>{{.}}</li>{{end}}
+  </ul>
+  <div class="mb-2">
+    {{if eq $s "skipped"}}do not{{end}} {{.ConditionPhrase}}
```
bluekeyes:
```suggestion
    {{if eq $s "skipped"}}do not {{end}}{{.ConditionPhrase}}
```

policy/predicate/author.go
```diff
@@ -29,17 +30,28 @@ type HasAuthorIn struct {
...
 		desc = fmt.Sprintf("The pull request author %q does not meet the required membership conditions", author)
 	}
 
-	return result, desc, err
+	predicateResult := common.PredicateResult{
+		Satisfied:       result,
+		Description:     desc,
+		ValuePhrase:     "Author",
+		Values:          []string{author},
+		ConditionPhrase: "meet the required membership conditions",
```
bluekeyes, line 47:
```suggestion
		ConditionPhrase: "meet the membership conditions",
```

policy/predicate/author.go
```diff
@@ -52,10 +64,22 @@ type OnlyHasContributorsIn struct {
...
+		ConditionsMap: map[string][]string{
+			"Organizations": pred.Organizations,
+			"Teams":         pred.Teams,
+			"Users":         pred.Users,
+		},
+	}
+
 	if err != nil {
-		return false, "", errors.Wrap(err, "failed to get commits")
+		return &predicateResult, errors.Wrap(err, "failed to get commits")
```
bluekeyes: In general, when returning an error, it's fine to return `nil` for other values, unless they contain information that is still useful when an error happens.

For predicates, I think we can return a nil result anytime there is any error (similar to the false and empty string of the old signature.) The approval and disapproval code only use the result when the error is nil.

policy/predicate/author.go
```diff
@@ -67,17 +91,26 @@ func (pred *OnlyHasContributorsIn) Evaluate(ctx context.Context, prctx pull.Cont
...
 		member, err := pred.IsActor(ctx, prctx, user)
 		if err != nil {
-			return false, "", err
+			return &predicateResult, err
 		}
 		if !member {
-			return false, fmt.Sprintf("Contributor %q does not meet the required membership conditions", user), nil
+			predicateResult.Description = fmt.Sprintf("Contributor %q does not meet the required membership conditions", user)
+			predicateResult.Values = []string{user}
+			return &predicateResult, nil
```
bluekeyes, line 108: In this predicate, you set `Satisifed` to `false` when creating the `PredicateResult` on L71, while in others, you set it immediately before returning.

I don't have a preference for which pattern to use, but I think it would be good to be consistent. Setting it right before returning seems to what you did most of the time.

policy/predicate/author.go
```diff
@@ -90,10 +123,22 @@ type HasContributorIn struct {
 
 var _ Predicate = &HasContributorIn{}
 
-func (pred *HasContributorIn) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred *HasContributorIn) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	commits, err := prctx.Commits()
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "Contributors",
```
bluekeyes:
```suggestion
		ValuePhrase:     "contributors",
```

policy/predicate/author.go
```diff
@@ -90,10 +123,22 @@ type HasContributorIn struct {
...
+		ConditionPhrase: "meet the required membership conditions ",
```
bluekeyes, line 131:
```suggestion
		ConditionPhrase: "meet the membership conditions ",
```

policy/predicate/author.go
```diff
@@ -127,26 +182,46 @@ type AuthorIsOnlyContributor bool
 
 var _ Predicate = AuthorIsOnlyContributor(false)
 
-func (pred AuthorIsOnlyContributor) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred AuthorIsOnlyContributor) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	commits, err := prctx.Commits()
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "author",
```
bluekeyes:
```suggestion
		ValuePhrase:     "authors",
```

policy/predicate/author.go
```diff
@@ -127,26 +182,46 @@ type AuthorIsOnlyContributor bool
...
+		ConditionPhrase: "meet",
```
bluekeyes:
```suggestion
		ConditionPhrase: "meet the condition",
```

policy/predicate/author.go
```diff
@@ -127,26 +182,46 @@ type AuthorIsOnlyContributor bool
...
+	}
+	if pred {
+		predicateResult.ConditionValues = []string{"author is the only contributor"}
+	} else {
+		predicateResult.ConditionValues = []string{"author is not the only contributor"}
+	}
```
bluekeyes:
```suggestion
	if pred {
		predicateResult.ConditionValues = []string{"they are the only contributors"}
	} else {
		predicateResult.ConditionValues = []string{"they are not the only contributors"}
	}
```

policy/predicate/branch.go
```diff
@@ -37,7 +37,16 @@ func (pred *TargetsBranch) Evaluate(ctx context.Context, prctx pull.Context) (bo
 		desc = fmt.Sprintf("Target branch %q does not match required pattern %q", targetName, pred.Pattern)
 	}
 
-	return matches, desc, nil
+	predicateResult := common.PredicateResult{
+		Satisfied:       matches,
+		Description:     desc,
+		Values:          []string{targetName},
+		ValuePhrase:     "branch",
+		ConditionPhrase: "match the required pattern",
```
bluekeyes:
```suggestion
		ValuePhrase:     "target branches",
		ConditionPhrase: "match the pattern",
```

policy/predicate/branch.go
```diff
@@ -59,7 +68,16 @@ func (pred *FromBranch) Evaluate(ctx context.Context, prctx pull.Context) (bool,
 		desc = fmt.Sprintf("Source branch %q does not match specified from_branch pattern %q", sourceBranchName, pred.Pattern)
 	}
 
-	return matches, desc, nil
+	predicateResult := common.PredicateResult{
+		Satisfied:       matches,
+		Description:     desc,
+		Values:          []string{sourceBranchName},
+		ValuePhrase:     "branch",
+		ConditionPhrase: "match the required pattern",
```
bluekeyes:
```suggestion
		ValuePhrase:     "from branches",
		ConditionPhrase: "match the pattern",
```

policy/predicate/file.go
```diff
@@ -32,24 +32,55 @@ type ChangedFiles struct {
...
+	for _, ignorePath := range pred.IgnorePaths {
+		ignorePaths = append(ignorePaths, ignorePath.String())
+	}
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "changed files",
+		ConditionPhrase: "match patterns",
+		ConditionsMap: map[string][]string{
+			"in path":        paths,
+			"while ignoring": ignorePaths,
```
bluekeyes:
```suggestion
		ConditionPhrase: "match",
		ConditionsMap: map[string][]string{
			"path patterns":        paths,
			"while ignoring": ignorePaths,
```

policy/predicate/file.go
```diff
@@ -32,24 +32,55 @@ type ChangedFiles struct {
 
 var _ Predicate = &ChangedFiles{}
 
-func (pred *ChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred *ChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	files, err := prctx.ChangedFiles()
```
bluekeyes: This should come after you create the `PredicateResult` so that it is immediately followed by the `if` statement checking the error.

policy/predicate/file.go
```diff
@@ -62,18 +93,40 @@ type OnlyChangedFiles struct {
 
 var _ Predicate = &OnlyChangedFiles{}
 
-func (pred *OnlyChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred *OnlyChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	files, err := prctx.ChangedFiles()
```
bluekeyes: Same thing here, keep this paired with the `if` statement checking the error.

policy/predicate/file.go
```diff
@@ -184,16 +255,36 @@ func (pred *ModifiedLines) Evaluate(ctx context.Context, prctx pull.Context) (bo
...
-		if !expr.IsEmpty() && expr.Evaluate(v) {
-			return true, "", nil
+	if !pred.Additions.IsEmpty() {
+		predicateResult.Values = []string{fmt.Sprintf("+%d", additions)}
+		predicateResult.ConditionValues = []string{"added lines" + pred.Additions.String()}
+		if pred.Additions.Evaluate(additions) {
+			predicateResult.Satisfied = true
+			return &predicateResult, nil
+		}
+	}
```
bluekeyes: I'd structure this as a single `switch` statement, which is a common way to write if-else changes in Go:


```go
switch {
case !pred.Additions.Empty() && pred.Additions.Evaluate(additions):
  predicateResult.Satisfied = true
  predicateResult.Values = []string{fmt.Sprintf("+%d lines", additions)}
  predicateResult.ConditionValues = []string{"added lines " + pred.Additions.String()}
  return &predicateResult, nil

case !pred.Deletions.Empty() && pred.Deletions.Evaluate(deletions):
  predicateResult.Statisfied = true
  // ...

// additional cases here
}
```

policy/predicate/file.go
```diff
@@ -184,16 +255,36 @@ func (pred *ModifiedLines) Evaluate(ctx context.Context, prctx pull.Context) (bo
...
+	if !pred.Deletions.IsEmpty() {
+		if pred.Deletions.Evaluate(deletions) {
+			predicateResult.Values = []string{fmt.Sprintf("-%d", deletions)}
+			predicateResult.ConditionValues = []string{"deleted lines" + pred.Deletions.String()}
```
bluekeyes:
```suggestion
			predicateResult.ConditionValues = []string{"deleted lines " + pred.Deletions.String()}
```

Missing space here... could also use `fmt.Sprintf("deleted lines %s", pred.Deletions)`.

policy/predicate/label.go
```diff
@@ -27,21 +27,30 @@ type HasLabels []string
 
 var _ Predicate = HasLabels([]string{})
 
-func (pred HasLabels) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred HasLabels) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
+
+	predicateResult := common.PredicateResult{
+		Satisfied:       false,
+		ValuePhrase:     "labels",
+		ConditionPhrase: "contain the required labels",
```
bluekeyes:
```suggestion
		ConditionPhrase: "contain the labels",
```

policy/predicate/signature.go
```diff
@@ -27,26 +27,40 @@ type HasValidSignatures bool
 
 var _ Predicate = HasValidSignatures(false)
 
-func (pred HasValidSignatures) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred HasValidSignatures) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	commits, err := prctx.Commits()
+
+	predicateResult := common.PredicateResult{
+		Satisfied:       false,
+		ConditionPhrase: "have valid signatures",
```
bluekeyes:
```suggestion
		ValuePhrase: "commits",
		ConditionPhrase: "have valid signatures",
```

This is one of the few cases where both `ConditionMap` and `ConditionValues` are empty. We'll want to either account for that in the template or change this to set one of them:

```go
ValuePhrase: "commits",
ConditionPhrase: "have",
ConditionValues: []string{"valid signatures"},
```

policy/predicate/signature.go
```diff
@@ -27,26 +27,40 @@ type HasValidSignatures bool
...
+
 	for _, c := range commits {
 		valid, desc := hasValidSignature(ctx, c)
+		commitHashes = append(commitHashes, c.SHA)
 		if !valid {
+			predicateResult.Values = []string{c.SHA}
 			if pred {
-				return false, desc, nil
+				predicateResult.Description = desc
+				return &predicateResult, nil
```
bluekeyes, line 52: These predicates look like another place where `Satisfied` was set to false upfront instead of right before returning.

policy/predicate/signature.go
```diff
@@ -59,33 +73,57 @@ type HasValidSignaturesBy struct {
...
 
-	signers := make(map[string]struct{})
+	signers := make(map[string]string)
+	var commitHashes []string
 
 	for _, c := range commits {
 		valid, desc := hasValidSignature(ctx, c)
 		if !valid {
-			return false, desc, nil
+			predicateResult.ValuePhrase = "commit hash"
```
bluekeyes:
```suggestion
			predicateResult.ValuePhrase = "commits"
```

policy/predicate/signature.go
```diff
@@ -59,33 +73,57 @@ type HasValidSignaturesBy struct {
 
 var _ Predicate = &HasValidSignaturesBy{}
 
-func (pred *HasValidSignaturesBy) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred *HasValidSignaturesBy) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	commits, err := prctx.Commits()
+
+	predicateResult := common.PredicateResult{
+		Satisfied:       false,
+		ConditionPhrase: "have valid signatures by",
```
bluekeyes:
```suggestion
		ConditionPhrase: "have valid signatures by members of",
```

I think this also only applies in the commit hash case. For the signer case, a better phrase might be `meet the membership conditions`

policy/predicate/signature.go
```diff
@@ -59,33 +73,57 @@ type HasValidSignaturesBy struct {
...
+		signerList = append(signerList, signer)
 		member, err := pred.IsActor(ctx, prctx, signer)
 		if err != nil {
-			return false, "", err
+			return &predicateResult, err
 		}
 		if !member {
-			return false, fmt.Sprintf("Contributor %q does not meet the required membership conditions for signing", signer), nil
+			predicateResult.Values = []string{signer}
+			predicateResult.ValuePhrase = "signer"
```
bluekeyes:
```suggestion
			predicateResult.ValuePhrase = "signers"
```

policy/predicate/signature.go
```diff
@@ -59,33 +73,57 @@ type HasValidSignaturesBy struct {
...
+			predicateResult.Description = fmt.Sprintf("Contributor %q does not meet the required membership conditions for signing", signer)
+			return &predicateResult, nil
 		}
 	}
-
-	return true, "", nil
+	predicateResult.Values = commitHashes
+	predicateResult.ValuePhrase = "commit hashes"
```
bluekeyes:
```suggestion
	predicateResult.ValuePhrase = "commits"
```

policy/predicate/signature.go
```diff
@@ -98,29 +136,53 @@ type HasValidSignaturesByKeys struct {
...
+
+	var commitHashes []string
 
 	for _, c := range commits {
 		valid, desc := hasValidSignature(ctx, c)
 		if !valid {
-			return false, desc, nil
+			predicateResult.Values = []string{c.SHA}
+			predicateResult.Description = desc
+			predicateResult.ValuePhrase = "commit"
```
bluekeyes:
```suggestion
			predicateResult.ValuePhrase = "commits"
```

policy/predicate/signature.go
```diff
@@ -98,29 +136,53 @@ type HasValidSignaturesByKeys struct {
 
 var _ Predicate = &HasValidSignaturesByKeys{}
 
-func (pred *HasValidSignaturesByKeys) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred *HasValidSignaturesByKeys) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	commits, err := prctx.Commits()
+
+	predicateResult := common.PredicateResult{
+		Satisfied:       false,
+		ConditionPhrase: "have valid signatures keys",
```
bluekeyes:
```suggestion
		ConditionPhrase: "have valid signatures by keys",
```

policy/predicate/signature.go
```diff
@@ -98,29 +136,53 @@ type HasValidSignaturesByKeys struct {
...
+			commitSHAs, ok := keys[c.Signature.KeyID]
+			if ok {
+				keys[c.Signature.KeyID] = append(commitSHAs, c.SHA)
+			} else {
+				keys[c.Signature.KeyID] = []string{c.SHA}
+			}
 		default:
-			return false, fmt.Sprintf("Commit %.10s signature is not a GPG signature", c.SHA), nil
+			predicateResult.Values = []string{"not GPG"}
+			predicateResult.ValuePhrase = "signature type"
```
bluekeyes:
```suggestion
			predicateResult.Values = []string{c.SHA}
			predicateResult.ValuePhrase = "commits" // this might already be set
			predicateResult.ConditionPhrase = "have GPG signatures"
```

policy/predicate/signature.go
```diff
@@ -129,11 +191,17 @@ func (pred *HasValidSignaturesByKeys) Evaluate(ctx context.Context, prctx pull.C
 			}
 		}
 		if !isValidKey {
-			return false, fmt.Sprintf("Key %q does not meet the required key conditions for signing", key), nil
+			predicateResult.Values = []string{key}
+			predicateResult.ValuePhrase = "key"
```
bluekeyes:
```suggestion
			predicateResult.ValuePhrase = "keys"
			predicateResult.ConditionPhrase = "exist in the set of allowed keys"
```

policy/predicate/status.go
```diff
@@ -27,10 +27,17 @@ type HasSuccessfulStatus []string
...
 var _ Predicate = HasSuccessfulStatus([]string{})
 
-func (pred HasSuccessfulStatus) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred HasSuccessfulStatus) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	statuses, err := prctx.LatestStatuses()
+
+	predicateResult := common.PredicateResult{
+		Satisfied:       false,
+		ValuePhrase:     "check",
+		ConditionPhrase: "has successful status",
```
bluekeyes:
```suggestion
		ValuePhrase:     "status checks",
		ConditionPhrase: "exist and pass",
```

Here's another place that doesn't set `ConditionMap` or `ConditionValues`

policy/predicate/status.go
```diff
@@ -46,13 +53,21 @@ func (pred HasSuccessfulStatus) Evaluate(ctx context.Context, prctx pull.Context
 	}
 
 	if len(missingResults) > 0 {
-		return false, "One or more statuses is missing: " + strings.Join(missingResults, ", "), nil
+		predicateResult.Values = missingResults
+		predicateResult.Description = "One or more statuses is missing: " + strings.Join(missingResults, ", ")
+		return &predicateResult, nil
```
bluekeyes, line 58: `Satisfied` set to false upfront instead of at return

policy/predicate/title.go
```diff
@@ -28,22 +28,45 @@ type Title struct {
...
 	title := prctx.Title()
 
+	predicateResult := common.PredicateResult{
+		Satisfied:       false,
+		ValuePhrase:     "title",
+		Values:          []string{title},
+		ConditionPhrase: "meet the pattern requirement",
+	}
+
+	var MatchPatterns, NotMatchPatterns []string
```
bluekeyes: style: local variables start with lowercase letters by convention (`matchPatterns`, `notMatchPatterns`)

policy/predicate/title.go
```diff
@@ -28,22 +28,45 @@ type Title struct {
 
 var _ Predicate = Title{}
 
-func (pred Title) Evaluate(ctx context.Context, prctx pull.Context) (bool, string, error) {
+func (pred Title) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
 	title := prctx.Title()
 
+	predicateResult := common.PredicateResult{
+		Satisfied:       false,
+		ValuePhrase:     "title",
```
bluekeyes:
```suggestion
		ValuePhrase:     "titles",
```

bluekeyes (changes requested):
Sorry for the flood of comments, but from a code perspective this looks good! Thanks for implementing the `PredicateResult` changes. Almost all of my suggestions are minor things to improve the phrasing of messages. There are also a couple style things that would be good to address.

Beyond that, we can talk about any CSS changes tomorrow and then I think this will be ready to merge.

There's a few more things we could do to make the messages read even better, but I think the current implementation (with the suggestions in this review) is good enough for an initial version, so I don't want to delay this any further.

policy/approval/approve.go
bluekeyes: Yeah, that's correct

policy/predicate/predicate.go
bluekeyes: Yeah, I think this is addressed by the updated code. Everything checks the `error` value first, so as long as the `*PredicateResult` is non-nil when the error is nil, we're following standard Go conventions.

server/templates/details.html.tmpl
bluekeyes: That's fair, let's leave it as "requires review" for now.

policy/predicate/title.go
yiweny: shouldn't there be only 1 title for each pr?

policy/predicate/author_test.go
yiweny: Yeah.

policy/predicate/author.go
yiweny: Yeah makes sense.

policy/predicate/author_test.go
yiweny: Thank you for pointing out! I must have forgotten about this.

policy/predicate/author.go
yiweny: What is your reason for removing the `required` here? I was following how the description was written on line 39. https://github.com/palantir/policy-bot/pull/407/files#diff-295712a7931b2fd620f63c7cfb9ed034b59530d1b9a3534a1da08599e108f93dR39

policy/predicate/author.go
yiweny: Same as above comment, the description field is using the term `required membership`. What makes you think it is better to remove the required here

policy/predicate/file.go
yiweny: I don't think I can put it in a `switch` statement actually. Because I am accumulating the `Values` and `ConditionValues` in each if condition. So in the case the PR does not satisfy the predicate, The `Values` and `ConditionValues` would contain information about all the three cases if not empty. If it satisfy the predicate, then it would only return the relevant condition. I would not be able to accumulate values using a `switch` statement.

bluekeyes: Sorry, you're right, I missed the accumulation that was happening.

policy/predicate/author.go
bluekeyes: As I mentioned on Slack, I thought the context of the message implied that the condition was required, but leaving "required" here is fine too.

policy/predicate/signature.go
```diff
@@ -98,29 +141,55 @@ type HasValidSignaturesByKeys struct {
...
+			predicateResult.Values = []string{c.SHA}
+			predicateResult.ValuePhrase = "commits"
+			predicateResult.ConditionPhrase = "have GPG signatures"
+			predicateResult.Description = fmt.Sprintf("Commit %.10s signature is not a GPG signature", c.SHA)
+			predicateResult.Satisfied = false
+			return &predicateResult, nil
 		}
 	}
 
+	var keyList []string
```
bluekeyes: Should this be used somewhere? Right now, it looks like you only append to it and never use the resulting list.

server/assets/css/main.css
```diff
@@ -177,6 +185,14 @@ a:hover {
...
 }
 
+.list-numbers > li {
+  @apply ml-2;
+}
+
+.list-star > li:before {
+  content: "*";
+}
+
```
bluekeyes: I think both of these rules are unused now

server/handler/frontend.go
```diff
@@ -73,3 +82,26 @@ func Static(prefix string, c *FilesConfig) http.Handler {
 
 	return http.StripPrefix(prefix, http.FileServer(http.Dir(dir)))
 }
+
+func hasRequires(result common.Actors) bool {
+	if len(result.Organizations) != 0 || len(result.Teams) != 0 || len(result.Users) != 0 {
```
bluekeyes: I meant to comment on this earlier, but forgot - `common.Actors` can also select users by their permissions level, which is a condition we're ignoring at the moment.

I don't think it should be part of this PR, but it would be nice to make a followup change to add minimal support for rules that use permissions. I don't think it's common in the policies you have been using for testing but is pretty common in other policies.

When we do add that support, I think you can replace this function with the existing `IsEmpty()` method on `common.Actors`.

server/templates/details.html.tmpl
```diff
@@ -56,6 +56,24 @@
 <li data-status="{{$s}}">
   <div class="bg-white p-2 shadow-sm max-w-lg status-stripe {{$s}}">
     {{template "result-details" .}}
+    {{if (or .PredicateResults (hasRequires .Requires))}}
+    <details class="bg-light-gray5 p-2 mt-2 text-sm">
+      <summary class="text-dark-gray3 font-semibold">Details</summary>
+      <div class="pt-2"></div>
+      <div class="border-t border-light-gray3">
```
bluekeyes: I think you should be able to combine these `div`s by adding `mt-2` to the `div` with the border. I like to avoid empty `div` when I can, and it should be possible here.

server/templates/details.html.tmpl
```diff
@@ -56,6 +56,24 @@
...
+       {{if .PredicateResults}}
+          <div class="pt-2 mt-2">
```
bluekeyes: I think this only needs `pt-2` now that the border is defined on a different element. That will also match the requires `div` below.

bluekeyes (commented):
Thanks for working through all the iterations here, this is looking good! I spotted two things that I think are unused and had some minor suggestions on the CSS (although I haven't tested the changes.) Once those are fixed, I think this is ready to merge.

policy/predicate/signature.go
yiweny: Yeah. Previously I returned both keys and the commit hashes. Since now we only want commit hashes, this should be removed. Good catch.

bluekeyes (approved):

PR #413, opened 2022-04-29
Improve details UI for some predicates

bluekeyes:
These are some minor adjustments that came up in testing:

* Replace "from branches" with "source branches" to make the resulting sentence less awkward
* Add padding to condition maps and rephrase the changed files condition to separate the top-level condition phrase from the sub-phrases in the map
* Add a special case for predicates that have no values to avoid rendering an empty list. I don't think it is necessary to show the condition details when there are no values to compare with. An alternative would be to render something like `[no {{.ValuePhrase}}]` as the only list element when there are no values, but I thought that could potentially get confused with actual values.


asvoboda (approved):

yiweny (approved):
Thank you for taking time to improve the UI!

PR #423, opened 2022-05-14
Include permissions in required review details

### LINKED ISSUE #416: Fix UI for rules with required review by permissions
When reviews are required by permissions the UI does not display that fact:
Screen Shot 2022-05-03 at 11 56 25 AM

Rule:
```
  - name: one admin has approved (PR contributors not allowed)
    options:
      allow_contributor: false
    requires:
      count: 1
      permissions: ["admin", "maintain"]
```

bluekeyes:
Ideally this would show the teams or users that have these permissions, but just displaying the permissions will be better than an empty spot in the UI.

Also remove an incorrect comment and add a helper function to translate the deprecated permissions to the new model since this happened in multiple places.

Fixes #416. Marking as a draft for now because I haven't tested it at all yet.

### DIFF
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -154,19 +154,11 @@ func (r *Rule) getReviewRequestRule() *common.ReviewRequestRule {
 		mode = common.RequestModeRandomUsers
 	}
 
-	perms := append([]pull.Permission(nil), r.Requires.Permissions...)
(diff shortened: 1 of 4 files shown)


bluekeyes:
Closing in favor of #425 (for the UI side) and #426 (for the permissions refactor.)

PR #425, opened 2022-05-18
Render required permissions

### LINKED ISSUE #416: Fix UI for rules with required review by permissions

ylee088:
Fixes #416.


server/handler/frontend.go
```diff
@@ -100,3 +106,11 @@ func getRequires(result *common.Result, githubURL string) map[string][]Membershi
 	}
 	return membershipInfo
 }
+
+func getPermissions(result *common.Result) []string {
+	permStrings := make([]string, len(result.Requires.Permissions))
+	for _, perm := range result.Requires.Permissions {
```
bluekeyes: I think you'll want to use the new `GetPermissions` function I added in #426 (once it merges) here to make sure we're including all permissions and there are no duplicates.

ylee088: Fixed, thank you!

ylee088: Oh I totally missed the "once it merges" part. Failing now, but will wait until #426 is merged.

server/handler/frontend.go
```diff
@@ -100,3 +106,11 @@ func getRequires(result *common.Result, githubURL string) map[string][]Membershi
 	}
 	return membershipInfo
 }
+
+func getPermissions(result *common.Result) []string {
+	permStrings := make([]string, len(result.Requires.Permissions))
+	for _, perm := range result.Requires.GetPermissions() {
+		permStrings = append(permStrings, perm.String())
+	}
```
bluekeyes: When allocating a slice upfront and then filling it, you have two options

1. Use `make([]string, 0, length)`, then use `append` in the loop
2. Use `make([]string, length)`, then set elements by index in the loop

As written, this will create a slice with `len(result.Requires.Permissions)` empty strings, then append the new string on to the end of that.

I think you'll also want to initialize the slice using the length of `result.Requires.GetPermissions()` instead of `result.Requires.Permissions`.

All that said, another option is to remove this function and turn the `pull.Permission` types in to strings in the template, using the `printf` template function.

server/templates/details.html.tmpl
```diff
@@ -148,3 +153,15 @@
...
+{{define "result-requires-permissions-details"}}
+  <b class="font-bold text-sm">This rule requires review from users with the following permissions</b>
+  <dl class="my-2">
+  {{range $k, $v := getPermissions .}}
+    <dt> {{$k}}</dt>
+    <dd>
+        <ul class="list-disc list-outside pl-6 py-2">{{range $v}}<li class="font-mono text-sm-mono">{{.}}</li>{{end}}</ul>
+    </dd>
+  {{end}}
+  </dl>
```
bluekeyes: Unlike the user/team/org section, `getPermissions` returns a `[]string` instead of a `map[string][]string`. I think you'll want to use a single `ul` rendered with one `range` expression for this, instead of the description list pattern.

bluekeyes:
Also, looks like `server/.DS_Store` was added incorrectly - can you remove that file?

server/handler/frontend.go
ylee088: Fixed, using option 1.

server/templates/details.html.tmpl
ylee088: Fixed, copy-pasta laziness :p

server/handler/frontend.go
```diff
@@ -61,11 +61,17 @@ func LoadTemplates(c *FilesConfig, basePath string, githubURL string) (templatet
...
 		},
 		"hasRequires": func(requires common.Actors) bool {
-			return !requires.IsEmpty()
+			return len(requires.Users) > 0 || len(requires.Teams) > 0 || len(requires.Organizations) > 0
 		},
 		"getRequires": func(results *common.Result) map[string][]Membership {
 			return getRequires(results, strings.TrimSuffix(githubURL, "/"))
 		},
+		"hasRequiresPermissions": func(requires common.Actors) bool {
+			return len(requires.Permissions) > 0
```
bluekeyes: I think this should check the result of `requires.GetPermissions()` to make sure it accounts for the legacy properties.

server/handler/frontend.go
```diff
@@ -100,3 +106,11 @@ func getRequires(result *common.Result, githubURL string) map[string][]Membershi
 	}
 	return membershipInfo
 }
+
+func getPermissions(result *common.Result) []string {
+	permStrings := make([]string, 0, len(result.Requires.GetPermissions()))
+	for _, perm := range result.Requires.GetPermissions() {
+		permStrings = append(permStrings, perm.String())
+	}
```
bluekeyes:
```suggestion
	perms := result.Requires.GetPermissions()
	permStrings := make([]string, 0, len(perms))
	for _, perm := range perms {
		permStrings = append(permStrings, perm.String())
	}
```

server/templates/details.html.tmpl
```diff
@@ -148,3 +153,10 @@
   {{end}}
   </dl>
 {{end}}
+
+{{define "result-requires-permissions-details"}}
+  <b class="font-bold text-sm">This rule requires review from users with the following permissions</b>
+  {{range $k, $v := getPermissions .}}
+    <ul class="list-disc list-outside pl-6 py-2">{{range $v}}<li class="font-mono text-sm-mono">{{.}}</li>{{end}}</ul>
+  {{end}}
```
bluekeyes:
```suggestion
  <ul class="list-disc list-outside pl-6 py-2">
    {{range getPermissions .}}<li class="font-mono text-sm-mono">{{.}}</li>{{end}}
  </ul>
```

server/handler/frontend.go
ylee088: Sorry, keep missing these!

bluekeyes (approved):

PR #427, opened 2022-05-20
Fix or condition for the details section to render permissions.

ylee088:


ylee088:
Screen Shot 2022-05-20 at 4 16 26 PM

server/templates/details.html.tmpl
```diff
@@ -56,7 +56,7 @@
 <li data-status="{{$s}}">
   <div class="bg-white p-2 shadow-sm max-w-lg status-stripe {{$s}}">
     {{template "result-details" .}}
-    {{if (or .PredicateResults (hasRequires .Requires))}}
+    {{if (or (.PredicateResults) (or (hasRequires .Requires) (hasRequiresPermissions .Requires)))}}
```
bluekeyes: In templates `or` can take multiple arguments, so I think this will work:

```suggestion
    {{if (or (.PredicateResults) (hasRequires .Requires) (hasRequiresPermissions .Requires))}}
```

ylee088: Fixed!

bluekeyes (approved):

PR #447, opened 2022-06-29
Fix issue with method defaults

### LINKED ISSUE #434: Overriding approval methods clears all defaults
We had a problem today where a policy added alternate comments for some rules, but expected the default behavior of counting GitHub reviews to still work. With the current implementation, setting any property of the `methods` block removes the defaults for all properties.

Consider setting defaults for fields that are not explicitly set by the policy file to reduce confusion, as the docs imply.

shravan1k:
When certain attributes are defined under 'methods', the other undefined attributes don't line up with the promised defaults. This PR resolves this issue by checking whether certain attributes were not specified and then adds defaults for them.

Closes #434

### DIFF
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -57,18 +57,19 @@ type RequestReview struct {
 
 func (opts *Options) GetMethods() *common.Methods {
 	methods := opts.Methods
-	defaultBool := true
+	defaultGithubReview := true
 	if methods == nil {
 		methods = &common.Methods{}
 	}
-	if len(methods.Comments) == 0 {
-		methods.Comments = []string{
+	if methods.Comments == nil {
+		defaultComments := []string{
 			":+1:",
 			"👍",
 		}
+		methods.Comments = &defaultComments
 	}
 	if methods.GithubReview == nil {
-		methods.GithubReview = &defaultBool
+		methods.GithubReview = &defaultGithubReview
 	}
 
 	methods.GithubReviewState = pull.ReviewApproved
@@ -86,10 +87,10 @@ func (r *Rule) Trigger() common.Trigger {
 
 	if r.Requires.Count > 0 {
 		m := r.Options.GetMethods()
-		if len(m.Comments) > 0 || len(m.CommentPatterns) > 0 {
+		if m.Comments != nil && len(*m.Comments) > 0 || len(m.CommentPatterns) > 0 {
 			t |= common.TriggerComment
 		}
-		if *m.GithubReview || len(m.GithubReviewCommentPatterns) > 0 {
+		if m.GithubReview != nil && *m.GithubReview || len(m.GithubReviewCommentPatterns) > 0 {
 			t |= common.TriggerReview
 		}
 	}
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -57,7 +57,6 @@ type RequestReview struct {
 
 func (opts *Options) GetMethods() *common.Methods {
 	methods := opts.Methods
-	defaultGithubReview := true
 	if methods == nil {
 		methods = &common.Methods{}
 	}
@@ -69,6 +68,7 @@ func (opts *Options) GetMethods() *common.Methods {
 		methods.Comments = &defaultComments
 	}
 	if methods.GithubReview == nil {
+		defaultGithubReview := true
 		methods.GithubReview = &defaultGithubReview
 	}
 
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -61,11 +61,10 @@ func (opts *Options) GetMethods() *common.Methods {
 		methods = &common.Methods{}
 	}
 	if methods.Comments == nil {
-		defaultComments := []string{
+		methods.Comments = []string{
 			":+1:",
 			"👍",
 		}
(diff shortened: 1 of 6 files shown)


policy/approval/approve.go
```diff
@@ -57,13 +57,24 @@ type RequestReview struct {
...
 		methods = &common.Methods{
 			Comments: []string{
 				":+1:",
 				"👍",
 			},
-			GithubReview: true,
+			GithubReview: &defaultBool,
+		}
+	} else {
+		if len(methods.Comments) == 0 {
```
asvoboda: I think you could simplify both the if and the else into one. Effectively, if `methods == nil`, then `methods.Comments` and `methods.GithubReview` will also be nil.

policy/common/methods.go
```diff
@@ -25,7 +25,7 @@ import (
 type Methods struct {
 	Comments                    []string `yaml:"comments,omitempty"`
 	CommentPatterns             []Regexp `yaml:"comment_patterns,omitempty"`
-	GithubReview                bool     `yaml:"github_review,omitempty"`
+	GithubReview                *bool    `yaml:"github_review,omitempty"`
```
asvoboda, line 28: Is the idea here that this is changed to a pointer to signal the difference between unset vs default false?

shravan1k: Yeah

policy/approval/approve.go
shravan1k: I agree, will make the change.

policy/approval/approve.go
```diff
@@ -57,15 +57,19 @@ type RequestReview struct {
 
 func (opts *Options) GetMethods() *common.Methods {
 	methods := opts.Methods
+	defaultBool := true
```
derekjobst: Let's use a more descriptive name here and in the other instances where `defaultBool` has been added. Perhaps `defaultGithubReview`?

policy/approval/approve.go
```diff
@@ -57,15 +57,19 @@ type RequestReview struct {
...
-				":+1:",
-				"👍",
-			},
-			GithubReview: true,
+		methods = &common.Methods{}
+	}
+	if len(methods.Comments) == 0 {
+		methods.Comments = []string{
+			":+1:",
+			"👍",
```
derekjobst, line 66: Will this allow a policy to be configured without comment approvals? In other words, will it still be possible to configure a policy where only github reviews (i.e. `github_review=true`) are acceptable forms of approval?

shravan1k: Good catch. New commit resolves this.

policy/common/methods.go
```diff
@@ -23,10 +23,10 @@ import (
 )
 
 type Methods struct {
-	Comments                    []string `yaml:"comments,omitempty"`
-	CommentPatterns             []Regexp `yaml:"comment_patterns,omitempty"`
-	GithubReview                bool     `yaml:"github_review,omitempty"`
-	GithubReviewCommentPatterns []Regexp `yaml:"github_review_comment_patterns,omitempty"`
+	Comments                    *[]string `yaml:"comments,omitempty"`
```
bluekeyes: Making this a pointer is not necessary. In Go, slices implicitly work like pointers but we don't usually think about it because a `nil` slice behaves the same as a non-`nil` empty slice. When a user does not specify the `comments` field in configuration, the value will be `nil`.

policy/approval/approve.go
```diff
@@ -82,10 +87,10 @@ func (r *Rule) Trigger() common.Trigger {
 
 	if r.Requires.Count > 0 {
 		m := r.Options.GetMethods()
-		if len(m.Comments) > 0 || len(m.CommentPatterns) > 0 {
+		if m.Comments != nil && len(*m.Comments) > 0 || len(m.CommentPatterns) > 0 {
```
bluekeyes: These `nil` checks and dereferencing are not necessary once `m.Comments` is reverted back to a `[]string` because `len(nil) == 0`

policy/common/methods.go
```diff
@@ -120,9 +120,11 @@ func deduplicateCandidates(all []*Candidate) []*Candidate {
 }
 
 func (m *Methods) CommentMatches(commentBody string) bool {
-	for _, comment := range m.Comments {
-		if strings.Contains(commentBody, comment) {
-			return true
+	if m.Comments != nil {
```
bluekeyes: Same idea here with the nil check and dereferencing: you can `range` over a `nil` slice without checking first.

bluekeyes (commented):
One comment about the slice handling, but otherwise this looks good.

We'll want to be careful about noting this in the release notes because it is technically a breaking change. As Derek alluded to, if I have a policy like:

```yaml
options:
  methods:
    github_review: true
```

it currently disables comments. In order to keep the same behavior after this PR, I'll have to change the policy to:

```yaml
options:
  methods:
    comments: []
    github_review:  true
```

This is consistent with what the docs have always claimed, so I'd count this a bug fix, but people might have been relying on the old broken behavior. We might want to audit at least our internal policies to see if this will impact anything.

policy/common/methods.go
shravan1k: Makes sense. Will revert that back.

bluekeyes (approved):

PR #448, opened 2022-07-01
feature: Validate incoming policy definition in PR changes

### LINKED ISSUE #362: policy-bot should validate policy changes
Currently the policy-bot check validates if the policy on the target branch is satisfied. In cases where the PR is modifying `policy.yml` the policy-bot check should also validate that the new policy is valid via `<policy-bot url>/api/validate` (valid meaning well formatted, not satisfied).

bmoylan:
Fixes https://github.com/palantir/policy-bot/issues/362

I looked for tests that exercise this code but couldn't find them. What is the best way to test this?

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -180,6 +180,9 @@ func (b *Base) Evaluate(ctx context.Context, installationID int64, trigger commo
 	if evaluator == nil {
 		return nil
 	}
+	if err := b.ValidateChangedHeadConfig(ctx, prctx, client, trigger); err != nil {
+		return err
+	}
 
 	result, err := b.EvaluateFetchedConfig(ctx, prctx, client, evaluator, fetchedConfig)
 	if err != nil {
@@ -266,6 +269,29 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
 	b.PostStatus(ctx, prctx, client, statusState, statusDescription)
 
 	return result, nil
+}
+
+func (b *Base) ValidateChangedHeadConfig(ctx context.Context, prctx pull.Context, client *github.Client, trigger common.Trigger) error {
+	logger := zerolog.Ctx(ctx)
+
+	files, err := prctx.ChangedFiles()
+	if err != nil {
+		return err
+	}
+	for _, f := range files {
+		switch f.Status {
+		case pull.FileAdded, pull.FileModified:
+			switch f.Filename {
+			case b.PullOpts.PolicyPath, ".github/" + b.PullOpts.PolicyPath:
+				logger.Debug().Msg("Pull request changed policy file. Evaluating head version.")
+
+				fetchedConfig := b.ConfigFetcher.ConfigForPRHead(ctx, prctx, client)
+				_, err = b.ValidateFetchedConfig(ctx, prctx, client, fetchedConfig, trigger)
+				return err
+			}
+		}
+	}
+	return nil
 }
 
 func (b *Base) RequestReviewsForResult(ctx context.Context, prctx pull.Context, client *github.Client, trigger common.Trigger, result common.Result) error {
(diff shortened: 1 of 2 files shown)


bluekeyes:
Unfortunately, the best way to test handler code is with a running instance. We don't have any kind of tests that use mock GitHub events at this time. If you don't have a local dev instance, we can probably test in staging after merging this.

server/handler/base.go
```diff
@@ -268,6 +271,29 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
 	return result, nil
 }
 
+func (b *Base) ValidateChangedHeadConfig(ctx context.Context, prctx pull.Context, client *github.Client, trigger common.Trigger) error {
+	logger := zerolog.Ctx(ctx)
+
+	files, err := prctx.ChangedFiles()
+	if err != nil {
+		return err
```
bluekeyes, line 279: One of the conditions this can fail in is when a pull request modifies more than 3000 files. My instinct is that validation of policy changes should be best effort and that we should not fail the overall evaluation for conditions like this if the actual policy may not be impacted by the same limit.

What do you think? I can see how it might be misleading if an invalid policy resulted in a passing check due to an error with the GitHub API or something.

server/handler/base.go
```diff
@@ -268,6 +271,29 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
...
+	}
+	for _, f := range files {
+		switch f.Status {
+		case pull.FileAdded, pull.FileModified:
+			switch f.Filename {
+			case b.PullOpts.PolicyPath, ".github/" + b.PullOpts.PolicyPath:
```
bluekeyes, line 285: I think `PolicyPath` is the only path that we need to check for here

server/handler/base.go
```diff
@@ -268,6 +271,29 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
...
+				logger.Debug().Msg("Pull request changed policy file. Evaluating head version.")
+
+				fetchedConfig := b.ConfigFetcher.ConfigForPRHead(ctx, prctx, client)
+				_, err = b.ValidateFetchedConfig(ctx, prctx, client, fetchedConfig, trigger)
+				return err
```
bluekeyes, line 290: I'm not sure if the error returned from `ValidateFetchedConfig` makes it clear where the error was. We may want to add an additional message to clarify that this is from the policy being modified in the pull request, rather than the existing policy in the repository.

bluekeyes (commented):
One reason I've been a bit hesitant about this feature is that it will cause all evaluations to load all the files in the pull request. If the policy doesn't otherwise have any rules that use file paths, this could be many extra requests for a limited use case. But in practice, I'm not sure what percentage of policies have no file conditions.

I'm trying to think of way to optimize this without adding too much complexity. One thought is to look at the incoming trigger and only validate files for triggers that could have changed content (i.e. `TriggerCommit`.) Unfortunately, I think this means that if you have an invalid policy, and then someone approves the PR or another evaluation happens, it could overwrite the failure from the bad policy.

Maybe there's an alternate way to check for policy modifications? Always fetching the policy file from the head branch and then comparing the content against the base branch might be faster.

stevoland:
Hi both,

We would like to help contribute an implementation that addresses the comments above:

1. Validation only happens if all policies are otherwise approved (optimise for most common case)
2. Validation always fetches the config from the head branch (checking if it has changed would often be an additional request)
3. (Optional) Failure while fetching config fails the check with details in the UI (I believe going to the details page re-evaluates the check and would thus resolve a temporary failure?)

Let me know your thoughts, thanks!

bluekeyes:
@stevoland I'm happy to work with you on a contribution if this is something you'd like to implement. I think this is no longer a priority for either Brad or myself because internally, we mostly use shared policies and have a verification step in CI for the shared policy repos.

Let's continue the discussion of the implementation on #362, but your proposal sounds like a good starting point to me. Fetching only the policy file from the head branch should avoid my concern about listing all the pull request files. I'm less sure about only validating the policy after all other rules pass. It might be annoying to find out you have to fix a policy after getting all the approvals you need and then have to get the same approvals again after pushing the fix. I suppose it is still better than merging a broken policy file though. Maybe if we only ran this validation check on specific `pull_request` events (like `created` and `synchronized`), that would be sufficient?

Issue #417, opened 2022-05-06
Feature: Include PR Body in Approval Candidates


agirlnamedsophia:
Thank you for building such a great tool, and already accepting several contributions from my team. We love it so far!

Part of our workflow involves providing a way for PR authors to opt out of some review requirements. Historically we've done this by scanning the PR Body Description for specific comments. It would be great to add this feature to PolicyBot by including the PR Body in the API Response and appending it to the list of comments that are used for approval analysis. 

I can see this as pairing well with the "allow_author" feature:
```
options:
  # If true, approvals by the author of a pull request are considered when
  # calculating the status. False by default.
  allow_author: true
```
This feature addition would mean the author doesn't have to add an extra "issue comment" after they've opened a PR with a specific comment type, they can just add it into the PR body description.
This might also be useful to other people who use bots to create/merge PRs who don't need specific types of review. 

I'd be happy to take a stab at implementing this feature and opening a PR if you agree that this would be valuable.

bluekeyes:
Thanks for using and contributing to Policy Bot! Before you implement this feature, I think it would help to have an example policy and workflow where approval in the PR body would help. What you described so far sounds like something we typically use predicates for, so while we don't have a PR body predicate right now, it may make more sense to add this as a new predicate type than as an additional source of approval candidates.

For an example with predicates, consider a case where authors apply labels to indicate they're opting out of review:

```yaml
policy:
  approval:
    - or:
        - reviewers have approved
        - pull request has the no-review label
        
approval_rules:
  - name: reviewers have approved
    requires:
      count: 1
      teams: ["palantir/reviewers"]

  - name: pull request has the no-review label
    if:
      has_labels:
        - "no-review"
    requires:
      count: 0 # indicates "auto-approval"
```

You can imagine combing this with other predicates as well: perhaps the author has to be part of a certain team or the pull request has to only change certain files. If labels don't work for you, maybe a new `body` predicate (similar to the `title` predicate) could work.

To summarize, instead of having many ways for people to self-approve, Policy Bot is designed to encode conditions like this in the policy by using predicates and the logical operators to enable rules that require no reviews. I'm having trouble thinking of situations where it makes more sense to self-approve as part of the initial PR.

But there could be valid workflows that don't 't fit this model or that I haven't thought about, so if you think that's the case here, I'd like to learn more!

agirlnamedsophia:
[...]

PR #454, opened 2022-07-22
Feature: Include PR Body in Approval Candidates

agirlnamedsophia:
This adds a solution for https://github.com/palantir/policy-bot/issues/417. 

As discussed in the attached issue, this adds support for reading the PR Body for approval patterns. 
It also invalidates a current approval if the body is edited, if the user has `ignore_edited_comments` in their policy options list.

Let me know if this looks okay. Especially curious if there are other better ways to have tested this implementation. Thank you again for considering this!


policy/approval/approve.go
```diff
@@ -296,6 +304,28 @@ func (r *Rule) filterEditedCandidates(ctx context.Context, prctx pull.Context, c
 	return allowedCandidates, nil
 }
 
+func (r *Rule) filterEditedBodyCandidates(ctx context.Context, prctx pull.Context, candidates []*common.Candidate) ([]*common.Candidate, error) {
```
bluekeyes: With the exception of the flag that's checked and part of a log message, this function seems identical to `filterEditedCommentCandidates`. I think it would be good to create a single function that can handle both case. Because both of these functions are only called after checking the option value, I don't think there's a reason to additionally check the option again within the function, so that removes one difference. For the log message, maybe add a `candidateType` parameter or similar?

This also raises a question about the `ignore_edited_body` option. Do you think it is helpful to control body edits independently from comment edits or would a single option (the existing `ignore_edited_comments`) be sufficient?

policy/common/methods.go
```diff
@@ -72,6 +73,29 @@ func (m *Methods) Candidates(ctx context.Context, prctx pull.Context) ([]*Candid
...
+		if m.BodyMatches(prBody.Body) {
+			createdAt := prBody.CreatedAt
+			lastEditedAt := prBody.LastEditedAt
+			var updatedAt time.Time
+
+			if !lastEditedAt.IsZero() {
+				updatedAt = lastEditedAt
+			} else {
+				updatedAt = createdAt
+			}
```
bluekeyes: style: I think this could be simplified:

```suggestion
			updatedAt := prBody.LastEditedAt
			if updatedAt.IsZero() {
				updatedAt = createdAt
			}
```

pull/context.go
```diff
@@ -233,3 +236,11 @@ type CollaboratorPermission struct {
 	// repository. If false, the permission is granted by the organization.
 	ViaRepo bool
 }
+
+type PrBody struct {
```
bluekeyes: style: `PRBody` or possibly just `Body`

pull/github.go
```diff
@@ -185,6 +185,40 @@ func (ghc *GitHubContext) Title() string {
 	return ghc.pr.Title
 }
 
+// if adding new fields to this struct, modify Locator#toV4() and Locator#IsComplete() as well
```
bluekeyes: I think this comment is not relevant for this new type

pull/github.go
```diff
@@ -705,6 +739,15 @@ func (ghc *GitHubContext) loadPagedData() error {
...
 			PullRequest struct {
+				CreatedAt    time.Time
+				LastEditedAt time.Time
+				Body         string
+				Editor       struct {
+					Login string
+				}
+				Author struct {
+					Login string
+				}
```
bluekeyes: I think these fields are unused in this query

bluekeyes (commented):
Thanks for implementing this feature and adding new tests! I had a few minor style/cleanup suggestions, but once those are fixed I think this should be good to merge.

policy/approval/approve.go
agirlnamedsophia: > This also raises a question about the ignore_edited_body option. Do you think it is helpful to control body edits independently from comment edits or would a single option (the existing ignore_edited_comments) be sufficient?

that's an excellent question. I *thiink* with how we wanted to use PR body values it would be ideal to have them function as separate entities. But I can see how it would also make sense to lump the pr into the single option of `ignore_edited_comments`.

I'll refactor the method to use a single one though and see how it looks from there.

agirlnamedsophia:
@bluekeyes thanks for your speedy feedback! I have implemented some changes. Let me knokw what you think.

policy/approval/approve.go
```diff
@@ -258,7 +259,14 @@ func (r *Rule) filteredCandidates(ctx context.Context, prctx pull.Context) ([]*c
...
 
 	if r.Options.IgnoreEditedComments {
-		candidates, err = r.filterEditedCandidates(ctx, prctx, candidates)
+		candidates, err = r.filterEditedCandidates(ctx, prctx, candidates, "comments")
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if r.Options.IgnoreEditedBody {
```
bluekeyes: Looking at this again, the way the "candidates" abstraction currently works means that we can only have a single option for ignoring edited candidates. Because `filterEditedCandidates` can't tell if a candidate came from a comment or a body, the first call to `filterEditedCandidates` will remove all edits, regardless of which flag is set.

If you think a single flag will work for your use case, let's remove the `ignore_edited_body` flag and update the docs for `ignore_edited_comments` to clarify that it applies to bodies too. If you think separate flags are required, I think we'll have to introduce a `Type` field  and a `CandidateType` "enum" to the `Candidate` struct so that the filtering can distinguish between types of candidates if needed.

bluekeyes (commented):
Thanks for the updates! The code looks good to me now, but on reading it again I realized there's still an issue with the dual flags and candidate filtering. I think the solution there will depend on your requirements for this feature.

devinburnette:
we also want to add something like this to the rule Trigger method [here](https://github.com/palantir/policy-bot/blob/develop/policy/approval/approve.go#L84), so that updates will automatically reflect on the PR instead of having to check the details page to trigger all:  
```
		if len(m.BodyPatterns) > 0 {
			t |= common.TriggerPullRequest
		}

```

agirlnamedsophia:
Thanks for the feedback @devinburnette and @bluekeyes - I've finally made the changes. Sorry for the delay, was on a long holiday.

pull/context.go
```diff
@@ -233,3 +236,11 @@ type CollaboratorPermission struct {
 	// repository. If false, the permission is granted by the organization.
 	ViaRepo bool
 }
+
+type Body struct {
+	Body         string
+	CreatedAt    time.Time
+	Editor       string
```
devinburnette: looks like we don't actually use editor anywhere, i think we were thinking about making a comparison in the ignore edit logic to allow edits to the body if the editor was the same as the the author.

in that case, making candidate type a thing like @bluekeyes suggested is probably worth considering. do you think that's still valuable?

devinburnette:
this lgtm

pull/github.go
```diff
@@ -185,6 +185,37 @@ func (ghc *GitHubContext) Title() string {
...
+			PullRequest v4PullRequestWithEditedAt `graphql:"pullRequest(number: $number)"`
+		} `graphql:"repository(owner: $owner, name: $name)"`
+	}
+	qvars := map[string]interface{}{
+		"owner":  githubv4.String(ghc.owner),
+		"name":   githubv4.String(ghc.repo),
+		"number": githubv4.Int(ghc.number),
+	}
+	if err := ghc.v4client.Query(ghc.ctx, &q, qvars); err != nil {
+		return &Body{}, errors.Wrap(err, "failed to load pull request details")
```
bluekeyes: nit: since this function returns a pointer, let's return `nil` on error instead of an empty struct

```suggestion
		return nil, errors.Wrap(err, "failed to load pull request details")
```

bluekeyes (commented):
One more minor suggestion and then this should be good to merge

bluekeyes (approved):
Thanks for implementing this and working with me through all the reviews. I'm not sure when the next release will be, but I'll do some testing and hopefully we can get a new version out in the few weeks.

PR #455, opened 2022-07-29
Add a simulate page

ashmeer7:
This allows a user to:
- Input a branch that has a policy file. The simulate page will evaluate
  that policy file on the given PR
- In put a user to simulate a fake approval for. The simulate page will
  evaluate the PR as if the user had submitted an approval
- Or both!

Abstracted away the majority of functionality into  a BaseDetails
handler. This allows consumers to submit a policy and a user to override
the defaults done by the Details handler. The Details page still exists
since it currently sends the information back to GitHub.

Here is a sample of the simulate page on a test repository: 
Screen Shot 2022-07-28 at 6 30 14 PM

### DIFF
--- a/server/assets/js/footer-link.js
+++ b/server/assets/js/footer-link.js
@@ -1,0 +1,17 @@
+(function() {
+    function addSimulateRedirect(){
+        const buttonId = "details-to-simulate"
+        function redirectToSimulate(event) {
+            const currentPath = location.href
+            let simulatePath = currentPath.replace("details", "simulate")
+            document.location = simulatePath
+        }
+        document.getElementById(buttonId).addEventListener("click", redirectToSimulate)
+    }
+
+    document.addEventListener('DOMContentLoaded', function() {
+        addSimulateRedirect()
+    }, false);
+
+
+})();
--- a/server/assets/js/prefill.js
+++ b/server/assets/js/prefill.js
@@ -1,0 +1,23 @@
+(function() {
+    function prefillSimulation(){
+        function prefillParam(param){
+            const queryString = window.location.search;
+            const urlParams = new URLSearchParams(queryString);
+            if (urlParams.has(param)) {
+                let paramValue = urlParams.get(param)
+                document.getElementById(param).value = paramValue
+            }
+        }
+
+        const BRANCH = "branch"
+        const USERNAME = "username"
+
+        prefillParam(BRANCH)
+        prefillParam(USERNAME)
+    }
+    document.addEventListener('DOMContentLoaded', function() {
+        prefillSimulation()
+    }, false);
+
+
+})();
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -114,7 +114,7 @@ func (b *Base) PostStatus(ctx context.Context, prctx pull.Context, client *githu
 	}
 
 	publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
-	detailsURL := fmt.Sprintf("%s/details/%s/%s/%d", publicURL, owner, repo, prctx.Number())
+	detailsURL := fmt.Sprintf("%s/details/%s/%s/%DetailsData", publicURL, owner, repo, prctx.Number())
 
 	contextWithBranch := fmt.Sprintf("%s: %s", b.PullOpts.StatusCheckContext, base)
 	status := &github.RepoStatus{
@@ -233,10 +233,16 @@ func (b *Base) ValidateFetchedConfig(ctx context.Context, prctx pull.Context, cl
 	return evaluator, nil
 }
 
-func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc FetchedConfig) (common.Result, error) {
-	logger := zerolog.Ctx(ctx)
-
+func (b *Base) EvaluateConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc FetchedConfig) common.Result {
 	result := evaluator.Evaluate(ctx, prctx)
+
+	return result
+}
+
+func (b *Base) PostEvaluatedResults(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc FetchedConfig, result common.Result) (common.Result, error) {
+	logger := zerolog.Ctx(ctx)
+	statusDescription := result.StatusDescription
+
 	if result.Error != nil {
 		msg := fmt.Sprintf("Error evaluating policy in %s: %s", fc.Source, fc.Path)
 		logger.Warn().Err(result.Error).Msg(msg)
@@ -244,8 +250,6 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
 		b.PostStatus(ctx, prctx, client, "error", msg)
 		return result, result.Error
 	}
-
-	statusDescription := result.StatusDescription
 
 	var statusState string
 	switch result.Status {
@@ -266,6 +270,13 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
 	b.PostStatus(ctx, prctx, client, statusState, statusDescription)
 
 	return result, nil
+}
+
+func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc FetchedConfig) (common.Result, error) {
+	res := b.EvaluateConfig(ctx, prctx, client, evaluator, fc)
+	var result, err = b.PostEvaluatedResults(ctx, prctx, client, evaluator, fc, res)
+
+	return result, err
 }
 
 func (b *Base) RequestReviewsForResult(ctx context.Context, prctx pull.Context, client *github.Client, trigger common.Trigger, result common.Result) error {
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -239,7 +239,7 @@ func (b *Base) EvaluateConfig(ctx context.Context, prctx pull.Context, client *g
 	return result
 }
 
-func (b *Base) PostEvaluatedResults(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc *FetchedConfig, result common.Result) (common.Result, error) {
+func (b *Base) PostEvaluatedResults(ctx context.Context, prctx pull.Context, client *github.Client, fc *FetchedConfig, result common.Result) (common.Result, error) {
 	logger := zerolog.Ctx(ctx)
 	statusDescription := result.StatusDescription
 
@@ -274,9 +274,8 @@ func (b *Base) PostEvaluatedResults(ctx context.Context, prctx pull.Context, cli
 
 func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc *FetchedConfig) (common.Result, error) {
 	res := b.EvaluateConfig(ctx, prctx, client, evaluator, fc)
-	var result, err = b.PostEvaluatedResults(ctx, prctx, client, evaluator, fc, res)
 
-	return result, err
+	return b.PostEvaluatedResults(ctx, prctx, client, fc, res)
 }
 
 func (b *Base) RequestReviewsForResult(ctx context.Context, prctx pull.Context, client *github.Client, trigger common.Trigger, result common.Result) error {
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -114,7 +114,7 @@ func (b *Base) PostStatus(ctx context.Context, prctx pull.Context, client *githu
 	}
 
 	publicURL := strings.TrimSuffix(b.BaseConfig.PublicURL, "/")
-	detailsURL := fmt.Sprintf("%s/details/%s/%s/%DetailsData", publicURL, owner, repo, prctx.Number())
+	detailsURL := fmt.Sprintf("%s/details/%s/%s/%d", publicURL, owner, repo, prctx.Number())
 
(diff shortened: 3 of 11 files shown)


bluekeyes:
Could you add a screenshot of what the simulate form and the button look like to the PR description?

server/assets/js/footer-link.js
```diff
@@ -0,0 +1,17 @@
+(function() {
+    function addSimulateRedirect(){
+        const buttonId = "details-to-simulate"
+        function redirectToSimulate(event) {
+            const currentPath = location.href
+            let simulatePath = currentPath.replace("details", "simulate")
+            document.location = simulatePath
+        }
+        document.getElementById(buttonId).addEventListener("click", redirectToSimulate)
```
bluekeyes, line 9: I don't think this needs to be a JS function. We should be able to use a link that we style to look like a button, where the target URL is rendered as part of the page.

server/assets/js/prefill.js
```diff
@@ -0,0 +1,23 @@
...
+                let paramValue = urlParams.get(param)
+                document.getElementById(param).value = paramValue
+            }
+        }
+
+        const BRANCH = "branch"
+        const USERNAME = "username"
+
+        prefillParam(BRANCH)
+        prefillParam(USERNAME)
```
bluekeyes, line 16: Instead of hardcoding specific fields here, it might be nice to make this a bit more generic. One idea is to add a `data-queryparam` attribute to all of the `input` tags that need to be pre-filled. Then the JS can use `querySelectorAll` to find all of them, read the value of of the attribute to get the parameter name, and then do the logic you have here to set the value to the right thing.

server/handler/base.go
```diff
@@ -233,10 +233,16 @@ func (b *Base) ValidateFetchedConfig(ctx context.Context, prctx pull.Context, cl
 	return evaluator, nil
 }
 
-func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc FetchedConfig) (common.Result, error) {
+func (b *Base) EvaluateConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc *FetchedConfig) common.Result {
+	result := evaluator.Evaluate(ctx, prctx)
+
+	return result
+}
```
bluekeyes, line 240: I'm not sure we need this function since it only calls a function on an argument with no other processing.

server/handler/base.go
```diff
@@ -233,10 +233,16 @@ func (b *Base) ValidateFetchedConfig(ctx context.Context, prctx pull.Context, cl
...
+
+func (b *Base) PostEvaluatedResults(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc *FetchedConfig, result common.Result) (common.Result, error) {
 	logger := zerolog.Ctx(ctx)
+	statusDescription := result.StatusDescription
```
bluekeyes, line 244: Any reason this moved up here instead of staying where it was?

server/handler/base.go
```diff
@@ -233,10 +233,16 @@ func (b *Base) ValidateFetchedConfig(ctx context.Context, prctx pull.Context, cl
...
 }
 
-func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc FetchedConfig) (common.Result, error) {
+func (b *Base) EvaluateConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc *FetchedConfig) common.Result {
+	result := evaluator.Evaluate(ctx, prctx)
+
+	return result
+}
+
+func (b *Base) PostEvaluatedResults(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc *FetchedConfig, result common.Result) (common.Result, error) {
```
bluekeyes: Let's remove `evaluator` as an argument here since it is not used

server/handler/base.go
```diff
@@ -268,6 +272,13 @@ func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, cl
 	return result, nil
 }
 
+func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc *FetchedConfig) (common.Result, error) {
+	res := b.EvaluateConfig(ctx, prctx, client, evaluator, fc)
+	var result, err = b.PostEvaluatedResults(ctx, prctx, client, evaluator, fc, res)
+
+	return result, err
```
bluekeyes: nit: if you're not modifying or using the results, you can return them directly:

```go
return b.PostEvaluatedResults(ctx, prctx, client, evaluator, fc, res)
```

server/handler/base_details.go
```diff
@@ -0,0 +1,228 @@
...
+	Result           *common.Result
+	PullRequest      *github.PullRequest
+	User             string
+	PolicyURL        string
+}
+
+func (h *DetailsBase) getURLParams(w http.ResponseWriter, r *http.Request) (string, string, int, error) {
+	owner := pat.Param(r, "owner")
+	repo := pat.Param(r, "repo")
+	//number :=
```
bluekeyes: Left over from testing?

server/handler/base_details.go
```diff
@@ -0,0 +1,228 @@
...
+type DetailsData struct {
+	Error            error
+	IsTemporaryError bool
+	Result           *common.Result
+	PullRequest      *github.PullRequest
+	User             string
+	PolicyURL        string
+}
+
+func (h *DetailsBase) getURLParams(w http.ResponseWriter, r *http.Request) (string, string, int, error) {
```
bluekeyes: nit: let's call this something that indicates the parameters are actually the PR details or PR locator

server/handler/base_details.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	return owner, repo, number, nil
+}
+
+func (h *DetailsBase) generatePrContext(ctx context.Context, owner string, repo string, number int) (pull.Context, error) {
+	installation, err := h.Installations.GetByOwner(ctx, owner)
+	if err != nil {
+		if _, notFound := err.(githubapp.InstallationNotFound); notFound {
+			//h.render404(w, owner, repo, number)
```
bluekeyes: Left over from testing?

server/handler/base_details.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	config := h.ConfigFetcher.ConfigForPRBranch(ctx, prCtx, client, branch)
+
+	return config, err
+}
+
+func (h *DetailsBase) generateEvaluationDetails(w http.ResponseWriter, r *http.Request, policyConfig *FetchedConfig, prCtx pull.Context) (*DetailsData, *github.Client, common.Evaluator, error) {
+	ctx := r.Context()
+	//logger := zerolog.Ctx(ctx)
```
bluekeyes: Left over from testing?

server/handler/base_details.go
```diff
@@ -0,0 +1,228 @@
...
+
+	owner := prCtx.RepositoryOwner()
+	repo := prCtx.RepositoryName()
+	number := prCtx.Number()
+
+	installation, err := h.Installations.GetByOwner(ctx, owner)
+	if err != nil {
+		if _, notFound := err.(githubapp.InstallationNotFound); notFound {
+			h.render404(w, owner, repo, number)
```
bluekeyes: Seems a bit weird to have some rendering happen in this function when it's mostly about producing a data object (and a client). I think it would be better to have the caller detect if it is correct to render the 404 page in the same place where it renders the other page.

server/handler/base_details.go
```diff
@@ -0,0 +1,228 @@
...
+		owner, repo, number,
+	)
+	http.Error(w, msg, http.StatusNotFound)
+}
+
+func (h *DetailsBase) render(w http.ResponseWriter, templateName string, data interface{}) error {
+	w.Header().Set("Content-Type", "text/html")
+	w.WriteHeader(http.StatusOK)
+
+	//fmt.Println(data)
```
bluekeyes: Left over from testing?

server/handler/fetcher.go
```diff
@@ -37,10 +36,8 @@ type ConfigFetcher struct {
 	Loader *appconfig.Loader
 }
 
-func (cf *ConfigFetcher) ConfigForPR(ctx context.Context, prctx pull.Context, client *github.Client) FetchedConfig {
-	base, _ := prctx.Branches()
-
-	c, err := cf.Loader.LoadConfig(ctx, client, prctx.RepositoryOwner(), prctx.RepositoryName(), base)
+func (cf *ConfigFetcher) ConfigForPRBranch(ctx context.Context, prctx pull.Context, client *github.Client, branch string) *FetchedConfig {
```
bluekeyes, line 39: What's the reason for returning a pointer now? It looks like the value is always non-nil.

Also, this should be probably be called `ConfigForBranch` since the "pr branch" would be the same thing as  what `ConfigForPR` does.

server/handler/simulate.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/pkg/errors"
+	"net/http"
+	"time"
+)
+
+type Simulate struct {
+	DetailsBase
+}
+
+type fakePrContext struct {
```
bluekeyes: nit: let's call this `SimulatePullContext` or something like that

server/handler/simulate.go
```diff
@@ -0,0 +1,90 @@
...
+	reviews, err := f.Context.Reviews()
+
+	for _, username := range f.additionalApprovals {
+		reviews = append(reviews, &pull.Review{CreatedAt: time.Now(), UpdatedAt: time.Now(), Author: username, State: pull.ReviewApproved})
+	}
+
+	return reviews, err
+}
+
+func new(ctx pull.Context) *fakePrContext {
```
bluekeyes: nit: calling this `new` shadows a builtin Go function; `NewSimulatePullContext` or something like that would be better

server/templates/details.html.tmpl
```diff
@@ -1,12 +1,14 @@
-{{/* templatetree:extends page.html.tmpl */}}
+{{/* templatetree:extends base-details.html.tmpl */}}
 {{define "title"}}{{.PullRequest.GetBase.GetRepo.GetFullName}}#{{.PullRequest.GetNumber}} - Details | PolicyBot{{end}}
```
bluekeyes: This is now duplicated with `base-details.html.tmpl`, as is `body-class` and some other templates. If we're going to extend a template, let's either remove these overrides or remove the definitions from the parent.

server/templates/details.html.tmpl
```diff
@@ -21,142 +23,10 @@
...
-
-{{define "result-requires-permissions-details"}}
-  <b class="font-bold text-sm">This rule requires review from users with the following permissions</b>
-  <ul class="list-disc list-outside pl-6 py-2">
-    {{range getPermissions .}}<li class="font-mono text-sm-mono">{{.}}</li>{{end}}
-  </ul>
-{{end}}
+{{define "footer"}}
+  <footer class="bg-white shadow-sm w-full z-10">
+    <button id="details-to-simulate" class="bg-blue3 hover:bg-blue5 text-white font-bold py-2 px-4 rounded focus:outline-none focus:shadow-outline" type="button">Simulate changes on this PR</button>
```
bluekeyes, line 22: As mentioned on the JS file, this should be a regular link that we style to look like a button if you want

server/templates/simulate.html.tmpl
```diff
@@ -0,0 +1,62 @@
...
+      {{.PullRequest.GetTitle}}
+    </h1>
+    <span class="text-xs text-dark-gray3 truncate max-w-full">
+      {{.User}}
+    </span>
+  </header>
+  <div class="bg-light-gray3 border-b border-gray3">
+    <div class="pl-5 pt-5">
+      <form class="w-full max-w-md" method="get">
+        <div class="md:flex md:items-center mb-6">
```
bluekeyes, line 23: As a general rule, I've tried to avoid the media query based classes (`md:` prefix) in favor of things that naturally scale and adjust as the screen size changes. I'd probably see how this works with the un-prefixed classes and then adjust from there.

bluekeyes (commented):
Sorry about the delay reviewing this. I think we'll want to make some improvements to the form UI, but I haven't had a chance to pull this down and mess around with the HTML and styles yet, so I don't have specific suggestions right now.

bluekeyes:
#476 is going to conflict with this, but it should make it much easier to evaluate a PR without posting status checks.

server/handler/base.go
ashmeer7: I went with this approach to contrast with `EvaluateFetchedConfig`. 

This one does the evaluating portion (and any processing if required) while `EvaluateFetchedConfig` evaluates and posts the results. This way, the simulate page can display what is happening without messing with the status of the PR.

server/handler/base.go
ashmeer7: It made it easier to understand what was going on IMO

server/handler/fetcher.go
ashmeer7: The pointer was so I could return `nil` values here: https://github.com/palantir/policy-bot/pull/455/files#diff-089a392be5ea1fe56b5ce5b40f9e6495be5ec02592d56c5bbd8fc097acf7f22aR109 This simplified the error handling logic.

ashmeer7:
Addressed some PR comments and fixed what I could fix without getting a whole test setup. I'll be back with more questions soon enough.

elliot-graebert-skydio:
@bluekeyes this came up when deploying Policy-Bot. Could we get this merged? Use case was simply trying to test out a new policy on a branch, and not knowing the right path to do so.

asvoboda:
A variant of this was done in https://github.com/palantir/policy-bot/pull/725. I think given the age of this PR we should close and re-open at a later time if we want to add a simulation UI

PR #459, opened 2022-08-17
fix ignore edited comments bug

devinburnette:
I incorrectly [assumed](https://github.com/palantir/policy-bot/pull/357) that github would always set updatedAt and submittedAt to be the same upon review comment creation, but apparently that isn't so and it just bit us. Same for createdAt and updatedAt for issue comments.

```
                "updatedAt": "2022-08-16T18:58:18Z",
                "submittedAt": "2022-08-16T18:58:17Z"
```
I was able to easily switch out updatedAt for lastEditedAt and it seems to work well. lastEditedAt is just always null until it is actually edited which makes it much easier to discern programmatically. probably should have went with this from the beginning, but who knew?

at least this bug, although annoying, wasn't broken in the other direction by incorrectly counting edited approvals but instead only intermittently not counting appropriate approvals when the ignore option is enabled.


bluekeyes (approved):
Nice catch, that's a pretty subtle edge case in the API.

PR #472, opened 2022-09-17
dont evaluate unless we need to

### LINKED ISSUE #470: maximum number of statuses for SHA and context
We got the following error today on a PR:
```
422 Validation Failed [{Resource:Status Field: Code:custom Message:This SHA and context has reached the maximum number of statuses.}]
```

I traced it down to what I believe to be our codecov integration which is routinely commenting and editing comments on PRs in this repo. Github's documentation shows that the maximum number of statuses that can be posted for a given SHA / context pair is 1000. So while I think this may be an outlier case, it still doesn't sound crazy to me that there would be 1000 events causing evaluation and subsequent status checks on a SHA.

I think under normal circumstances we shouldn't hit this error and this is the first PR that I've seen behave this way. But it does make me think if there are ways to prevent this somehow without needing to retain the state of the last evaluation if nothing needs to change. Like maybe we could first query the most recent status of that SHA / context and if the recent evaluation doesn't change the status or message text then we don't need to post anything?? something like that, idk, what do you think? It does add another API call for every event, but it might be better than bumping up against an imposed limit and having to push another commit and re-approve any invalidated rules, etc.

relevant docs:
https://docs.github.com/en/rest/commits/statuses#create-a-commit-status

devinburnette:
closes #470 

As mentioned in #470 there are certain situations where we could run into API limitations on the number of status checks posted to a SHA / context pair. This should cut down on some unnecessary API calls being made currently during evaluation.


server/handler/issue_comment.go
```diff
@@ -100,6 +100,10 @@ func (h *IssueComment) Handle(ctx context.Context, eventType, deliveryID string,
 		return nil
 	}
 
+	if !h.affectsApproval(event.GetComment().GetBody(), fc.Config.ApprovalRules) {
```
bluekeyes: I think this misses an edge case with edited comments. The previous `detectAndLogTampering` call exits early if the edit was by the same user that authored the original comment. Since `affectsApproval` is only looking at the current body, if a user edits their own comment to remove an approval, I think this will cause us to skip evaluating the PR (only the old body affects approval, and we don't check that.)

To catch that case, I think we need to look at the previous content for `edited` events as well, possibly assuming that a comment affects approval if the old content is missing (e.g. in #379, where a GHE bug removed this field.)

As a general rule, I want to make sure this type of optimization will over-evaluate if there are errors or uncertainty, rather than under-evaluate.

server/handler/pull_request_review.go
```diff
@@ -39,13 +40,55 @@ func (h *PullRequestReview) Handle(ctx context.Context, eventType, deliveryID st
...
 		Owner:  event.GetRepo().GetOwner().GetLogin(),
 		Repo:   event.GetRepo().GetName(),
 		Number: event.GetPullRequest().GetNumber(),
 		Value:  event.GetPullRequest(),
 	})
 }
+
+func (h *PullRequestReview) affectsApproval(reviewState string, rules []*approval.Rule) bool {
+	for _, rule := range rules {
+		if reviewState == string(rule.Options.GetMethods().GithubReviewState) {
```
bluekeyes: nit: it might be better to cast the parameter before calling this function, rather than casting the rule value

server/handler/pull_request_review.go
```diff
@@ -39,13 +40,55 @@ func (h *PullRequestReview) Handle(ctx context.Context, eventType, deliveryID st
...
+		return err
+	}
+
+	fc := h.ConfigFetcher.ConfigForPR(ctx, prctx, client)
+
+	if !h.affectsApproval(event.GetReview().GetState(), fc.Config.ApprovalRules) {
+		return nil
+	}
+
 	return h.Evaluate(ctx, installationID, common.TriggerReview, pull.Locator{
```
bluekeyes: Sorry about all the boilerplate you have to copy to get access to the config prior to calling `Evaluate`. But now that we have a `FetchedConfig`, this should probably call `ValidateFetchedConfig` and `EvaluateFetchedConfig` instead, since `Evaluate` repeats the client creation and config fetching.

bluekeyes (dismissed):
Thanks for proposing a fix! I think the comment editing edge case is the main thing to address, but I had some minor style suggests too.

server/handler/issue_comment.go
```diff
@@ -141,9 +137,18 @@ func (h *IssueComment) detectAndLogTampering(ctx context.Context, prctx pull.Con
...
+	case "edited":
+		originalBody = event.GetChanges().GetBody().GetFrom()
+
+	case "created", "deleted":
+		originalBody = event.GetComment().GetBody()
+	}
+
 	for _, rule := range rules {
-		if rule.Options.GetMethods().CommentMatches(actualComment) {
+		if rule.Options.GetMethods().CommentMatches(originalBody) {
```
devinburnette: how about this? this will look at all events but only run them through evaluation if the current or previous matches a rule pattern.

server/handler/issue_comment.go
```diff
@@ -111,15 +115,7 @@ func (h *IssueComment) Handle(ctx context.Context, eventType, deliveryID string,
...
 
-	var originalBody string
-	switch event.GetAction() {
-	case "edited":
-		originalBody = event.GetChanges().GetBody().GetFrom()
-
-	case "deleted":
-		originalBody = event.GetComment().GetBody()
-
-	default:
```
devinburnette, line 115: tried to dry this up a bit.

server/handler/issue_comment.go
bluekeyes: I think this is still missing a case: when the action is `edited` we need to test both the original body and the new body to catch events where an approval was removed or added.

server/handler/pull_request_review.go
```diff
@@ -39,13 +40,64 @@ func (h *PullRequestReview) Handle(ctx context.Context, eventType, deliveryID st
...
+
+	client, err := h.NewInstallationClient(installationID)
+	if err != nil {
+		return err
+	}
+
+	v4client, err := h.NewInstallationV4Client(installationID)
+	if err != nil {
+		return err
+	}
```
bluekeyes: In #476, I did some refactoring to reduce this boiler plate

server/handler/pull_request_review.go
```diff
@@ -39,13 +40,64 @@ func (h *PullRequestReview) Handle(ctx context.Context, eventType, deliveryID st
...
+	if !h.affectsApproval(reviewState, fc.Config.ApprovalRules) {
+		return nil
+	}
+
+	_, err = h.Base.EvaluateFetchedConfig(ctx, prctx, client, evaluator, fc)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
bluekeyes, line 86: #476 also adds a new function that we'll need to call here to make sure things like review requests and review dismissal still happen.

bluekeyes (commented):
I started on some refactoring that I think will make this a bit cleaner (and will also create merge conflicts). If you don't mind, I can take over this PR once my change is in and get it updated and merged.

server/handler/issue_comment.go
devinburnette: oh right, i see what you mean now.

bluekeyes (approved):

ashmeer7 (approved):

devinburnette:
lgtm, thanks

PR #476, opened 2022-09-26
Add EvalContext to reduce handler duplication

bluekeyes:
Previously, we defined a bunch of common functionality for evaluating PRs on the Base handler. This worked, but had a few flaws:

* When a handler needed to drop to a lower level to insert logic between steps, there was a lot of client construction boilerplate. This is not totally eliminated, but is reduced.

* As new side-effects are added, it was easy to miss calling them from places that had dropped down to the lower-level functions. For example, when adding review dismissal, we missed called it from the comment handler.

* Functions took a lot of similar arguments that could be defined on a single struct and shared.

* There was no easy way to evalute a PR without posting status check

* The Base handler was getting hard to read with all the evaluation functionality mixed in.

The EvalContext type contains all the configuration and clients needed to evaluate a single PR for a request. Most handlers call Base.Evaluate like before, but when dropping to a lower level, handlers now create and use an EvalContext. Once created, they call 3 methods in sequence adding whatever extra logic is needed in between.

### DIFF
--- a/server/handler/base.go
+++ b/server/handler/base.go
@@ -16,31 +16,17 @@ package handler
 
 import (
 	"context"
-	"fmt"
-	"math/rand"
-	"os"
-	"strings"
-	"time"
 
 	"github.com/google/go-github/v45/github"
 	"github.com/palantir/go-baseapp/baseapp"
 	"github.com/palantir/go-githubapp/githubapp"
-	"github.com/palantir/policy-bot/policy"
 	"github.com/palantir/policy-bot/policy/common"
-	"github.com/palantir/policy-bot/policy/reviewer"
 	"github.com/palantir/policy-bot/pull"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
-	"github.com/shurcooL/githubv4"
 )
 
 const (
-	DefaultPolicyPath       = ".policy.yml"
-	DefaultSharedRepository = ".github"
-	DefaultSharedPolicyPath = "policy.yml"
-
-	DefaultStatusCheckContext = "policy-bot"
-
 	LogKeyGitHubSHA = "github_sha"
 )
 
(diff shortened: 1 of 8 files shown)


server/handler/base.go
```diff
@@ -55,96 +41,11 @@ type Base struct {
 	AppName string
 }
 
-type PullEvaluationOptions struct {
```
bluekeyes, line 58: Moved to `eval_options.go`

server/handler/base.go
```diff
@@ -55,96 +41,11 @@ type Base struct {
...
-
-func (p *PullEvaluationOptions) SetValuesFromEnv(prefix string) {
-	setStringFromEnv("POLICY_PATH", prefix, &p.PolicyPath)
-	setStringFromEnv("SHARED_REPOSITORY", prefix, &p.SharedRepository)
-	setStringFromEnv("SHARED_POLICY_PATH", prefix, &p.SharedPolicyPath)
-	setStringFromEnv("STATUS_CHECK_CONTEXT", prefix, &p.StatusCheckContext)
-	p.fillDefaults()
-}
-
-func (b *Base) PostStatus(ctx context.Context, prctx pull.Context, client *github.Client, state, message string) {
```
bluekeyes, line 104: Moved to `EvalContext`

server/handler/base.go
```diff
@@ -156,360 +57,41 @@ func (b *Base) PreparePRContext(ctx context.Context, installationID int64, pr *g
...
+		Options:   b.PullOpts,
+		PublicURL: b.BaseConfig.PublicURL,
 
-	return nil
+		PullContext: prctx,
+		Config:      fetchedConfig,
+	}, nil
 }
 
-func (b *Base) ValidateFetchedConfig(ctx context.Context, prctx pull.Context, client *github.Client, fc FetchedConfig, trigger common.Trigger) (common.Evaluator, error) {
```
bluekeyes, line 203: Moved to `EvalContext#ParseConfig`

server/handler/base.go
```diff
@@ -156,360 +57,41 @@ func (b *Base) PreparePRContext(ctx context.Context, installationID int64, pr *g
...
-			Str("event_trigger", trigger.String()).
-			Str("policy_trigger", policyTrigger.String()).
-			Msg("No evaluation necessary for this trigger, skipping")
-		return nil, nil
-	}
-
-	return evaluator, nil
-}
-
-func (b *Base) EvaluateFetchedConfig(ctx context.Context, prctx pull.Context, client *github.Client, evaluator common.Evaluator, fc FetchedConfig) (common.Result, error) {
```
bluekeyes, line 247: Moved to `EvalContext#EvaluatePolicy`

server/handler/base.go
```diff
@@ -156,360 +57,41 @@ func (b *Base) PreparePRContext(ctx context.Context, installationID int64, pr *g
...
-		err := errors.Errorf("Evaluation resulted in unexpected status: %s", result.Status)
-		return result, err
-	}
-
-	b.PostStatus(ctx, prctx, client, statusState, statusDescription)
-
-	return result, nil
-}
-
-func (b *Base) RequestReviewsForResult(ctx context.Context, prctx pull.Context, client *github.Client, trigger common.Trigger, result common.Result) error {
```
bluekeyes, line 282: Moved to `eval_context_reviewers.go` along with support functions

server/handler/base.go
```diff
@@ -156,360 +57,41 @@ func (b *Base) PreparePRContext(ctx context.Context, installationID int64, pr *g
...
-	}
-
-	if review.CreatedAt.Before(time.Now().Add(-5 * time.Second)) {
-		return "was invalidated by another commit"
-	}
-
-	return "didn't include a valid comment pattern"
-}
-
-func (b *Base) dismissStaleReviewsForResult(ctx context.Context, prctx pull.Context, v4client *githubv4.Client, result common.Result) error {
```
bluekeyes, line 448: Moved to `eval_context_dismissal.go` along with support functions

server/handler/issue_comment.go
```diff
@@ -81,34 +74,34 @@ func (h *IssueComment) Handle(ctx context.Context, eventType, deliveryID string,
...
 	}
 
-	result, err := h.Base.EvaluateFetchedConfig(ctx, prctx, client, evaluator, fc)
+	result, err := evalCtx.EvaluatePolicy(ctx, evaluator)
 	if err != nil {
 		return err
 	}
 
-	return h.Base.RequestReviewsForResult(ctx, prctx, client, common.TriggerComment, result)
+	evalCtx.RunPostEvaluateActions(ctx, result, common.TriggerComment)
```
bluekeyes, line 100: This fixes a bug from when review dismissal was added: we weren't dismissing reviews in response to comment events evaluations

devinburnette: i thought that was intentional, because we're only looping through the [reviews](https://github.com/palantir/policy-bot/pull/476/files#diff-9a1aa4ae85323f98b833a9551817bf9388a01bfef333c49b28c08927f5512f2fR49), but that's a good point, maybe we should also do the same for comments too. I think we might need to change up the dismissal logic a bit to support it.

bluekeyes: I think you're right that this doesn't change behavior (no reviews should be dismissed in response to comment events), but that's also true of other events where dismissal was evaluated by virtue of being part of `Evaluate`.

My main goal is to make all the "post-review" actions consistent so it's harder to miss in future cases where it does matter. If we want to optimize review dismissal further, I think it should check if the `trigger` relates to reviews or commits, similar to what reviewer assignment does.

server/handler/eval_context.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
+	"github.com/shurcooL/githubv4"
+)
+
+// EvalContext contains common fields and methods used to evaluate policy
+// requests. Handlers construct an EvalContext once they decide to handle a
+// request or event, then call the appropriate methods for each stage of
+// evaluation. Handlers with no special requirements can simply call Evaluate.
+type EvalContext struct {
```
asvoboda, line 35: Thanks for doing this refactor. I think the EvalContext struct is a nice reorganization of the previous code.

asvoboda (approved):

PR #485, opened 2022-10-13
Fix review dismissals bug

### LINKED ISSUE #480: bug: review approval dismissals are misleading when predicates are not satisfied
For example, the use case we saw had the status predicate:
```
if:
  has_successful_status:
    - some_status_check
```

Policy bot said that a review is dismissed because it didn't have a valid pattern. I think we should probably just allow review dismissals for results with a skipped evaluation assuming that evaluation will be triggered again when and if the predicate is ever satisfied.

In other words, if the status is skipped, then evaluation can't happen so there's no chance of accidentally allowing an approval that should have been dismissed, until a final evaluation happens.

I'm not sure it's the best approach or exactly the best way to implement, but it's what I got. What do you think?

/cc @bluekeyes

devinburnette:
Closes https://github.com/palantir/policy-bot/issues/480

In https://github.com/palantir/policy-bot/pull/463 we introduced a feature to dismiss stale reviews, while iterating on it some additional logic slipped in to also dismiss reviews that didn't match review patterns if people were using the `github_review_comment_patterns` field in their policies. This worked well until a predicate was included in a policy causing certain reviews to be skipped from evaluation and therefore not added to the list of approved candidates. This manifested in buggy/broken behavior with a misleading dismissal reason until the predicate was satisfied.

After deliberating in the issue above, I think we agreed that the dismissal of non-matching patterns is just a little too heavy handed and doesn't really fit well with dismissing "stale" reviews. In other words, if we aren't also dismissing all other reviews that can never satisfy a rule under any condition than the feature seems broken. So this PR removes dismissal of non-matching patterns, but keeps dismissal for both edited and invalidated reviews, classifying those as "stale" if people have the `ignore_edited_comments` or `invalidate_on_push` fields set to true in their policies. In addition, this PR also highlights which patterns will satisfy which rules on the details page in hopes to clear up confusion for folks using approval comments.

Screen Shot 2022-10-13 at 6 21 30 PM

### DIFF
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -33,7 +33,7 @@ type Rule struct {
 	Description string               `yaml:"description"`
 	Predicates  predicate.Predicates `yaml:"if"`
 	Options     Options              `yaml:"options"`
-	Requires    Requires             `yaml:"requires"`
+	Requires    common.Requires      `yaml:"requires"`
 }
 
 type Options struct {
@@ -76,12 +76,6 @@ func (opts *Options) GetMethods() *common.Methods {
 	return methods
 }
 
-type Requires struct {
-	Count int `yaml:"count"`
-
-	common.Actors `yaml:",inline"`
-}
-
 func (r *Rule) Trigger() common.Trigger {
 	t := common.TriggerCommit
 
@@ -111,7 +105,7 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
 	res.Name = r.Name
 	res.Description = r.Description
 	res.Status = common.StatusSkipped
-	res.Requires = common.Requires{Count: r.Requires.Count, Actors: r.Requires.Actors}
+	res.Requires = r.Requires
 	res.Methods = r.Options.GetMethods()
 
 	var predicateResults []*common.PredicateResult
@@ -174,10 +168,10 @@ func (r *Rule) getReviewRequestRule() *common.ReviewRequestRule {
 	}
 
 	return &common.ReviewRequestRule{
-		Users:         r.Requires.Users,
-		Teams:         r.Requires.Teams,
-		Organizations: r.Requires.Organizations,
-		Permissions:   r.Requires.GetPermissions(),
+		Users:         r.Requires.Actors.Users,
+		Teams:         r.Requires.Actors.Teams,
+		Organizations: r.Requires.Actors.Organizations,
+		Permissions:   r.Requires.Actors.GetPermissions(),
 		RequiredCount: r.Requires.Count,
 		Mode:          mode,
 	}
@@ -228,7 +222,7 @@ func (r *Rule) IsApproved(ctx context.Context, prctx pull.Context, candidates []
 			continue
 		}
 
-		isApprover, err := r.Requires.IsActor(ctx, prctx, c.User)
+		isApprover, err := r.Requires.Actors.IsActor(ctx, prctx, c.User)
 		if err != nil {
 			return false, "", errors.Wrap(err, "failed to check candidate status")
 		}
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -111,7 +111,8 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
 	res.Name = r.Name
 	res.Description = r.Description
 	res.Status = common.StatusSkipped
-	res.Requires = r.Requires.Actors
+	res.Requires = common.Requires{Count: r.Requires.Count, Actors: r.Requires.Actors}
+	res.Methods = r.Options.GetMethods()
 
 	var predicateResults []*common.PredicateResult
 
--- a/policy/common/methods.go
+++ b/policy/common/methods.go
@@ -179,3 +179,7 @@ func (m *Methods) BodyMatches(prBody string) bool {
 	}
 	return false
 }
+
+func (m *Methods) IsEmpty() bool {
+	return m == nil || (len(m.Comments) == 0 && len(m.CommentPatterns) == 0 && len(m.GithubReviewCommentPatterns) == 0 && len(m.BodyPatterns) == 0)
+}
--- a/policy/common/methods.go
+++ b/policy/common/methods.go
@@ -179,7 +179,3 @@ func (m *Methods) BodyMatches(prBody string) bool {
 	}
 	return false
 }
-
-func (m *Methods) IsEmpty() bool {
-	return m == nil || (len(m.Comments) == 0 && len(m.CommentPatterns) == 0 && len(m.GithubReviewCommentPatterns) == 0 && len(m.BodyPatterns) == 0)
-}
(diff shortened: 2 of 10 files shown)


policy/approval/approve.go
```diff
@@ -111,7 +111,8 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
 	res.Name = r.Name
 	res.Description = r.Description
 	res.Status = common.StatusSkipped
-	res.Requires = r.Requires.Actors
+	res.Requires = common.Requires{Count: r.Requires.Count, Actors: r.Requires.Actors}
```
devinburnette: this felt a little funny, but i kept getting cyclical dependency issues when just trying to use the existing `approval.Requires`

server/handler/frontend.go
```diff
@@ -60,14 +60,20 @@ func LoadTemplates(c *FilesConfig, basePath string, githubURL string) (templatet
...
 		},
-		"hasRequires": func(requires common.Actors) bool {
-			return len(requires.Users) > 0 || len(requires.Teams) > 0 || len(requires.Organizations) > 0
+		"hasActors": func(requires *common.Requires) bool {
+			return len(requires.Actors.Users) > 0 || len(requires.Actors.Teams) > 0 || len(requires.Actors.Organizations) > 0
 		},
-		"getRequires": func(results *common.Result) map[string][]Membership {
-			return getRequires(results, strings.TrimSuffix(githubURL, "/"))
+		"hasPatterns": func(results *common.Result) bool {
+			return !results.Methods.IsEmpty()
```
devinburnette: there's one thing I think I'm missing here and that's what if a github review approval (without comments) would satisfy the policy just fine? I think i can make it only show the Patterns if `github_review: true` and `github_review_comment_patterns` is not set.

devinburnette: but then it's like the logic is slipping into the frontend a little bit.

policy/common/methods.go
```diff
@@ -179,3 +179,7 @@ func (m *Methods) BodyMatches(prBody string) bool {
 	}
 	return false
 }
+
+func (m *Methods) IsEmpty() bool {
+	return m == nil || (len(m.Comments) == 0 && len(m.CommentPatterns) == 0 && len(m.GithubReviewCommentPatterns) == 0 && len(m.BodyPatterns) == 0)
```
devinburnette: oh i can just put it here, and maybe call this something different.

policy/approval/approve.go
bluekeyes: Moving types to `common` is normal as they end up used in `common.Result`, so I think this is fine.

policy/common/result.go
```diff
@@ -59,14 +59,20 @@ type ReviewRequestRule struct {
 	Mode RequestMode
 }
 
+type Requires struct {
```
bluekeyes, line 62: Instead of duplicating `Requires` in this package, I think it might be better to move `approval.Requires` to this package and then fix the references in the `approval` packages. That should reduce confusion between the two types in the future since they're otherwise the same.

policy/common/methods.go
bluekeyes: I think this also needs to check that `m.GithubReview` is non-nil to account for all fields.

server/handler/frontend.go
```diff
@@ -60,14 +60,20 @@ func LoadTemplates(c *FilesConfig, basePath string, githubURL string) (templatet
 
 			return r
 		},
-		"hasRequires": func(requires common.Actors) bool {
-			return len(requires.Users) > 0 || len(requires.Teams) > 0 || len(requires.Organizations) > 0
+		"hasActors": func(requires *common.Requires) bool {
```
bluekeyes, line 63: Thanks for renaming some of these existing functions, on review the old naming scheme was kind of confusing.

server/handler/frontend.go
bluekeyes: What if instead of thinking of this as patterns-only, we show all methods? I think that removes the need for `hasPatterns` (every rule will have methods) and then we can change `getPatterns` to `getMethods` and update the template accordingly (see later comments)

server/templates/details.html.tmpl
```diff
@@ -65,14 +65,19 @@
...
           <div class="pt-2">
-          {{template "result-requires-details" .}}
+          {{template "result-actors-details" .}}
           </div>
         {{end}}
-        {{if (hasRequiresPermissions .Requires)}}
+        {{if (hasPatterns .)}}
           <div class="pt-2">
-          {{template "result-requires-permissions-details" .}}
+          {{template "result-patterns-details" .}}
```
bluekeyes: I think it would make sense to reorder these templates so it reads something like:

> This rule needs N approvals from people in these groups or people with these permissions. Approvals can use these patterns.

As currently structured, it kind of implies that the sections are independent so that the pattern section doesn't also require the approvers have certain team membership/permissions.

server/templates/details.html.tmpl
```diff
@@ -142,10 +147,22 @@
   </ul>
 {{end}}
 
-{{define "result-requires-details"}}
-  <b class="font-bold text-sm">This rule requires review from</b>
+{{define "result-patterns-details"}}
+  <b class="font-bold text-sm">This rule requires {{.Requires.Count}} review(s) using any of these patterns</b>
```
bluekeyes: Good call adding the count to all of these - I'm not sure why we didn't add it earlier. I think it would be better to only show the count once before all the other approval details. With the count repeated, people could interpret it as if they need N reviews for each section instead of N reviews total for the rule.

bluekeyes (commented):
Thanks for fixing this and expanding the details view! I had some suggestions on the UI part, but the rest of the change looks good.

server/handler/frontend.go
devinburnette: sgtm

policy/common/methods.go
devinburnette: yea i realized that, but with the new approach, i think we don't need it anymore.

server/templates/details.html.tmpl
devinburnette: yea that's fair, i was trying to play with the wording so that wouldn't be the case, but I think this is a better idea.

server/handler/frontend.go
```diff
@@ -91,24 +94,47 @@ func Static(prefix string, c *FilesConfig) http.Handler {
...
+		patternInfo["Body Patterns"] = append(patternInfo["Body Patterns"], bodyPattern.String())
+	}
+	if *result.Methods.GithubReview {
+		if len(result.Methods.GithubReviewCommentPatterns) > 0 {
+			for _, githubReviewCommentPattern := range result.Methods.GithubReviewCommentPatterns {
+				patternInfo["Github Review Comment Patterns + Github Review Approval"] = append(patternInfo["Github Review Comment Patterns + Github Review Approval"], githubReviewCommentPattern.String())
+			}
+		} else {
+			patternInfo["Github Review State"] = append(patternInfo["Github Review State"], string(result.Methods.GithubReviewState))
+		}
```
devinburnette, line 115: this is the only part where i feel like the logic slipped into the view a little, but i thought the result would be more helpful for humans to read.

bluekeyes (approved):
Thanks for addressing the comments and sorry about the delayed review, this looks good to merge

PR #500, opened 2022-11-18
Retry commit loading when pushed dates are missing

bluekeyes:
Back in 1d06dfa771, I removed retries for pushed date loading, preferring to always use a fallback API. This seemed to work better in most cases, but we're still seeing persistent but temporary failures loading pushed dates for certain merge commits.

Add retries for the whole process again, covering both the initial pull request API and the fallback commit APIs, in the hope that this gives enough time for GitHub to update the internal data.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -798,13 +798,47 @@ func (ghc *GitHubContext) loadPagedData() error {
 func (ghc *GitHubContext) loadCommits() ([]*Commit, error) {
 	log := zerolog.Ctx(ghc.ctx)
 
-	rawCommits, err := ghc.loadRawCommits()
-	if err != nil {
-		return nil, err
+	// GitHub's API is eventually consistent for commit pushed dates. We can
+	// sometimes get around this by using alternate APIs (see notes in
+	// loadCommitsOnce), but for some merge commits, that still fails. The only
+	// fix I know of in this case is to wait and try again.
+	//
+	// Retry with exponential backoff until it works or we hit the max total
+	// latency to avoid users thinking the bot got stuck or dropped an event.
+	const baseDelay = 1 * time.Second
+	const maxLatency = 20 * time.Second
+
+	start := time.Now()
+	for delay := baseDelay; true; delay *= 2 {
+		head, commits, err := ghc.loadCommitsOnce()
+		if err != nil {
+			return nil, err
+		}
+		if head.PushedAt != nil {
+			return commits, nil
+		}
+
+		if time.Since(start)+delay >= maxLatency {
+			break
+		}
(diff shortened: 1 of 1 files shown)


pull/github.go
```diff
@@ -798,13 +798,47 @@ func (ghc *GitHubContext) loadPagedData() error {
...
 
+	// GitHub's API is eventually consistent for commit pushed dates. We can
+	// sometimes get around this by using alternate APIs (see notes in
+	// loadCommitsOnce), but for some merge commits, that still fails. The only
+	// fix I know of in this case is to wait and try again.
+	//
+	// Retry with exponential backoff until it works or we hit the max total
+	// latency to avoid users thinking the bot got stuck or dropped an event.
+	const baseDelay = 1 * time.Second
+	const maxLatency = 20 * time.Second
```
bluekeyes, line 809: I think doing a time-bounded retry is a bit easier to think about than a retry count, but open to switching this to a more traditional N attempts where we manually compute the expected max latency.

pull/github.go
```diff
@@ -798,13 +798,47 @@ func (ghc *GitHubContext) loadPagedData() error {
...
+		log.Debug().Dur("delay", delay).Str("sha", ghc.pr.HeadRefOID).Msg("Head commit is missing pushed data, sleeping and trying again")
+		time.Sleep(delay)
+	}
+
+	return nil, &TemporaryError{
+		fmt.Sprintf("Commit %.10s does not have a pushed date. This is usually caused by delays in the GitHub API", ghc.pr.HeadRefOID),
+	}
+}
+
+func (ghc *GitHubContext) loadCommitsOnce() (head *Commit, commits []*Commit, err error) {
```
ashmeer7, line 833: I'll assume that this is only called one function above so we can have it do whatever.

pull/github.go
ashmeer7: Is there a way we can add loggging to make it clear that our worker thread is in a long sleep? If for whatever reason we've failed a few times in a row, its hard to know why its not doing anything from the logs.

pull/github.go
```diff
@@ -798,13 +798,47 @@ func (ghc *GitHubContext) loadPagedData() error {
...
+		}
+		if head.PushedAt != nil {
+			return commits, nil
+		}
+
+		if time.Since(start)+delay >= maxLatency {
+			break
+		}
+		log.Debug().Dur("delay", delay).Str("sha", ghc.pr.HeadRefOID).Msg("Head commit is missing pushed data, sleeping and trying again")
+		time.Sleep(delay)
```
ashmeer7, line 825: This message was going to be my suggestion for the above, so I think I'm happy.

ashmeer7 (approved):

PR #503, opened 2022-11-30
feat: add reject property to changed_files config

villers:
I add the possibility to refuse a rule if a file match a given regular expression 

```
  changed_files:
      - "secrets/.*"
    reject:
      - "secrets/prod.yaml"
```

### DIFF
--- a/policy/predicate/file.go
+++ b/policy/predicate/file.go
@@ -28,12 +28,13 @@ import (
 type ChangedFiles struct {
 	Paths       []common.Regexp `yaml:"paths"`
 	IgnorePaths []common.Regexp `yaml:"ignore"`
+	RejectPaths []common.Regexp `yaml:"reject"`
 }
 
 var _ Predicate = &ChangedFiles{}
 
 func (pred *ChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
-	var paths, ignorePaths []string
+	var paths, ignorePaths, rejectPaths []string
 
 	for _, path := range pred.Paths {
 		paths = append(paths, path.String())
@@ -43,12 +44,17 @@ func (pred *ChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (*co
 		ignorePaths = append(ignorePaths, ignorePath.String())
 	}
 
+	for _, rejectPath := range pred.RejectPaths {
+		rejectPaths = append(rejectPaths, rejectPath.String())
+	}
+
 	predicateResult := common.PredicateResult{
 		ValuePhrase:     "changed files",
 		ConditionPhrase: "match",
 		ConditionsMap: map[string][]string{
-			"path patterns":  paths,
-			"while ignoring": ignorePaths,
+			"path patterns":   paths,
+			"while ignoring":  ignorePaths,
+			"while rejecting": rejectPaths,
 		},
 	}
 
@@ -58,6 +64,15 @@ func (pred *ChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (*co
 	}
 
 	changedFiles := []string{}
+
+	for _, f := range files {
+		if anyMatches(pred.RejectPaths, f.Filename) {
+			predicateResult.Values = []string{f.Filename}
+			predicateResult.Description = f.Filename + " was rejected"
+			predicateResult.Satisfied = false
+			return &predicateResult, nil
+		}
+	}
 
 	for _, f := range files {
 
--- a/README.md
+++ b/README.md
@@ -159,8 +159,8 @@ if:
   # "changed_files" is satisfied if any file in the pull request matches any
   # regular expression in the "paths" list. If the "ignore" list is present,
   # files in the pull request matching these regular expressions are ignored
-  # by this rule.
-  #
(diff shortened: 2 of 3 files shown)


silly-eliah (approved):

afarvacque (approved):

bluekeyes:
Thanks for proposing this feature, @villers. Could you provide some more details about what you are trying to achieve and an example policy that uses this new property? I'm having some trouble understanding how this is different from the existing `ignore` property.

villers:
@bluekeyes 

The purpose of the `reject` property is to allow policy-bot to reject a rule. If one of configured patterns matches, the policy-bot will considered this rule as not statisfied.

Example:

We want to reject a rule if one of the files of `.cloud/terraform` folder is modified. With the new property, the configuration will look like:

```yaml
policy:
  approval:
    - or:
      - Terraform modifications are approved by TeamA
      - Auto approved if the .cloud/terraform folder is not modified

approval_rules:
  - name: Terraform modifications are approved by TeamA
    if:
      changed_files:
        paths:
          - '^\.cloud/terraform/.*'
    requires:
      count: 1
      teams:
        - Organiazion/TeamA
    options:
      request_review:
        enabled: true
        mode: teams
  - name: Auto approved if the .cloud/terraform folder is not modified
    if:
      changed_files:
        paths:
          - '.*'
        reject: # the reject property that allow to do the rejection feature
          - '^\.cloud/terraform/.*'
```

From what I understand of `ignore`, policy-bot will not take into account the file if the pattern match the file path. But it didn't guarantee that the rule will not be statified if one file match the `changed_files.paths` patterns.

policy/predicate/file.go
```diff
@@ -59,6 +65,15 @@ func (pred *ChangedFiles) Evaluate(ctx context.Context, prctx pull.Context) (*co
...
 	changedFiles := []string{}
 
+	for _, f := range files {
+		if anyMatches(pred.RejectPaths, f.Filename) {
+			predicateResult.Values = []string{f.Filename}
+			predicateResult.Description = f.Filename + " was changed"
+			predicateResult.Satisfied = false
+			return &predicateResult, nil
+		}
+	}
```
burahimu: **question (non-blocking):** why do you have to do another loop here ? 🤔 You cannot do it in the initial `range` ?

villers: a this line the range is break with return https://github.com/palantir/policy-bot/pull/503/files#diff-9ab6ff99f800968214d5d7039c7db7b84b4ca6f859b85ee288fef0f37c8a7e79R89

villers: if a file is found before one of the rejected files, the rule will be validated when I don't want it

burahimu: Yes you're right 💯

burahimu (approved):

bluekeyes:
Thanks for the explanation. For the policy you shared, we usually implement this behavior by using `and` to combine rules instead of `or`. For example:

```yaml
policy:
  approval:
    - Terraform modifications are approved by TeamA
    - All other changes are auto-approved

approval_rules:
  - name: Terraform modifications are approved by TeamA
    if:
      changed_files:
        paths:
          - '^\.cloud/terraform/.*'
    requires:
      count: 1
      teams:
        - Organiazion/TeamA
    options:
      request_review:
        enabled: true
        mode: teams
  - name: All other changes are auto-approved
    requires:
      count: 0
```

When a PR modifies any file in `.cloud/terraform/` both rules apply, so the overall policy is only satisfied when Team A approves. If the PR only modifies non-Terraform files, then the first rule is `skipped` and does not contribute to the policy at all, while the second rule passes and the overall policy is successful.

By extending the pattern (and combing rules that use `changed_files` with rules that use `only_changed_files`), we've been able to implement all the workflows like this that we've found so far, although I acknowledge that it can be confusing to discover. Do you think this will work for you?

In the meantime, your implementation looks good to me, so let me think a bit more about if the `reject` property is something we'd like to support.

villers:
> Thanks for the explanation. For the policy you shared, we usually implement this behavior by using `and` to combine rules instead of `or`. For example:
> 
> ```yaml
> policy:
>   approval:
>     - Terraform modifications are approved by TeamA
>     - All other changes are auto-approved
> 
> approval_rules:
>   - name: Terraform modifications are approved by TeamA
>     if:
>       changed_files:
>         paths:
>           - '^\.cloud/terraform/.*'
>     requires:
>       count: 1
>       teams:
>         - Organiazion/TeamA
>     options:
>       request_review:
>         enabled: true
>         mode: teams
>   - name: All other changes are auto-approved
>     requires:
>       count: 0
> ```
> 
> When a PR modifies any file in `.cloud/terraform/` both rules apply, so the overall policy is only satisfied when Team A approves. If the PR only modifies non-Terraform files, then the first rule is `skipped` and does not contribute to the policy at all, while the second rule passes and the overall policy is successful.
> 
> By extending the pattern (and combing rules that use `changed_files` with rules that use `only_changed_files`), we've been able to implement all the workflows like this that we've found so far, although I acknowledge that it can be confusing to discover. Do you think this will work for you?
> 
> In the meantime, your implementation looks good to me, so let me think a bit more about if the `reject` property is something we'd like to support.

Your solution works :)

PR #504, opened 2022-11-30
SSH signatures support

fpoussin:
Hi,

SSH signed commits are currently ignored because the field type was unknown.

This PR adds basic support for it.

### DIFF
--- a/policy/predicate/signature_test.go
+++ b/policy/predicate/signature_test.go
@@ -30,7 +30,7 @@ func TestHasValidSignatures(t *testing.T) {
 
 	testCases := []SignatureTestCase{
 		{
-			"ValidSignature",
+			"ValidGpgSignature",
 			&pulltest.Context{
 				AuthorValue: "mhaypenny",
 				CommitsValue: []*pull.Commit{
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @fpoussin! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes (approved):
Thanks! This looks good for the `has_valid_signatures` and `has_valid_signatures_by` predicates. As a follow-up, we should probably add a `key_fingerprints` property to the `has_valid_signatures_by_keys` predicate, but I think this is fine to merge for partial SSH support.

bluekeyes:
Hey @fpoussin, it turns out this is incompatible with GitHub Enterprise 3.6 and earlier, which do not support SSH signature. Running the GraphQL query fails with the error: `No such type SshSignature, so it can't be a fragment condition`

I'm going to see if there's a way to enable the feature only when the target GitHub instance supports it, but I might have to temporarily revert this PR until GitHub Enterprise 3.7 is more widely deployed (probably early in 2023.)

PR #507, opened 2022-12-01
Revert "SSH signatures support"

bluekeyes:
Reverts palantir/policy-bot#504. Unfortunately, this is incompatible with GitHub Enterprise 3.6 and earlier. I have some ideas of how to support this via feature detection, but want to make sure we can continue to take other fixes and features while still running on GHE 3.6. I'll file an issue to track this.


ashmeer7:
👍

PR #536, opened 2023-02-16
Hide skipped rules by default in the UI

### LINKED ISSUE #395: Allow “Hide Skipped Rules” to default to enabled
The setting to hide skipped rules is basically always what I want, and especially for complex rule sets it’s really crucial to parsing who needs to approve.

Can we add an option so that it’s possible to switch it to enabled-by-default?

kmark:
When browsing the UI, you’re usually only concerned with the policy rules that actually applied to the specific pull request. This PR changes the “Hide skipped rules” toggle to be on by default.

Fixes #395.

### DIFF
--- a/server/assets/js/filter.js
+++ b/server/assets/js/filter.js
@@ -1,9 +1,8 @@
 (function() {
   const hidden = [];
 
-  function toggleFilter(evt) {
-    const enabled = evt.target.checked;
-    if (enabled) {
+  function toggleFilter(target) {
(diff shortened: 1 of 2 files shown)


ericanderson (approved):

bluekeyes (approved):
Thanks! In my test with a small-ish policy (10 rules), I didn't notice any flash of skipped rules on load and this otherwise worked as expected.

PR #554, opened 2023-03-25
feat: add merge group handler in support of github merge queues

### LINKED ISSUE #548: [Feature Request] Merge queue compatibility
Hi there!

Our team is investigating turning on [merge queue](https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/) for our repository and in doing so, ran into an issue (expectedly) with PolicyBot. Merge queue does not operate on PRs but rather allows you to add PRs to a queue once all their required checks are green, after which it creates specialized branches with the naming structure `gh-readonly-queue/main/pr-number-<hash>`. PolicyBot never "runs" because that branch isn't associated with a pull request. Even if it _did_ try to run, there'd be no way of adding approvals.

So, a general question: Is there any existing functionality that might make this possible? If not, some ideas in no particular order:

1. Since merge queue branches have a special [trigger event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group) within Github Actions workflows (`merge_group`), we could create a workflow that runs Policy Bot whenever a `merge_group` checks requested event occurs. If the binary allows for us to invoke commands which add approvals, we could automatically provide necessary approvals because we know the code changes have already been approved (as that's a requirement before adding a PR to the queue).
2. If the above won't work, could we somehow add functionality to PolicyBot that would allow it to consider a branch approved to merge based on the approval status of an open PR? In other words, if PR 9243 has been approved and added to the queue and a branch `gh-readonly-queue/main/9243*` has been created, we could consider that branch "approved to merge". This may be considered safe because you may not push to either branch (the original one or the queue one) while the PR is in the queue. You must remove the PR from the queue, push changes (therefore revoking PolicyBot approvals), and then get approvals and add to the queue again.

_Note: These ideas are based on the expectation that we can provide a status check update for a specific commit on a branch (rather than a PR) like Github Actions do!_

I'll add more ideas in comments as they come to me, but was hoping to get the discussion started in the meantime. Thank you!

devinburnette:
Closes https://github.com/palantir/policy-bot/issues/548

Adds support for github merge queues by unconditionally posting a successful status check on the merge group branch when policy bot sees a `merge_group` github webhook event.

Worked together with @btrautmann to get this done. Let us know if something like this is what you had in mind as a solution for this.

### DIFF
--- a/server/handler/merge_group.go
+++ b/server/handler/merge_group.go
@@ -87,7 +87,7 @@ func (h *MergeGroup) Handle(ctx context.Context, eventType, devlieryID string, p
 	}
 
 	if err := PostStatus(ctx, client, owner, repository, headSHA, status); err != nil {
-		logger.Err(errors.WithStack(err)).Msg("Failed to post repo status")
+		logger.Err(errors.WithStack(err)).Msg("Failed to post status check for merge group")
 	}
 
 	return nil
--- a/server/handler/merge_group.go
+++ b/server/handler/merge_group.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Palantir Technologies, Inc.
+// Copyright 2023 Palantir Technologies, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/server/handler/merge_group.go
+++ b/server/handler/merge_group.go
@@ -18,9 +18,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/google/go-github/v50/github"
 	"github.com/palantir/go-githubapp/githubapp"
+	"github.com/palantir/policy-bot/policy/common"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
 )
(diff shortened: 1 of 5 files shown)


server/handler/merge_group.go
```diff
@@ -0,0 +1,94 @@
...
+	fetchedConfig := h.ConfigFetcher.ConfigForRepositoryBranch(ctx, client, owner, repository, baseBranch)
+
+	evalCtx := &EvalContext{
+		Client:    client,
+		Options:   h.PullOpts,
+		PublicURL: h.BaseConfig.PublicURL,
+		Config:    fetchedConfig,
+	}
+
+	_, err = evalCtx.ParseConfig(ctx, common.TriggerStatic)
```
devinburnette: might have to make a new method for validating the policy since ParseConfig doesn't return an evaluator for static trigger, so as written this will post a status even if there's no policy file, and i don't think that's what we want

bluekeyes: We might not need to parse the configuration at all. I think it would be enough to only check `fetchedConfig` to determine if a policy file exists, since the main goal is to not post status checks in repositories that don't enable Policy Bot. If a PR is added to the merge queue, presumably the policy existed and was valid at the time of merge, so I think it could actually be confusing if the policy was changed to be invalid while the PR was in the queue and it led to a failed status.

devinburnette: good point, sounds good to me, will make that change

server/handler/merge_group.go
```diff
@@ -0,0 +1,86 @@
...
+	logger := zerolog.Ctx(ctx)
+	installationID := githubapp.GetInstallationIDFromEvent(&event)
+	client, err := h.NewInstallationClient(installationID)
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
+		return errors.Wrap(err, "failed to create evaluation context")
+	}
```
bluekeyes: Looks like there's a double error check here

bluekeyes (commented):
Thanks for adding this, the implementation looks good to me as a starting point. I imagine we might need to make adjustments as people gain experience with merge queues, but this should be a safe way to unblock testing. One minor code issue and then this should be good to merge.

bluekeyes (approved):
Thanks!

PR #556, opened 2023-04-03
build: for linux/arm64 platform

ab77:
Hi folks, hopefully it's as simple as that...

### DIFF
--- a/godel/config/dist-plugin.yml
+++ b/godel/config/dist-plugin.yml
@@ -13,6 +13,8 @@ products:
       os-archs:
       - os: linux
         arch: amd64
+      - os: linux
+        arch: arm64
     dist:
       output-dir: build
       disters:


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @ab77! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes:
Thanks! I think this will do the right thing for the distribution, but I'll have to check if it will also build a `linux/arm64` container, or if we need some extra configuration for that.

bluekeyes (approved):
After following up with internal teams, it looks like our build tool doesn't have full support for multi-architecture containers yet - I filed #558 to track fixing this. Until then, you may need to build your own container using the distribution (which should have both platform binaries once this merges.)

PR #577, opened 2023-05-13
Improve review dismissal behavior

bluekeyes:
Existing dismissal logic compares all reviews on the PR to the set of reviews that might satisfy any rule. This does not work well when policies combine rules that invalidate on push with rules that do not.

In this situation, reviews are not dismissed when new commits are pushed because they might satisfy the rules that do not invalidate. Instead, a user's review is dismissed when they submit a second review. This is because we only consider the most recent review from a user, which means their previous review doesn't count for any rules and is eligible to be dismissed.

We chose this approach to try to dismiss reviews that did not match required comment patterns, but ended up removing that before merging the feature.

This commit switched back to the original approach, which is to collect specific reviews to dismiss, rather than trying to infer what is safe to dismiss. This should be more reliable and is easier to reason about.

### Testing / Edge Cases

I've tested a few different scenarios with my local dev instance and I think it works correctly. There are probably a few edge cases that will still show up (the space of all policies and review orders is large), but I think this is strictly more correct than the current behavior.

Two quirks I observed so far:

1. If a single approval satisfies rules A and B where A is invalidated by pushes and B is not, pushing a new commit will not dismiss the review (because it still satisfies B). This is the intended behavior, but it could confuse people. I think in practice conflicting rules like this are behind predicates.
2. If you're in the situation in (1), the PR is reapproved, and then you push a change that skips B, only the most recent approval is dismissed. There's still an older approval hanging around (it doesn't affect the Policy Bot status) because of the [deduplicateCandidates](https://github.com/palantir/policy-bot/blob/9aee3c7274e46af6463738ba5e8bd815365da257/policy/common/methods.go#L134) function, which only keeps the most recent candidate from a user. I think this is probably fine for now - changing it would mean auditing the approval logic to make sure it accounts for duplicates instead.

### Future Extensions

I don't remember if Policy Bot has permission to edit comments already, but we could probably make this work for comments as well. It could edit the comment to add a line at the bottom explaining that the comment is now invalid.

### DIFF
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -132,21 +132,22 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
 	}
 	res.PredicateResults = predicateResults
 
-	allowedCandidates, err := r.FilteredCandidates(ctx, prctx)
+	candidates, dismissals, err := r.FilteredCandidates(ctx, prctx)
 	if err != nil {
 		res.Error = errors.Wrap(err, "failed to filter candidates")
 		return
 	}
 
-	approved, msg, err := r.IsApproved(ctx, prctx, allowedCandidates)
+	approved, approvers, err := r.IsApproved(ctx, prctx, candidates)
 	if err != nil {
 		res.Error = errors.Wrap(err, "failed to compute approval status")
 		return
 	}
 
-	res.AllowedCandidates = allowedCandidates
+	res.Approvers = approvers
+	res.Dismissals = dismissals
+	res.StatusDescription = r.statusDescription(approved, approvers, candidates)
 
-	res.StatusDescription = msg
 	if approved {
 		res.Status = common.StatusApproved
 	} else {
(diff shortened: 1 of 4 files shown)


asvoboda (approved):
LGTM. Stepping back, I wonder if we should consider enabling/disabling this feature at the server level. If its too difficult/confusing to infer why or what is happening, is it valuable?

bluekeyes:
>  I wonder if we should consider enabling/disabling this feature at the server level. If its too difficult/confusing to infer why or what is happening, is it valuable?

I don't think a server-level option is valuable right now. I'm pretty confident that with this version, it will always be clear why Policy Bot dismissed a review (or if it's not, that's a bug we can fix.) The confusing cases I've found are situations where a user thinks a review should be dismissed, but Policy Bot does not dismiss it. I believe these will be rare, and usually indicate that your policy is doing something weird (e.g. for changes to a single file, one group has to re-approve on push, but another group does not.)

PR #600, opened 2023-07-07
feat: Add support for env

DepickereSven:


server/handler/eval_options.go
```diff
@@ -72,6 +72,7 @@ func (p *PullEvaluationOptions) SetValuesFromEnv(prefix string) {
 	setStringFromEnv("SHARED_REPOSITORY", prefix, &p.SharedRepository)
 	setStringFromEnv("SHARED_POLICY_PATH", prefix, &p.SharedPolicyPath)
 	setStringFromEnv("STATUS_CHECK_CONTEXT", prefix, &p.StatusCheckContext)
+	setStringFromEnv("DO_NOT_LOAD_COMMIT_PUSHED_DATE", prefix, &p.StatusCheckContext)
```
asvoboda: This is a bool, and needs to be assigned to `&p.DoNotLoadCommitPushedDate`

DepickereSven: It's fixed, I forgot that valid remark!

asvoboda:
Thanks for adding this. I'll include it in my PR.

PR #612, opened 2023-07-24
Add global cache for pushed times

bluekeyes:
This might be a premature optimization, but after testing, I'm worried that listing statuses on busy repositories will end up being too expensive, even with the HTTP caching. Since these values are safe to cache forever, add a simple in-memory LRU cache in each server instance to minimize API calls.

Also add a test for the PushedAt context function that verifies the caching behavior.


asvoboda (approved):

pull/github.go
```diff
@@ -328,19 +338,40 @@ func (ghc *GitHubContext) Commits() ([]*Commit, error) {
...
 	if err != nil {
 		return time.Time{}, err
 	}
 
-	if ghc.pushedAt == nil {
-		ghc.pushedAt = make(map[string]time.Time)
-	}
 	ghc.pushedAt[sha] = pushedAt
+	if gc := ghc.globalCache; gc != nil {
+		gc.SetPushedAt(repoID, sha, pushedAt)
```
bluekeyes: I need to fix this so that it doesn't cache the zero time, since that indicates "no pushed date" and is not globally cacheable.

Note that we might end up with the same commit having two pushed times as a result: the first evaluation will use the evaluation time and then future evaluations will use the time of the first status. I'd like to cache the evaluation time instead, but still trying to figure out how to do that in a way that makes sense, since it interacts with the commit filtering implemented at a different level.

bluekeyes:
Note for myself on how to fix the caching when there are no statuses:

1. Remove search and commit sorting logic from the approval evaluator
2. Change `PushedAt` to always return a valid time
3. In `PushedAt`, if the commit does not have any statuses, list all commits and then work towards the head until we find a status. If we hit the head commit with no status, use the evaluation timestamp.
4. Record the discovered time as the push time for all SHAs evaluated in (3)

This scanning will only happen when there are ignored commits (otherwise, we're always starting with the head commit), so should be pretty rare and the global cache will also help reduce the expense.

This will also fix a bug when evaluating PRs that have a batch push ending in an ignored commit followed directly by another push containing only ignored commits.

asvoboda (approved):

PR #625, opened 2023-08-15
Add anchors to relevant example regexps

### LINKED ISSUE #624: Misleading documentation about file path regular expressions
The documentation makes it look like path regexps implicitly anchor at the start and end, but I've confirmed for myself that this isn't the case. It'd probably be better to just put "^" and "$" on each of the paths/ignores in the documentation, to avoid confusion for others. For example, `"^server/views/.*\\.tmpl$"` rather than `"server/views/.*\\.tmpl"`.

See screenshot from README.md for context.

image

bluekeyes:
All policy-bot regular expressions are un-anchored, but the examples implied they might be. Add explicit anchors in all of the cases where it might be unclear.

Closes #624.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ approval_rules:
     if:
       only_changed_files:
         paths:
-          - "staging/.*"
+          - "^staging/.*$"
     requires:
       count: 0
 ```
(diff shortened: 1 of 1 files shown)


asvoboda (approved):

srabraham:
Thank you!

PR #636, opened 2023-09-26
Implemented RequestedCount

### LINKED ISSUE #634: Add feature to use request more reviewers than required count in case of random-users
We would like to invite more users to review a pull request than the required count. We can user the 'all-users' but this might be a bit much in to big teams.

This is an example rule

```yaml
- name: Code change
  if:
    has_successful_status:
      - "build"
  requires:
    count: 2
    teams:
    - "<org>/engineering"
  options:
    invalidate_on_push: true
    request_review:
      enabled: true
      mode: random-users
    methods:
      github_review: true
```

The engineering team is quite big so we would like to invite a few more users but still require a review count of 2.

 It would be nice if in case of `random-users` we can add a setting to the `request_review` section to set a percentage to invite.  

```yaml
    request_review:
      enabled: true
      mode: random-users
      requested_reviewers: 4
``` 

This will then add 4 reviewers to the PR if the rule is matched. 

I would sugegst the following behaviour
* `request_reviewers` is undefined by default
* if `request_reviewers` is set, this this the number of reviewers (no count+requested reviewers)

If you like we can create a PR to implement this.

AtzeDeVries:
Made a small change in sugested implementation, instead of `requested_reviewers` i used `requested_count` which more clearly points to a number. 

Fixes: https://github.com/palantir/policy-bot/issues/634

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,12 +357,11 @@ options:
     # Defaults to 'random-users'.
     mode: all-users|random-users|teams
     
-    # requested_count modifies how many users are requested to review the pr. If 
-    # requested_count is not set or set to 0 it will use requires.count
-    # Settings this is usefull when you want to request more reviewers than the 
-    # required count if you are using mode: random-users. 
-    # Defaults to '0'
-    requested_count: 0
+    # count sets the number of users requested to review the pull request when
+    # using the `random-users` mode. If count is not set or set to 0, request the
+    # number of users set by requires.count. Setting this is useful when you want
+    # to request more reviewers than the required count. Defaults to 0.
+    count: 0
 
   # "methods" defines how users may express approval.
   methods:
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @AtzeDeVries! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes (approved):
Thanks for adding this! After thinking about it a bit, I renamed the configuration field to just "count" - it already appears within the `request_reviews` block, so also putting "request" in the name felt redundant.

bluekeyes:
If you want to test this out before the next release, you can use the snapshot Docker tag:

```
docker pull palantirtechnologies/policy-bot:snapshot
```

AtzeDeVries:
ok thanks! Any hint when next release will be happening?

PR #656, opened 2023-12-08
Add option to expand reviewers in details UI

### LINKED ISSUE #49: FR: UI should expand which teams (and users) can actually approve something
## What happened

A common workflow:

1. I make a PR, all the CI goes green, but the overall light is still stuck at yellow
2. I click through to the policy bot UI to see what I need to do to get my thing merged
    screen shot 2019-01-24 at 12 42 12
    Unfortunately the UI doesn't tell me who I need to ping.

3. I go back to the main file browser, hit `t` to browse files, then read the `policy.yml` file to see which teams are considered maintainers
4. I go to the org homepage
5. Click teams
6. Search for my relevant team, click it
7. Click the members tab

Finally I can go message one of those people.

## What should have happened

I'd like this to require less steps - ideally, I'd like the policy-bot UI (and reported commit status summary) to give me a bit more information about who I need to go talk to.

bluekeyes:
If the `options.expand_required_reviewers` server option is set, when a user expands a rule in the details UI, policy-bot will expand the teams, organizations, and permission levels that can approve that rule in to a list of user names. This can make it easier to figure out who needs to approve your PR.

This feature is disabled by default because it can expose otherwise private information about membership. Only enable it in trusted environments.

This uses the HTMX library to handle loading user lists in the frontend on demand, instead of precomputing all of the information every time someone views the details of a PR. For now, I've committed a minified copy of the library, but we'll probably want to install it with `yarn` in the future.

Here's a quick screen capture that shows both the feature and the error behavior:

policy-bot-reviewers

Since this is purely information, if there are any problems loading information, we warn the user that results may be incomplete but otherwise don't fail.

I'd also like to revisit the layout of users, since a single list will be pretty long if there are a lot of users. I think columns would work, but I was wasting too much time trying to make a column layout that didn't fall apart with really long usernames. I'll revisit this either before this PR merges or in a follow-up.

Closes #49.


server/handler/details.go
```diff
@@ -39,49 +42,111 @@ type Details struct {
...
+	if state == nil {
+		return nil
+	}
+
+	ctx := state.Ctx
+	evalCtx := state.EvalContext
 
-	owner := pat.Param(r, "owner")
-	repo := pat.Param(r, "repo")
+	var data struct {
```
bluekeyes, line 65: This diff is weird, but everything after this line in this function is the same as before, except for some variable names. I extracted all the stuff that came before this in to helper functions for reuse.

server/handler/eval_options.go
```diff
@@ -73,6 +80,8 @@ func (p *PullEvaluationOptions) SetValuesFromEnv(prefix string) {
 	setStringFromEnv("SHARED_REPOSITORY", prefix, &p.SharedRepository)
 	setStringFromEnv("SHARED_POLICY_PATH", prefix, &p.SharedPolicyPath)
 	setStringFromEnv("STATUS_CHECK_CONTEXT", prefix, &p.StatusCheckContext)
+	setBoolFromEnv("EXPAND_REQUIRED_REVIEWERS", prefix, &p.ExpandRequiredReviewers)
+	setBoolFromEnv("POST_INSECURE_STATUS_CHECKS", prefix, &p.PostInsecureStatusChecks)
```
jamestoyer, line 84: I assume that this is an existing setting that's missing from the env vars?

jamestoyer (approved):

server/handler/eval_options.go
bluekeyes: Yeah, I figured I would add it for consistency now that there's a `setBoolFromEnv` function

PR #657, opened 2023-12-08
Update colors to match Blueprint v4/v5

bluekeyes:
In Blueprint v4, the colors were modernized and contrast was improved between the 1 and 5 values of each color. Because we use a limited set of colors, this should be a pretty subtle change.


derekjobst (approved):
Nice

PR #658, opened 2023-12-09
Simplify "hide skipped rules" implementation

bluekeyes:
It's always bothered me that the "hide skipped rules" toggle had to remove elements from the DOM to hide them so that the `:last-child` selector would work. This was fine when the toggle was off by default, but became a problem once we switched it to default on: skipped rules appear and then suddenly disappear once the DOM loads and the script runs.

The new version does all of the hiding with CSS rules, so the JS toggle only flips an attribute value instead adding or removing nodes. While this needed a bit of server support in the template, it should make the loading experience better and the toggle faster with large policies. And because the tree is otherwise static, there's no disadvantage to attaching extra information for styling in the template.

This is going to conflict with #656, but I'll fix whichever one merges second.


server/assets/css/main.css
```diff
@@ -30,12 +30,12 @@ details summary > * {
   @apply pl-4;
 }
 
-.tree > li {
+.tree .node {
```
bluekeyes, line 33: Had to remove the child combinator to make the top-level skipped / not skipped attribute rules work, which then required switching to a class to not mess up other list items by accident.

jamestoyer (dismissed):

asvoboda (approved):

PR #664, opened 2023-12-14
Update all actions to the latest versions

bluekeyes:
We should probably configure Dependabot to do this, but here's a one-time bulk update to get started.


asvoboda (approved):
Agreed, let's get the bots working

PR #665, opened 2023-12-15
Add dependabot for actions, golang, and npm

asvoboda:


bluekeyes (approved):
Hopefully this won't conflict too much with the Excavator check for Go dependencies

asvoboda:
https://github.com/palantir/go-githubapp/blob/develop/.github/dependabot.yml doesn't seem to cause much trouble

Issue #709, opened 2024-02-08
If no rule matches can policy-bot not set a failed status on the PR?


RoryDoherty:
I'm trying to use policy-bot along with bulldozer to automate the merging of certain PRs which could be opened by bots or scripts
In bulldozer I have set the "policy-bot: master" as a required status check

I only want policy-bot to add a commit status to PRs where it has verified that it does match one of the rules and passes
If it fails the check or doesn't match the rule then I want policy-bot to ignore it and not add a commit status
![image](https://github.com/palantir/policy-bot/assets/92513817/3e19e9e2-4c76-4e53-a5f4-a3d55e3df4cf)

Is this possible? 
I assumed if I had no disapproval set in the .policy.yml that it would not create failed status checks

bluekeyes:
The way we usually solve this problem (automatically merging PRs from bots or scripts) is a bit different:

1. Policy Bot is always a required status check in GitHub
   1. PRs from bots that modify the right files are automatically approved
   2. PRs from humans require review (you could also auto-approve these if you wanted)
2. Bulldozer will always wait for required status checks (as configured in GitHub) to pass
3. When bots open PRs, they add the `merge when ready` label to the PR. This label is the trigger that enables Bulldozer to merge the PR when all the checks are passing. You could also have another app or a GitHub Action add labels like this.

In other words, while Policy Bot is a precondition to merge, it is not the trigger that activates Bulldozer. This has worked pretty well for us and I think it's useful to be able to say that a check is always required and should always be present on a PR.

----

That said, you are not the first person to ask for Policy Bot to skip PRs where the policy does not match any rules. We added the explicit failure as a safety measure to help prevent policies that accidentally approved or ignore changes that they shouldn't have. We developed Policy Bot primarily as a compliance tool, so these safety measures are important.

I think we could add a server-level flag to control this behavior so people who are using Policy Bot for convenience rather than compliance could disable this and skip posting a status when no rules match. Unlike the initial implementation in #711, I think we want to only skip posting a status when there are no matching rules. The pending status is useful to indicate that you probably have a condition wrong in the policy (something should have been approved, but it was not) and the failure/error status is useful if something goes wrong during evaluation.

Making this a server-level flag is important to us because for our workflows, we don't want individual projects to be able to opt-in to this feature. We'd leave it disabled (which should also be the default), but you and other users could enable it when you deploy Policy Bot.

Issue #387, opened 2022-02-17
[Question] - Add Review Approval Feature


shresthaujjwal:
I would like to add a new feature where the bot would be able to add a review approval in the PR if all the policy checks is passed. Would you be able open for this contribution ?

derekjobst:
Hi @shresthaujjwal,

Could you describe the workflow(s) you are hoping to enable with this potential feature? In what way would a review approval enable functionality beyond what can be done with a required status check?

Thanks!

shresthaujjwal:
@derekjobst basically we have branch protection enabled in all our repository which requires atleast one approval in order for us to be able to auto-merge. For auto-merge, we would be using `bulldozer`. So we plan on installing policy bot and bulldozer bot to all the repository. Add this feature to policy bot would help bulldozer bot to automerge the pr. That is why i wanted to contribute and add this functionality to policy so that if all policy checks passes it would also approve the pr. 

Let me know if this helps.

shresthaujjwal:
@derekjobst any questions or concerns with this ?

asvoboda:
Generally we suggest making policy-bot the blocking status check (along with other test statuses as required for your use case). Once a user reviews, policy-bot posts a status rather than a "Github Approval". Bulldozer can then auto-merge without issue.

derekjobst:
Specifically, you can enforce the "one approval" requirement with a polcy-bot rule rather than the native GitHub branch protection.

shresthaujjwal:
@derekjobst @asvoboda our use case is little different. Not all prs qualify for automerge. Only few prs that qualify for automerge if all criteria fits. For other prs that don't qualify they need peer review and approval that is our SOD policy (Segregation of Duties). But since I can only have 1 rule in branch protection for that branch the prs that qualify for automerge also need approval that is why i am looking into adding this feature in policy bot. Hope that clarifies my concern

shresthaujjwal:
@derekjobst @asvoboda any question or concern with this ? Also can you share the policy configuration example for `you can enforce the "one approval" requirement with a polcy-bot`

asvoboda:
You can write complex logical AND OR statements in policy bot, combining rules based around authorship, files modified, and other properties of the pull request. Take a look at the examples in 

https://github.com/palantir/policy-bot/blob/develop/config/policy-examples/complicated.yml

You can change the required value to 0 to create a scenario where policy bot immediately approves with human intervention.

shresthaujjwal:
[...]

PR #711, opened 2024-02-08
Add Option to not post failure status to github when there is no successful policy

RoryDoherty:
This should address the following issue:

https://github.com/palantir/policy-bot/issues/709

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,11 @@ approval_rules:
           - "^staging/.*$"
     requires:
       count: 0
+
+# Settings which affect how policy-bot acts
+settings:
+  # Set the below to true if you only want policy-bot statuses on PRs where the policy has evaluated to true
+  only_post_success_status: false
 ```
 
 #### Notes on YAML Syntax
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -37,6 +37,11 @@ type RemoteConfig struct {
 type Config struct {
 	Policy        Policy           `yaml:"policy"`
 	ApprovalRules []*approval.Rule `yaml:"approval_rules"`
+	Settings      Settings         `yaml:"settings"`
+}
+
+type Settings struct {
+	OnlyPostSuccessStatus bool `yaml:"only_post_success_status"`
 }
 
 type Policy struct {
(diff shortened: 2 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @RoryDoherty! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

RoryDoherty:
I also think the settings in config could be used to implement this: https://github.com/palantir/policy-bot/issues/387
Could have another option called approve_pr_on_success_status

bluekeyes:
Thanks for your contribution! Could you please recreate this as two PRs, one that adds the equality operator and one for the status changes? The equality operator seems straightforward and should be easy to merge, while the status behavior is something I'd like to think about more. While the code change is minimal, it adds some new concepts to Policy Bot that deserve consideration.

RoryDoherty:
@bluekeyes thanks for the feedback, I've updated this PR so that it is just the option to not set failure status checks and opened this PR https://github.com/palantir/policy-bot/pull/712 with the changes for adding the equals operator for file modifications

RoryDoherty:
@bluekeyes would you be able to take a look at this PR anytime soon?
I'm eager to also implement https://github.com/palantir/policy-bot/issues/387 as these missing features are blocking us from currently rolling this out
I don't want to start on 387 however until I know your happy with the config settings being in the `.policy.yml` file

bluekeyes:
Sorry about the delay getting back to this. As I mentioned on the linked issue, I think this feature is something we can add with two changes:

1. Only skip posting a status when there are no matching rules. If there are matching rules, pending and failure status should still be posted.
2. Make the option to configure this a feature of the server configuration rather than the policy. We use Policy Bot on thousands of repositories and we want to make sure the behavior is consistent, where individual projects cannot opt-in to this behavior. A server-level flag would allow people who don't care about this consistency to get the skipping behavior.

RoryDoherty:
Unfortunately the changes you've suggested would make this feature unusable for us
1. If a matching policy fails I just want it to be ignored completely, I wouldn't want to post the failure status as it unnecessarily adds failed checks to a commit which would make people think something was wrong with the commit 
2. Would it be possible to have this as a server level flag to allow the setting in the policy file, This way you can have repos mix and match but only if the server has the option enabled? This would save from having to deploy two different instances to get the same feature set

PR #712, opened 2024-02-09
Add Equals operator for file modifications checks

### LINKED ISSUE #623: Allow '=' as comparison operator
The `modified_lines` condition allows only `<` or `>`.
I'd like to ask for an exact `=` option too.

The use case is that I want to allow different requirements for changes made by our automation, which should modify exactly two lines in a specific file.

There is no way to specify that accurately enough with `<` and `>`.

Thanks.

RoryDoherty:
Closes #623

The `<!-- omit in toc -->` changes in the Readme are for VSCode as it has a feature where it auto updates the TOC on save and this keeps the headings the same as what you already had in the Readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# policy-bot
+# policy-bot <!-- omit in toc -->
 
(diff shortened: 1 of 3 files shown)


bluekeyes (approved):
Thanks for adding this and including tests!

PR #725, opened 2024-03-07
Adding Simulation API

atatkin:
This pr adds a new `POST` route `api/simulate/:owner/:repo/:number`. This route requires a GitHub bearer token with access to read the the underlying pull request.

When called, policy-bot will re-evaluate the policy of the referenced pr and return a `simulated` result back to the caller. The result does NOT get written back to the pull request as a `STATUS`.

A request body can be included specifying a number of options which can modify how the simulation runs. These should be relatively easy to extend with additional options in the future.

Today this supports:
- Simulating new comments.
- Simulating new reviews.
- Simulating ignoring existing reviews and comments.
- Simulating a change of the `base` branch.

The route can also be used without any options to trigger policy-bot to re-evaluate the pull request as is.

 The returned simulated result does not currently contain all of the fields of the base `result` but it should be easy to extend as needed.


pull/simulated/context_test.go
```diff
@@ -0,0 +1,237 @@
...
+func reviewAuthors(reviews []*pull.Review) []string {
+	var authors []string
+	for _, c := range reviews {
+		authors = append(authors, c.Author)
+	}
+
+	return authors
+}
+
+type testPullContext struct {
```
bluekeyes: The `pulltest` package provides an existing test implementation of `pull.Context` that you can use

pull/simulated/options.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"time"
+
+	"github.com/palantir/policy-bot/pull"
+)
+
+// Options should contain optional data that can be used to modify the results of the methods on the simulated Context.
+type Options struct {
+	Ignore             string
```
bluekeyes: I think we should clarify and expand this field. Instead of combing both comments and reviews in a single field, I think having `IgnoreComments` and `IgnoreReviews` would be better. In the future, I could see adding an option to ignore commits or files and a generic `Ignore` field would be ambiguous.

I think these should also accept multiple entries so you could several users at once. It might even make sense to use the [`common.Actors`](https://github.com/palantir/policy-bot/blob/develop/policy/common/actor.go#L28) type here, which would allow ignoring items by organization, team, and permissions too.

If we do use the types from `common`, I think we should move this `simulated` package to be a subpackage of the `policy` package, instead of a subpackage of `pull`. It might make sense to do that anyway, since this is more of a decorator than a full implementation.

pull/simulated/options.go
```diff
@@ -0,0 +1,73 @@
...
+	AddApprovalComment string
+	AddApprovalReview  string
```
bluekeyes: More details below, but I think we'll need to take more information than just the identity of the author. Like with ignore rules, we should also accept multiple comments or reviews so that you can simulate policies with more complicated approval conditions.

A full simulation would also accept permission and membership details for these users, so that they don't have to actually exist in GitHub, but I think using real users is fine to start with.

pull/simulated/options.go
```diff
@@ -0,0 +1,73 @@
...
+		filteredReviews = append(filteredReviews, review)
+	}
+
+	return filteredReviews
+}
+
+func (o *Options) addApprovalComment(comments []*pull.Comment) []*pull.Comment {
+	return append(comments, &pull.Comment{
+		CreatedAt:    time.Now(),
+		LastEditedAt: time.Now(),
```
bluekeyes: During evaluation, Policy Bot may list comments or reviews multiple times, relying on the `pull.Context` implementation to cache them after the first request. To avoid returning varying data, I think we'll want to use a fixed time here, either provided by the user, or set internally when the `simulated.Context` or the `Options` objects are created.

Taking the value from the user would allow things like testing `invalidate_on_push` by simulating comments that happened in the past.

pull/simulated/options.go
```diff
@@ -0,0 +1,73 @@
...
+		Author:       o.AddApprovalComment,
+		Body:         ":+1:",
```
bluekeyes: Individual rules in the policy can chose what strings count for approval, so we'll want to accept the comment content from the user instead of always using `:+1:`.

pull/simulated/options.go
```diff
@@ -0,0 +1,73 @@
...
+	})
+}
+
+func (o *Options) addApprovalReview(reviews []*pull.Review) []*pull.Review {
+	return append(reviews, &pull.Review{
+		CreatedAt:    time.Now(),
+		LastEditedAt: time.Now(),
+		Author:       o.AddApprovalReview,
+		State:        pull.ReviewApproved,
```
bluekeyes: Like with comments, we'll want to use a fix date. Reviews can optionally require certain text as well to count as an approval, so accepting a `Body` value might be nice.

server/handler/simulate.go
```diff
@@ -0,0 +1,124 @@
...
+		PublicURL: h.BaseConfig.PublicURL,
+
+		PullContext: simulatedPRCtx,
+		Config:      fetchedConfig,
+	}, nil
+}
+
+func getSimulatedOptions(r *http.Request) simulated.Options {
+	var options simulated.Options
+	if r.URL.Query().Has(ignoreParam) {
```
bluekeyes: I think we're going to quickly run in to limitations trying to accept the simulation parameters as query parameters. Instead, I think we should make this a `POST` request and accept the simulation payload as a JSON body.

server/handler/simulate.go
```diff
@@ -0,0 +1,124 @@
...
+
+	simulatedPRCtx := simulated.NewContext(prctx, options)
+
+	baseBranch, _ := simulatedPRCtx.Branches()
+	owner := simulatedPRCtx.RepositoryOwner()
+	repository := simulatedPRCtx.RepositoryName()
+
+	fetchedConfig := h.ConfigFetcher.ConfigForRepositoryBranch(ctx, client, owner, repository, baseBranch)
+
+	return &EvalContext{
```
bluekeyes: I think we should avoid using `EvalContext` here and instead return the config and the simulation context directly. The `EvalContext` type is really designed to simplify posting statuses and doing other actions as part of evaluation that we don't need here and I don't want to accidentally start calling one of those functions.

server/handler/simulate_status.go
```diff
@@ -0,0 +1,103 @@
...
+	Simulate
+}
+
+type SimulateStatusResponse struct {
+	Status      string `json:"status"`
+	Description string `json:"description"`
+}
+
+type APIError struct {
+	Status int    `json:"status"`
```
bluekeyes: I think returning a `status` field with a different type based on whether the request was successful or an error will be confusing. I'd omit the status code here (it will already by in the HTTP response) and only include the `error` field (at which point, you could probably inline the whole thin in `writeAPIError`).

server/handler/simulate_status.go
```diff
@@ -0,0 +1,103 @@
...
+		logger.Error().Msg("failed to parse pull request parameters from request")
+		writeAPIError(w, http.StatusBadRequest)
+		return
+	}
+
+	installation, err := h.Installations.GetByOwner(ctx, owner)
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to get installation for org")
+		writeAPIError(w, http.StatusInternalServerError)
+		return
```
bluekeyes: For any situation where you are returning a 500 error, you could instead return the error directly (using `hatpear.Try` to wrap the handler) and rely on the [default baseapp error handler](https://github.com/palantir/go-baseapp/blob/develop/baseapp/error.go#L62-L65) to produce an appropriate response.

server/middleware/repo_auth.go
```diff
@@ -0,0 +1,96 @@
...
+// param and a valid GitHub token with access to the provided repo. Supports both personal and
+// server to server tokens.
+func RepoAuth(validator TokenValidator) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return hatpear.TryFunc(func(w http.ResponseWriter, r *http.Request) error {
+			ctx := r.Context()
+
+			token := getBearerToken(r)
+			if token == "" {
+				return errors.New("missing token")
```
bluekeyes: All of these will turn in to generic "500 Internal Server Error" responses, which is not very helpful for authentication issues. I think we'll want to either write a response directly from this middleware or define a custom error that implements the [`StatusCode()`](https://github.com/palantir/go-baseapp/blob/develop/baseapp/error.go#L29-L31) method to take advantage of baseapp's ability to change response codes.

server/middleware/repo_auth.go
```diff
@@ -0,0 +1,96 @@
...
+
+	client, err := v.ClientCreator.NewTokenClient(token)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to create token client")
+	}
+
+	// validating that the client can read the repo works for both oauth and server to server tokens
+	repo, _, err := client.Repositories.Get(r.Context(), owner, repoName)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to get repo")
```
bluekeyes: We should handle 404 errors specially here since GitHub will return a 404 both if the repository does not exist and if the token doesn't have permissions.

server/middleware/repo_auth.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return repo.GetOwner().GetLogin() == owner && repo.GetName() == repoName, nil
```
bluekeyes: If the repository was transferred or renamed, the token can have access but the URL parameters may not match the API values. I think we can return `true` here without checking anything else?

server/middleware/repo_auth.go
```diff
@@ -0,0 +1,96 @@
...
+		return false, errors.New("no repo provided")
+	}
+
+	client, err := v.ClientCreator.NewTokenClient(token)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to create token client")
+	}
+
+	// validating that the client can read the repo works for both oauth and server to server tokens
+	repo, _, err := client.Repositories.Get(r.Context(), owner, repoName)
```
bluekeyes: For application tokens, it's possible that an app will have permission to get repository metadata, but not have permission to read pull requests. Would it make sense to try reading the specific pull request instead, and if so, should that happen in the handler (where we have to read it anyway) instead of in middleware?

server/server.go
```diff
@@ -249,6 +250,16 @@ func New(c *Config) (*Server, error) {
 	}))
 	mux.Handle(pat.New("/details/*"), details)
 
+	simulateHandler := handler.Simulate{Base: basePolicyHandler}
+	simulateAPI := goji.SubMux()
+	simulateAPI.Use(middleware.RepoAuth(&middleware.GitHubTokenValidator{ClientCreator: cc}))
+	simulateAPI.Handle(pat.Get("/status"), &handler.SimulateStatus{
```
bluekeyes: As mentioned in the handler, I think this should be a POST request. I also don't think the `/status` suffix is necessary - to me, it is implied by the rest of the API.

server/server.go
```diff
@@ -249,6 +250,16 @@ func New(c *Config) (*Server, error) {
...
+		Simulate: simulateHandler,
+	})
+
+	// all routes on the simulateAPI need to include the 'owner' and 'repo' params
+	mux.Handle(pat.New("/api/simulate/:owner/:repo/:number/*"), simulateAPI)
```
bluekeyes: Can we move this up to L229 so all of the of the API routes are together?

bluekeyes (commented):
Thanks for adding this and including simulation features beyond what we need for the initial use case. Overall, I think this implementation makes sense, but I had some suggestions for how the simulation options and the handler should work.

server/middleware/repo_auth.go
atatkin: Good idea, I updated the auth middleware to validate that the token can read the pr instead of the repo.

We could perform this lookup in the handler since we have to do that anyway, but it would require passing the bearer token through to the handler which probably means storing it in `context`.

I'm open to that if it's better to store the token in `context` vs having to lookup the same pull request twice. Though I do believe the github client is using caching, so perhaps that second lookup is not so bad?

bluekeyes: If we checked authentication in the handler, we'd probably skip the middleware completely. The handler would take care of everything using some helper function instead. The PR request should be cached, but it might still be nice to only do it once.

atatkin: oh I see so skip the middleware entirely and perform the token validation in the handler itself, that would work too since currently the simulate feature is a single route and I don't really see it growing to multiple routes anytime soon.

atatkin: Moved the token auth into the `simulate` handler and removed the now unneeded middleware.

policy/simulated/context_test.go
```diff
@@ -0,0 +1,403 @@
...
+				CollaboratorsValue: test.Collaborators,
+			},
+			options: test.Options,
+		}
+
+		comments, err := context.Comments()
+		if test.ExpectedError {
+			assert.Error(t, err, message)
+		} else {
+			assert.NoError(t, err, test, message)
```
bluekeyes: I think you only want to pass `message` in this assertion? Passing both `test` and `message` will try to use `test` as a format string.

This should also probably be `require.NoError` so that we don't run the `assert.Equals` when we already know the comments won't be correct.

policy/simulated/context_test.go
```diff
@@ -0,0 +1,403 @@
...
+			assert.Equal(t, getCommentAuthors(test.ExpectedComments), getCommentAuthors(comments), message)
```
bluekeyes: If we're going to extract the authors from `ExpectedComments` and otherwise not use any other fields, it might be easier to define the expectation as a `[]string` in the test struct.

policy/simulated/context_test.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+}
+
+func getCommentAuthors(comments []*pull.Comment) []string {
+	var authors []string
+	for _, c := range comments {
+		authors = append(authors, c.Author)
+	}
+
+	return authors
```
bluekeyes, line 164: I recommend sorting this so that the test doesn't depend on the authors appearing in the right order.

policy/simulated/options.go
```diff
@@ -0,0 +1,235 @@
...
+
+	return base, head
+}
+
+type Actors struct {
+	Users         []string `json:"users"`
+	Teams         []string `json:"teams"`
+	Organizations []string `json:"organizations"`
+	Permissions   []string `json:"permissions"`
+}
```
bluekeyes: I think we can add `json` tags to the existing `common.Actors` struct to avoid have to redefine it here and have a conversion function.

policy/simulated/options.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	var filteredComments []*pull.Comment
+	actors, err := o.IgnoreComments.toCommonActors()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, comment := range comments {
+		isActor, err := actors.IsActor(ctx, prCtx, comment.Author)
```
bluekeyes: Looking at `IsActor`, it looks like it doesn't actually use the first `ctx` argument. If you want to simplify this and avoid passing a `context.Context` around, I think we could remove that argument.

policy/simulated/context.go
```diff
@@ -0,0 +1,65 @@
...
+	return &Context{Context: pullContext, options: options}
+}
+
+func (c *Context) Comments() ([]*pull.Comment, error) {
+	comments, err := c.Context.Comments()
+	if err != nil {
+		return nil, err
+	}
+
+	comments, err = c.options.filterIgnoredComments(c.ctx, c.Context, comments)
```
bluekeyes: It feels odd to me to have most of the logic for the context implementation actually defined on the options type. Did you consider having `Options` only define functions for setting defaults and then defining the manipulation logic either on the `Context` type or as private functions in this file?

policy/simulated/options_test.go
```diff
@@ -0,0 +1,91 @@
...
+		"add_comments":[
+			{"author":"iignore", "body":":+1:"}
+		],
+		"add_reviews":[
+			{"author":"iignore", "body":":+1:", "state":"approved"}
+		],
+		"base_branch":"test-base"
+	}`
+
+	req, err := http.NewRequest(http.MethodPost, "", bytes.NewBuffer([]byte(body)))
```
bluekeyes: You can use [`httptest.NewRequest`](https://pkg.go.dev/net/http/httptest#NewRequest) to avoid checking this error.

policy/simulated/options_test.go
```diff
@@ -0,0 +1,91 @@
...
+	assert.NoError(t, err)
+
+	opt, err := NewOptionsFromRequest(req)
+	assert.NoError(t, err)
```
bluekeyes: `require.NoError` to avoid continuing the test when there is an error

server/apierror/error.go
```diff
@@ -0,0 +1,31 @@
...
+	"strings"
+
+	"github.com/palantir/go-baseapp/baseapp"
+)
+
+type ErrorResponse struct {
+	Error string `json:"error"`
+}
+
+func WriteAPIError(w http.ResponseWriter, code int, message ...string) error {
```
bluekeyes: Could this be a function in the `handler` package or does it benefit from adding a new package?

If we do keep it in a new package, let's change the name to `Write` so that usage in other packages doesn't duplicate the "API error" part.

server/handler/simulate.go
```diff
@@ -0,0 +1,182 @@
...
+
+	token := getToken(r)
+	if token == "" {
+		return apierror.WriteAPIError(w, http.StatusUnauthorized, "missing token")
+	}
+
+	client, err := h.NewTokenClient(token)
+	if err != nil {
+		logger.Error().Msg("failed to create token client")
+		return apierror.WriteAPIError(w, http.StatusUnauthorized, "token invalid")
```
bluekeyes: This should be an internal error, I'm pretty sure `NewTokenClient` doesn't actually validate anything about the token and this will only happen if something fails creating the client instance (like when parsing the URL).

server/handler/simulate.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	owner, repo, number, ok := parsePullParams(r)
+	if !ok {
+		logger.Error().Msg("failed to parse pull request parameters from request")
```
bluekeyes: If we're going to return a descriptive error to the client, there's usually no need to log it as well (unless there's extra internal information that is useful.) In particular, 4xx errors are usually noise in server logs because they indicate that the client did something wrong.

server/handler/simulate.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	pr, _, err := client.PullRequests.Get(ctx, owner, repo, number)
+	if err != nil {
+		if isNotFound(err) {
+			logger.Error().Err(err).Msg("could not find pull request")
+			return apierror.WriteAPIError(w, http.StatusNotFound, "failed to find pull request")
+		}
+
+		return errors.Wrap(err, "failed to get pr")
```
bluekeyes: nit:
```suggestion
		return errors.Wrap(err, "failed to get pull request")
```
for consistency with the other messages

server/handler/simulate.go
```diff
@@ -0,0 +1,182 @@
...
+	simulatedCtx, config, err := h.newSimulatedContext(ctx, installation.ID, loc, options)
+	switch {
+	case err != nil:
+		return nil, errors.Wrap(err, "failed to generate eval context")
+	case config.LoadError != nil:
+		return nil, errors.Wrap(config.LoadError, "failed to load policy file")
+	case config.ParseError != nil:
+		return nil, errors.Wrap(config.ParseError, "failed to parse policy")
+	case config.Config == nil:
+		return nil, errors.New("no policy file found in repo")
```
bluekeyes: It might be worth returning an empty / nil result here, especially since picking a different base branch could easily lead to this error.

server/handler/simulate.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	evaluator, err := policy.ParsePolicy(config.Config)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get policy evaluator")
+	}
+
+	result := evaluator.Evaluate(ctx, simulatedCtx)
+	if result.Error != nil {
+		return nil, errors.Wrapf(err, "error evaluating policy in %s: %s", config.Source, config.Path)
```
bluekeyes: `newSimulationResponse` also includes handling for when `result.Error` is not nil. I think we should defer to that and return the error in a "successful" simulation response, since that's how it would appear if Policy Bot encountered the same error in a real evaluation.

server/apierror/error.go
```diff
@@ -0,0 +1,31 @@
...
+	baseapp.WriteJSON(w, code, ErrorResponse{Error: strings.Join(message, "; ")})
```
bluekeyes: It looks like all of the existing usage passes a single error message - I think we should force that via the API (right now, you could also provide no message). If we need multiple errors in the future, marshalling them as an `errors` field that is a list might be better.

policy/simulated/options.go
atatkin: Yeah I considered that, but then `permissions` on `common.Actors` is defined as a `[]pull.Permission` which is a `uint8` and I thought it would be more friendly for the request body to reference the permissions by their string value and convert them into the `uint8` during the conversion function.

server/apierror/error.go
atatkin: Ah yes, it was intentionally written that way to allow an error response without a specific error message, but I ended up never doing that. I'll update it to require a single message.

server/apierror/error.go
atatkin: Yeah it can, it was in a separate package because it was being used by the `middleware` originally too, but now that we are performing the token check in the handler there is no need for it to be in a separate package. Will consolidate.

policy/simulated/context.go
atatkin: Makes sense, updated so the logic was on `Context` instead of the options.

policy/simulated/options.go
atatkin: Good idea, I looked into removing that but it's used a bunch of times by other packages and felt like it would bloat this pr, tracking: https://github.com/palantir/policy-bot/issues/727

policy/simulated/context_test.go
atatkin: Nice catch, thanks!

policy/simulated/options.go
bluekeyes: Ah, `pull.Permission` defines a custom `UnmarshalText` function so that you can use strings. While it's apparently missing the `yaml` tag, the strings work fine when used in YAML configuration and the `encoding/json` package will also use `UnmarshalText` when it is defined.

atatkin: Oh nice, I missed that. I moved the `options` to use the built in structs and added the necessary json tags.

policy/common/actor.go
```diff
@@ -26,17 +26,17 @@ import (
...
+	Organizations []string `yaml:"organizations" json:"organizations"`
 
 	// Deprecated: use Permissions with "admin" or "write"
 	Admins             bool `yaml:"admins"`
 	WriteCollaborators bool `yaml:"write_collaborators"`
 
 	// A list of GitHub collaborator permissions that are allowed. Values may
 	// be any of "admin", "maintain", "write", "triage", and "read".
-	Permissions []pull.Permission
+	Permissions []pull.Permission `json:"permissions"`
```
bluekeyes:
```suggestion
	Permissions []pull.Permission `yaml:"permissions" json:"permissions"`
```

Let's add the `yaml` tag too for consistency, it probably should have been there the whole time.

policy/common/actor.go
```diff
@@ -26,17 +26,17 @@ import (
...
-	Users         []string `yaml:"users"`
-	Teams         []string `yaml:"teams"`
-	Organizations []string `yaml:"organizations"`
+	Users         []string `yaml:"users" json:"users"`
+	Teams         []string `yaml:"teams" json:"teams"`
+	Organizations []string `yaml:"organizations" json:"organizations"`
 
 	// Deprecated: use Permissions with "admin" or "write"
 	Admins             bool `yaml:"admins"`
 	WriteCollaborators bool `yaml:"write_collaborators"`
```
bluekeyes: If you want to ignore these when using JSON, let's add the `json:"-"` tag. Otherwise, they'll be unmarshalled, but use the field name for the keys.

policy/simulated/options.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/palantir/policy-bot/pull"
+	"github.com/pkg/errors"
+)
+
+// Options should contain optional data that can be used to modify the results of the methods on the simulated Context.
+type Options struct {
+	IgnoreComments *common.Actors `json:"ignore_comments"`
+	IgnoreReviews  *common.Actors `json:"ignore_reviews"`
+	AddComments    []pull.Comment `json:"add_comments"`
+	AddReviews     []pull.Review  `json:"add_reviews"`
```
bluekeyes: Sorry, I should have clarified in the last review: I think having custom simulation types for comments and reviews is the right idea. `common.Actors` is already used in public APIs (the configuration file), but the `pull` types are currently internal.

By adding `json` tags to `pull.Comment`/`pull.Review` and using them in the simulation API, we make them part of the public interface and will need to maintain compatibility going forward. They also have fields that aren't really relevant to the simulation right now (like ID, SHA, and teams for reviews).

Custom types here will let us decouple the internal handling of comments and reviews from what we accept in simulation requests so that we can evolve either part without affecting the other.

bluekeyes (commented):
Thanks for all the fixes, I think this is almost ready to merge. Could you please also add a small section to the README about using the new API? I think it can go in the "Testing and Debugging Policies" section.

policy/simulated/options.go
atatkin: Ack, i'll restore the custom types for the `pull.` types and keep the tags for the `common.Actors` type.

policy/simulated/options.go
```diff
@@ -51,30 +51,83 @@ func NewOptionsFromRequest(r *http.Request) (Options, error) {
...
+
+type Review struct {
+	ID           string
+	SHA          string
+	CreatedAt    *time.Time `json:"created_at"`
+	LastEditedAt *time.Time `json:"last_edited_at"`
+	Author       string     `json:"author"`
+	Body         string     `json:"body"`
+	State        string     `json:"state"`
+	Teams        []string   `json:"teams"`
```
bluekeyes: Let's minimize the set of review fields we accept for now - we can always add more later, but taking away fields is harder. I think that means we should accept `CreatedAt,` `LastEditedAt`, `Author`, `Body`, and `State`. For everything else, let's either remove the field or set `json:"-"` so it is ignored when unmarshalling.

bluekeyes (commented):
One more small suggestion on the `Review` type that I think we should address before merging, but then I think this is good to merge.

bluekeyes (approved):

PR #730, opened 2024-03-15
Add support for oauth2 beind proxy

### LINKED ISSUE #724: Unable to run policy-bot behind a reverse-prxoy
Even if configuration states differently ([here](https://github.com/palantir/policy-bot/blob/develop/config/policy-bot.example.yml#L6-L7)), it's not possible to operate the policy-bot behind a reverse proxy.

This is caused by the wrong setup of the github authorization flow in this case.
The github default setup does create the redirect url (callback) at the end of the auth flow with the requests host header. This header is in case of a reverse-proxy set to the ip/host of that proxy.
Instead it should be the configured proxy host.

Can anyone confirm this behavior?

KnisterPeter:
This change does use the public_url config value to create the oauth2 callback url. This logic is only setup
if the config value has been set.

Closes #724

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -218,29 +218,19 @@ func New(c *Config) (*Server, error) {
 		Base: basePolicyHandler,
 	}
 
-	oauth2Params := []oauth2.Param{
-		oauth2.ForceTLS(forceTLS),
-		oauth2.WithStore(&oauth2.SessionStateStore{
-			Sessions: sessions,
-		}),
-		oauth2.OnLogin(handler.Login(c.Github, basePath, sessions)),
-	}
-	if c.Server.PublicURL != "" {
-		u, err := url.Parse(c.Server.PublicURL)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to parse public_url")
-		}
-		u.Path = oauth2.DefaultRoute
-		oauth2Params = append(oauth2Params, oauth2.WithRedirectURL(u.String()))
-	}
-
 	// additional API routes
 	mux.Handle(pat.Get("/api/health"), handler.Health())
 	mux.Handle(pat.Put("/api/validate"), handler.Validate())
 	mux.Handle(pat.Post("/api/simulate/:owner/:repo/:number"), hatpear.Try(simulateHandler))
 	mux.Handle(pat.Get(oauth2.DefaultRoute), oauth2.NewHandler(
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @KnisterPeter! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

server/server.go
```diff
@@ -218,17 +218,29 @@ func New(c *Config) (*Server, error) {
...
 
+	oauth2Params := []oauth2.Param{
+		oauth2.ForceTLS(forceTLS),
+		oauth2.WithStore(&oauth2.SessionStateStore{
+			Sessions: sessions,
+		}),
+		oauth2.OnLogin(handler.Login(c.Github, basePath, sessions)),
+	}
+	if c.Server.PublicURL != "" {
+		u, err := url.Parse(c.Server.PublicURL)
```
bluekeyes: This is already parsed as `publicURL` on L73, so I think you can re-use it here. Since the earlier logic also requires that the value is not empty, I think we can unconditionally set the `WithRedirectURL` parameter.

server/server.go
```diff
@@ -218,17 +218,29 @@ func New(c *Config) (*Server, error) {
 		Base: basePolicyHandler,
 	}
 
+	oauth2Params := []oauth2.Param{
+		oauth2.ForceTLS(forceTLS),
```
bluekeyes: The `ForceTLS` option is only used to generate URLs, so we should be able to remove it now that we're setting a fixed URL.

bluekeyes (commented):
Thanks! I had two minor suggestions that should simplify this.

KnisterPeter:
@bluekeyes That's indeed much cleaner 😃

bluekeyes (approved):

fordneild:
It seems to me that this prohibits hosting policy-bot at a non root url, such as my-domain.com/policy-bot/ since my path is getting wiped by the default one. Am i missing something?

bluekeyes:
@fordneild good point, I think that was an oversight here - the `public_url` value should be allowed to contain subpaths and the OAuth routes should be added to those paths. I filed #804 to track this.

fordneild:
Thanks for the quick fix!

PR #742, opened 2024-03-26
Load pull request draft status from GraphQL API

### LINKED ISSUE #741: has_successful_status causes review requests while PR has draft status
If I interpret the docs correctly:

>  Automatically request reviewers when a Pull Request is opened
>   if this rule is pending, there are no assigned reviewers, and if the
>   Pull Request is not in Draft.

`request_review` should not request reviews if the PR has draft status.

The behaviour seems correct with this config:

```
  - name: Approval
    requires:
      count: 2
      teams:
        - "xxx"
    options:
      allow_author: false
      allow_non_author_contributor: true
      invalidate_on_push: true
      ignore_edited_comments: true
      ignore_update_merges: true
      request_review:
        enabled: true
        mode: random-users
        count: 3
      methods:
        comments: []
        github_review: true
```

but after adding a `has_successful_status` condition, reviews are requested when the checks pass even if the PR remains in draft:

```
  - name: Approval
    if:
      has_successful_status:
        - "checks"
    requires:
      count: 2
      teams:
        - "xxx"
    options:
      allow_author: false
      allow_non_author_contributor: true
      invalidate_on_push: true
      ignore_edited_comments: true
      ignore_update_merges: true
      request_review:
        enabled: true
        mode: random-users
        count: 3
      methods:
        comments: []
        github_review: true
```

(Thanks for a great app btw)

bluekeyes:
Back in 2019, we disabled this field because it was part of a GraphQL preview. As a result, IsDraft was always false for PRs that we loaded via the GraphQL query. For the most part, this doesn't matter, but can cause problems with reviewer requests triggered via webhooks (like status events) that do not provide complete pull request details.

This field has been out of preview for a long time now and is available in all supported versions of GitHub, so it should be safe to load.

Fixes #741.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -1000,9 +1000,7 @@ type v4PullRequest struct {
 	State     string
 
 	IsCrossRepository bool
-
-	// This field is in GraphQL Preview, so don't ask for it just yet
-	IsDraft bool `graphql:""`
+	IsDraft           bool
 
 	HeadRefOID     string
 	HeadRefName    string


asvoboda (approved):
It's always nice fixing an old "temporary" fix from a few years ago (#125)

Issue #627, opened 2023-08-22
Confusing behavior with skipped checks.


Andrei-Predoiu:
I set up the following rule setup due to git API not reporting our "tests" statuses until they are finished.

The goal is to not require approval for dependabot if the test pass but require approvals from everyone else even if the tests pass. 
Instead, when someone other than dependabot makes a PR, the step is skipped and they can merge without approval.

What is the correct way to set my rule up?

![image](https://github.com/palantir/policy-bot/assets/5424946/f62b98b8-b688-4ac8-a0cb-8ea7c0eb20fc)

```
policy:
  approval:
    - or:
      - and:
          - tests passed
          - owner has approved
      - and:
          - dependabot is making the PR
          - or:
              - tests passed
              - owner has approved
  disapproval:
    requires:
      organizations:
        - "org"
approval_rules:
  - name: tests passed
    if:
      has_successful_status:
        - "security checks"
        - "tests"
    requires:
      count: 0
  - name: owner has approved
    requires:
      count: 1
      teams:
        - "org/team"
  - name: dependabot is making the PR
    if:
      has_author_in:
        users:
          - "dependabot[bot]"
          - "dependabot-circleci[bot]"
    requires:
      count: 0
```

bluekeyes:
The way we usually solve this is to directly make the `tests` status a required status check in GitHub. This way, the policy only reflects approval.

If you need to combine this all in the policy for some reason (maybe you want human developers to be able to ignore failing tests if they have approval), try combining the conditions in a single rule:

```yaml
policy:
  approval:
    - or:
      - owner has approved
      - depandabot PR has passing tests
  disapproval:
    requires:
      organizations:
        - "org"

approval_rules:
  - name: owner has approved
    requires:
      count: 1
      teams:
        - "org/team"
  - name: dependabot PR has passing tests
    if:
      has_author_in:
        users:
          - "dependabot[bot]"
          - "dependabot-circleci[bot]"
      has_successful_status:
        - "security checks"
        - "tests"
    requires:
      count: 0
```

misund:
In your suggested config, the `dependabot PR has passing tests` rule will be satisfied while the `security checks` and `tests` status checks are pending, even if they eventually will fail or time out. That is the confusing behavior.

The key name is `has_successful_status`, but the behavior is more like "does not have unsuccessful status".

Andrei-Predoiu:
[...]

PR #750, opened 2024-04-04
Accept successful status checks as approval

bluekeyes:
Many users have requested a way to write policies that conditionally require status checks, for instance by only requiring passing tests for automated dependency update pull requests. While Policy Bot has a predicate for status checks, in practice it was hard to write these types of policies using predicates, because they skips the rule when the status is missing/pending/failed, instead of leaving the rule pending.

This change makes it possible to require passing status checks as an approval condition for a rule. For conditional status check policies, this means you can write rules that remain pending until the passing status checks are present. It's also possible to combine status check approval with normal actor-based approval.

See #627

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,15 +420,14 @@ requires:
   # requests for details.
   permissions: ["write"]
 
-  # Deprecated: use 'permissions: ["admin"]'
+  # "statuses" is a list of status checks that must be successful for the rule
+  # to count as approved. If set, statuses are an additional requirement beyond
+  # the approvals required by "count".
   #
-  # Allows approval by admins of the org or repository
-  # admins: true
-
-  # Deprecated: use 'permissions: ["write"]'
-  #
-  # Allows approval by users who have write on the repository
-  # write_collaborators: true
+  # For example, if "count" is 1 and "statuses" contains the "build" and "test"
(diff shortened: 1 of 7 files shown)


Andrei-Predoiu:
Thanks for tackling this topic, your changes would add some much awaited functionality.
Firstly, this only works for statuses and it would be nice to work for all rules.
In policy bot the rules are the ones that are skipped or passed.
My suggestion  is to add a new parameter to all rules that can change the behavior so a rule cannot be skipped. Something like "enforce" or "mandatory"

Fx:
```yaml
- name: dependabot is making the PR
  mandatory: true
  if:
    has_author_in:
      users:
        - "dependabot[bot]"
        - "dependabot-circleci[bot]"
```

bluekeyes:
Thanks for the suggestion. I don't think the idea of a `mandatory` field fits well with the rest Policy Bot's design, as it changes the behavior and meaning of predicates, but I think I understand the behavior you'd like to see. 

I created an alternate proposal in #752 that adds a new `conditions` field to the `requires` section. This field can accept any key that is currently allowed in the `if` block, but instead of skipping or enabling the rule, it will contribute to the approved or pending status.

bluekeyes:
Closing in favor of #752.

PR #752, opened 2024-04-06
Add support for approval conditions

bluekeyes:
Many users try to use Policy Bot as a general purpose rule engine, constructing policies that enforce things about status checks, labels, and other conditions. This was poorly supported because predicates skip rules when the conditions don't match, whereas users working in this model expect the rule to apply, but remain in the pending state.

This commit introduces the concept of "conditions", which are predicates that count toward approval. All of the existing predicates tests are available, but when used as a condition they control whether the rule is pending or approved, instead of active or skipped.

Conditions can combine with regular approval requirements, but I expect most uses will either use predicates and regular approval or conditions without additional approval.

Due to limitations in how predicates work internally, the details UI shows limited information about conditions. I hope to fix this in a follow-up, but I wanted to separate the main change from the predicate refactoring.

Finally, this adds a new README section explaining how to design a policy and what all the different terms mean.

I'm open to a different name for this new feature, since it's easy to accidentally say "condition" when talking about predicates, but I couldn't think of anything else that fit better.

This is an alternative to #750, which only works for status checks. See also #627.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ UI to view the detailed approval status of any pull request.
 [required status check]: https://help.github.com/articles/enabling-required-status-checks/
 [required reviews]: https://help.github.com/articles/about-required-reviews-for-pull-requests/
 
+- [Designing Policies](#designing-policies)
 - [Configuration](#configuration)
   - [policy.yml Specification](#policyyml-specification)
   - [Approval Rules](#approval-rules)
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -40,11 +41,83 @@ UI to view the detailed approval status of any pull request.
...
+When writing a rule, you can react to the state of pull request by using either
+predicates or required conditions. When would you use one over the other?
+
+* Use **predicates** to enable a rule when a pull request is in a specific
+  state (or to skip the rule when a pull request is _not_ in that state.)
+  Predicates are useful when you want special approval requirements like
+  requiring extra approval for files that involve security or automatically
+  approving dependency updates.
+
+* Use **conditions** when the pull request state itself is important for
```
TomVanGreevenbroek, line 107: Did you mean `requirements` here?

README.md
```diff
@@ -40,11 +41,83 @@ UI to view the detailed approval status of any pull request.
...
+* `skipped` - one or more predicates are not true
+* `error` - something went wrong while evaluating the rule
+
+For the purposes of the `and` and `or` operators:
+
+* `approved` is equivalent to `true`
+* `pending` and `error` are equivalent to `false`
+* `skipped` completely removes the rule from the condition
+
+### Predicates vs Conditions
```
TomVanGreevenbroek: Did you mean `Requirements` instead of `Conditions` here? Or maybe in line 75 you mean to put `Conditions` rather than `Requirements`

Andrei-Predoiu:
I think this is a good solution.

server/templates/details.html.tmpl
```diff
@@ -213,6 +221,11 @@
 
 {{define "result-reviews-count"}}This rule requires at least {{.Count}} approval{{if gt .Count 1}}s{{end}}{{end}}
 
+{{define "result-conditions-details"}}
+  {{/* TODO(bkeyes): this is a placeholder until I can refactor predicate rendering */}}
```
asvoboda, line 225: If this something you want to do in a follow up PR?

asvoboda (dismissed):
This seems like a pretty straightforward approach to including conditions.

I'm less certain about how we'd want to use this internally, but it seems like a relatively friendly feature to include for oss.

server/templates/details.html.tmpl
bluekeyes: Yeah, trying to refactor the predicate display in this PR was going to overwhelm the actual changes.

README.md
bluekeyes: Updated to use "required conditions" - "requirements" includes both normal review/approval from users and the new conditions feature, so I want to distinguish between them.

nvelat (approved):

bluekeyes:
If you are able to deploy snapshot versions, this feature is now available in the `palantirtechnologies/policy-bot:snapshot` image for testing.

PR #755, opened 2024-04-10
feat: changed_files.exclude_paths

### LINKED ISSUE #754: Feature Request: Predicate to skip rule if a file was changed
### Situation

We are running into limitations with policy bot and would like a new feature. 

Given 2 PRs which should each trigger a rule specific to them 
- PR 1 - trigger rule 1 and skip rule 2 
  - a file is introduced: `foo/bar.json
  - an assortment of miscellaneous files are edited at the root or in many other directories
  - no files in `foo/archive/` are touched
- PR 2 - trigger rule 2 and skip rule 1
	- the file is moved: `foo/bar.json -> foo/archive/bar.json`

In this exact scenario, there's no way simple way to create a rule which would apply two rules separately. This is because there's no way to negate/exclude certain paths explicitly. If go supported negative look-ahead regex statements, this would not be a problem.

<details>
<summary>Current Rule Predicates (Convoluted)</summary>

```yaml
- name: Rule 1
    if:
      # The idea here is to include all possible file changes during PR 1.
      # There is no way to negate paths, so we have to create an exhaustive list.
      # Which will effectively allow us to negate paths, by excluding them from the list.
      # Purposefully Ignored Paths
      # - .github/
      # - foo/*/ (for the purpose of excluding directory/archive)
      # - node_modules/
      only_changed_files:
        paths:
          # json changes in root directory - ie. no / in the path/filename
          - "^[a-zA-Z0-9_.-]+json$"
          # specific dot files in root directory
          # changed files in directory/* but not in directory/*/* (ie. no subdirectories)
          - "^foo/[^/]+$"
          # changes in every other subdirectory (allows recursive subdirectories)
          - "^directory-1/.*"
          - "^directory-2/.*"
# ...

- name: Rule 2 
    if:
      changed_files:
        paths:
          - "^foo/archive/.*"
...
```

</details>

### Proposed Feature

Have a `excludePaths` feature for the `changed_files` predicate. 

Behaviour:

* `changed_files` is satisfied if:
  * any file in the pull request matches any regular expression in the "paths" list, and every file in the pull request does not match any regular expression in the `excludePaths` list
  * If the "ignore" list is present, files in the pull request matching these regular expressions are ignored by both the `paths` and `excludePaths` rules

### Solution

Having a functionality similar to the proposed `excludePaths` in `changed_files` would solve the above scenario we are facing because we could negate files explicitly, rather than having to try and negate files by creating an exhaustive list with `only_changed_files`.
- Example solution with changed files and exclude paths:  `paths: "^foo/a-zA-Z0-9_.-+$"` and `excludePaths: "^foo/archive/.*`

---

I already have some local changes which I believe implement this functionality. Figured I'd open an issue and see if there's any alternatives suggested for negating files ~~before I PR this~~.

erikburt:
Adds a field to the `changed_files` predicate that allows you to exclude paths, ensuring that the rule is not applied if a file matching an entry in `changed_files.exclude_paths` is present.

---

Closes #754

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ UI to view the detailed approval status of any pull request.
 [required status check]: https://help.github.com/articles/enabling-required-status-checks/
 [required reviews]: https://help.github.com/articles/about-required-reviews-for-pull-requests/
 
+- [Designing Policies](#designing-policies)
 - [Configuration](#configuration)
   - [policy.yml Specification](#policyyml-specification)
   - [Approval Rules](#approval-rules)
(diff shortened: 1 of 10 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @erikburt! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #756, opened 2024-04-17
feat: no_changed_files predicate

### LINKED ISSUE #754: Feature Request: Predicate to skip rule if a file was changed

erikburt:
### Summary 

Adds a new `no_changed_files` predicate that allows you to ensure that the rule is not applied if a file matching an entry in `no_changed_files.paths` is present.

This is a negation of the `changed_files` predicate, and the implementation logic reflects that.

Closes #754
Related #755


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @erikburt! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

policy/predicate/file.go
```diff
@@ -141,6 +141,49 @@ func (pred *OnlyChangedFiles) Trigger() common.Trigger {
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	predicateResult := common.PredicateResult{
+		Values:          changedFilesPredicateResult.Values,
+		Satisfied:       !changedFilesPredicateResult.Satisfied,
+		ValuePhrase:     "excluded changed files",
+		ConditionPhrase: "match",
```
erikburt: Wasn't sure how to word these.

Perhaps: 
```
ValuePhrase:     "changed files",
ConditionPhrase: "cannot match",
```

policy/predicate/file.go
```diff
@@ -141,6 +141,49 @@ func (pred *OnlyChangedFiles) Trigger() common.Trigger {
...
+		ConditionsMap: map[string][]string{
+			"path patterns":  changedFilesPredicateResult.ConditionsMap["path patterns"],
+			"while ignoring": changedFilesPredicateResult.ConditionsMap["while ignoring"],
+		},
```
erikburt:
```suggestion
		ConditionsMap:   changedFilesPredicateResult.ConditionsMap,
```

bluekeyes:
Thanks for implementing this. I think it looks good, but I need to think a bit more about:

* If the expected behavior is exactly the negation of `changed_files` or if there should be some difference (particularly for ignored files)
* If the value/condition phrases are correct for the UI

I'll try to get back to you on those points in the next few days.

policy/predicate/file.go
```diff
@@ -141,6 +141,44 @@ func (pred *OnlyChangedFiles) Trigger() common.Trigger {
...
+		return nil, err
+	}
+
+	predicateResult := common.PredicateResult{
+		Values:          changedFilesPredicateResult.Values,
+		Description:     changedFilesPredicateResult.Description,
+		Satisfied:       !changedFilesPredicateResult.Satisfied,
+		ValuePhrase:     "changed files",
+		ConditionPhrase: "cannot match",
+		ConditionsMap:   changedFilesPredicateResult.ConditionsMap,
```
bluekeyes: The way phrases are used is pretty awkward for negative conditions: when the predicate is skipped the `ConditionPhrase` appears after `do not`. In this case that leads to "The changed files [...] do not cannot match path patterns ..."

The best single word I can think of right now is "avoid", leading to "The changed files [...] avoid path patterns ..." and "The changed files [...] do not avoid path patterns ...". This works, but is not great.

A more involved option is to modify the template to do something different here. Maybe if the `ConditionPhrase` value already has `do not` as a prefix it could remove it when the predicate is skipped instead of adding it.

policy/predicate/file.go
```diff
@@ -141,6 +141,44 @@ func (pred *OnlyChangedFiles) Trigger() common.Trigger {
...
+	}
+
+	if changedFilesPredicateResult.Satisfied {
+		predicateResult.Description = "No changed files match the excluded patterns"
+	}
```
bluekeyes: I think this is condition is inverted. When `changedFilesPredicateResult.Satisfied` is true, the `Description` field contains the file that matched, which is the correct explanation for why this rule is _not_ satisfied. When it is false, we can change the description to "No changed files match the specified patterns"

(I think "excluded" gives the wrong idea, so "specified" is a bit more generic than the original "required")

bluekeyes (commented):
I think this is good in terms of the logic: a `no_changed_files` predicate should be the exact opposite of a `changed_files` predicate. But I had some suggestions for improving how this displays in the details UI.

I'm also thinking that the logic for the metadata values (description, conditions, etc.) might be clearer if the predicate was implemented directly, instead of by modifying the results of the `ChangedFiles` predicate. Most of the code in both predicates is for setting up this metadata, so we're not saving all that much code by delegating to the other predicate.

erikburt:
Thank you for the review! Sorry for the delay, was caught up with some deadlines.

I've changed the phrases around using some of your suggestions. To get around the weird double negatives I added a "SkipPhrase" to the predicate result which is universally "do not" except for the NoChangedFiles predicate where it is empty.

I've also tried to simplify the logic where possible. I do think it's probably best to reuse the ChangedFiles predicate to ensure this is a direct negation. If you would like me to duplicate and reverse the logic, I can do that as well.

Let me know what you think.

server/templates/details.html.tmpl
```diff
@@ -158,7 +158,7 @@
         <ul class="list-disc list-outside pl-6 py-2">
           {{range .Values}}<li class="font-mono text-sm-mono">{{.}}</li>{{end}}
         </ul>
-        {{if eq $s "skipped"}}do not {{end}}{{.ConditionPhrase}}
+        {{if eq $s "skipped"}}{{.SkipPhrase}} {{end}}{{.ConditionPhrase}}
```
bluekeyes: One issue with using `SkipPhrase` like this is that we'll need to either set a value in _all_ of the existing predicates or provide a way to use a default value when it isn't set.

policy/predicate/file.go
```diff
@@ -141,6 +143,47 @@ func (pred *OnlyChangedFiles) Trigger() common.Trigger {
...
+		Satisfied:     !changedFilesPredicateResult.Satisfied,
+		Values:        changedFilesPredicateResult.Values,
+		ConditionsMap: changedFilesPredicateResult.ConditionsMap,
+	}
+
+	if predicateResult.Satisfied {
+		predicateResult.ConditionPhrase = "do not match"
+		predicateResult.Description = "No changed files match the specified patterns"
+	} else {
+		predicateResult.ConditionPhrase = "should not match"
```
bluekeyes: I think this may be better as `"match"`. That way, the overall expression will read `This rule is skipped because the changed files [...] match the path patterns [...]`

bluekeyes (commented):
Thanks for the updates. I think the `SkipPhrase` makes sense but we'll need to either provide a default value or set `"do not"` as the value in all existing predicates to avoid breaking the details display for non-file predicates.

policy/predicate/file.go
erikburt: Sounds good - I've changed this to "match" in both scenarios with the reworked use of skip phrase.

server/templates/details.html.tmpl
erikburt: I've modified the code so a skip phrase is not directly needed. The logic now allows you to reverse when to add the "do not", defaulting to current behaviour for all predicates.

bluekeyes (approved):
Thanks for the updates!

PR #789, opened 2024-06-27
Add support for allowing skipped checks in `has_successful_status`

### LINKED ISSUE #760: Feature Request: Option to count skipped jobs in has_successful_status
GitHub branch protection rules and rulesets [count skipped jobs as "success"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks). Given GitHub's underpowered required checks config, this "feature" is commonly used with [path filtering](https://github.com/dorny/paths-filter) to avoid running unnecessary jobs in monorepos etc.

We also use policy-bot to ping reviewers directly when a PR is in a reviewable state (including minimum checks pass).

As `has_successful_status` only counts "success" state we have had to consider these workarounds:

#### 1. Only policy-bot as the required check

The path filtering is moved to the policies - expensive jobs run unnecessarily

#### 2. Duplicate the conditions in policy-bot

High risk of the filtering logic getting out-of-sync

#### 3. Additional job (chosen)

Add an additional job that is dependent on the conditional jobs and fails if any of the conditional job results is not "success" or "skipped". This job is referenced in `has_successful_status`.

Small extra expense, some additional complexity and confusing PR check summary.

Thanks for your time!

iainlane:
The `has_successful_status` predicate currently requires predicates to be present and successful in order to pass. But workflows can be run conditionally - for example only if certain paths change - and it is currently not very convenient to write a policy which considers such skipped workflows as passing. The only feasible workaround is to duplicate the path filters in the policy, and this quickly gets unwieldy and prone to getting out-of-sync in large repositories.

Here we add direct support for specifying such rules. This is done by introducing a new alernative form that `has_successful_status` predicates can take:

```yaml
has_successful_status:
  options:
    skipped_is_success: true
  statuses:
    - "status 1"
    - "status 2"
```

In this mode, we will consider the `skipped` result as acceptable. The current form:

```yaml
has_successful_status:
  - "status 1"
  - "status 2"
```

remains supported. We have done this by implementing a custom unmarshaling function to be able to handle both forms.

Closes: #760

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,24 +322,15 @@ if:
     deletions: "> 100"
     total: "> 200"
 
-  # "has_successful_status" is satisfied if the status checks that are specified
-  # are marked successful on the head commit of the pull request.
-  has_successful_status:
-    - "status-name-1"
-    - "status-name-2"
-    - "status-name-3"
-
-  # "has_successful_status" can be configured to count "skipped" statuses as
-  # successful. This can be useful in combination with path filters where
-  # workflows only run on parts of the tree. They are required to succeed only
-  # if they run.
-  # has_successful_status:
-  #  options:
-  #    skipped_is_success: true
-  #  statuses:
-  #    - "status-name-1"
-  #    - "status-name-2"
-  #    - "status-name-3"
+  # "has_status" is satisfied if the status checks that are specified are
+  # finished and concluded with one of the conclusions specified.
+  # "conclusions" is optional and defaults to ["success"].
+  has_status:
+    conclusions: ["success", "skipped"]
+    statuses:
+      - "status-name-1"
+      - "status-name-2"
+      - "status-name-3"
 
   # "has_labels" is satisfied if the pull request has the specified labels
   # applied
@@ -522,16 +513,17 @@ requires:
   # count as approved. If present, conditions are an additional requirement
   # beyond the approvals required by "count".
   #
-  # For example, if "count" is 1 and "conditions" contains the "has_successful_status"
+  # For example, if "count" is 1 and "conditions" contains the "has_status"
   # condition with the "build" status, the rule is only approved once the
   # "build" status check passes and one authorized reviewer leaves a review.
   conditions:
-    # The "conditions" block accepts all of the keys documented as part
-    # of the "if" block for predicates. The "has_successful_status" key is
-    # shown here as an example of one type of condition.
-    has_successful_status:
-      - "build"
-      - "vulnerability scan"
+    # The "conditions" block accepts all of the keys documented as part of the
+    # "if" block for predicates. The "has_status" key is shown here as an
+    # example of one type of condition.
+    has_status:
+      statuses:
+        - "build"
+        - "vulnerability scan"
 ```
 
 ### Approval Policies
--- a/README.md
+++ b/README.md
@@ -329,6 +329,18 @@ if:
     - "status-name-2"
     - "status-name-3"
 
+  # "has_successful_status" can be configured to count "skipped" statuses as
+  # successful. This can be useful in combination with path filters where
+  # workflows only run on parts of the tree. They are required to succeed only
+  # if they run.
+  # has_successful_status:
+  #  options:
+  #    skipped_is_success: true
+  #  statuses:
+  #    - "status-name-1"
+  #    - "status-name-2"
+  #    - "status-name-3"
+
   # "has_labels" is satisfied if the pull request has the specified labels
   # applied
   has_labels:
--- a/README.md
+++ b/README.md
@@ -322,6 +322,14 @@ if:
     deletions: "> 100"
     total: "> 200"
 
+  # DEPRECATED: Use "has_status" below instead, which is more flexible.
+  # "has_successful_status" is satisfied if the status checks that are specified
+  # are marked successful on the head commit of the pull request.
+  has_successful_status:
+    - "status-name-1"
+    - "status-name-2"
+    - "status-name-3"
+
   # "has_status" is satisfied if the status checks that are specified are
   # finished and concluded with one of the conclusions specified.
   # "conclusions" is optional and defaults to ["success"].
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -48,8 +48,6 @@ type Options struct {
 
 	RequestReview RequestReview `yaml:"request_review"`
 
-	CountSkippedStatusAsPassed bool `yaml:"count_skipped_status_as_passed"`
-
 	Methods *common.Methods `yaml:"methods"`
 }
 
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -48,6 +48,8 @@ type Options struct {
 
 	RequestReview RequestReview `yaml:"request_review"`
 
+	CountSkippedStatusAsPassed bool `yaml:"count_skipped_status_as_passed"`
+
 	Methods *common.Methods `yaml:"methods"`
 }
 
--- a/policy/predicate/predicate.go
+++ b/policy/predicate/predicate.go
@@ -16,6 +16,9 @@ package predicate
 
 import (
 	"context"
+	"fmt"
+	"slices"
+	"strings"
 
 	"github.com/palantir/policy-bot/policy/common"
 	"github.com/palantir/policy-bot/pull"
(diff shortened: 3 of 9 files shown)


palantirtech:

iainlane:
I'll square off the CLA question internally as soon as I can, but in the meantime hopefully we can proceed with reviews 😁

bluekeyes:
Thanks for this proposal! Looking at an actual implementation, I'm now thinking this could work better as a new predicate that deprecates `has_successful_status`. This new `has_status` predicate would let you specify both the contexts and the allowed conclusions:

```yaml
# Option 1: same conclusions for all contexts
has_status:
  conclusions: ["success", "skipped"]
  contexts:
    - "status 1"
    - "status 2"

# Option 2: conclusions per context
has_status:
  "status 1": ["success"]
  "status 2": ["success", "skipped"]
```

Internally, we can convert the `has_successful_status` predicate in to this new one to avoid duplicating logic. The new predicate also avoids having two forms of a single predicate and allows policies that can work with failure, neutral, and other conclusions. What do you think?

iainlane:
Cheers for the review 👍

> Thanks for this proposal! Looking at an actual implementation, I'm now thinking this could work better as a new predicate that deprecates `has_successful_status`. 

Sure, that makes sense. It was a bit awkward that we had two forms in one predicate.

Take a look at what I've got now? It's fundamentally the same as what I had before, except now:-

1. `has_successful_status` is kept separate
2. Any conclusion can be given, instead of only allowing `skipped` as we did before

We've still got a custom unmarshal function here, which handles unpacking either form into the same struct so they can share the implementation.

A bit more stuff has appeared to handle the list of statuses but it's not too bad.

> ```yaml
> # Option 1: same conclusions for all contexts
> has_status:
>   conclusions: ["success", "skipped"]
>   contexts:
>     - "status 1"
>     - "status 2"
> 
> # Option 2: conclusions per context
> has_status:
>   "status 1": ["success"]
>   "status 2": ["success", "skipped"]
> ```

I went for the first one. You can express the second in terms of it as far as I can see, and it feels to me like it'll be more common to want to apply the same conclusions to a whole predicate than select them individually. But I don't feel that strongly about this - it can be changed to the other form easily enough.

I've used `statuses` rather than `contexts`. That's only because I've not used that word in the context of GitHub CI and I found it bounced off me a bit. But it also doesn't matter too much.

iainlane:
I was just testing this in a real deployment, instead of using the testsuite... and this doesn't work like I thought it did.

Given this workflow

```yaml
on:
  pull_request:
    paths:
      - a/**

jobs:
   # some jobs here
```

There is _nothing_ reported when the job isn't run due to the path filter not being matched. Or at least nothing that I can find.

I think the PR is still useful for conditional _jobs_, but not conditional _workflows_, unless there is a way we can see those.

bluekeyes:
> There is _nothing_ reported when the job isn't run due to the path filter not being matched. Or at least nothing that I can find.

Yeah, that sounds right. If a workflow trigger doesn't match, GitHub skips doing anything with the workflow. Posting `skipped` statuses for the jobs probably requires doing at least some evaluation to determine what jobs exist.

It's pretty difficult to account for things like this, since all statuses start out as missing until the reporting system responds to a webhook and posts the initial value. When Policy Bot evaluates a PR, there's no way that I know of to distinguish between a status that will never appear and one that will appear, but just hasn't appeared yet.

If you're able to share more details about what you want to achieve with your policy (maybe on the original issue), I can try to think about alternate approaches that might work.

policy/approval/approve.go
```diff
@@ -48,6 +48,8 @@ type Options struct {
 
 	RequestReview RequestReview `yaml:"request_review"`
 
+	CountSkippedStatusAsPassed bool `yaml:"count_skipped_status_as_passed"`
```
bluekeyes: I think this is leftover from a previous iteration? It appears unused now

README.md
```diff
@@ -322,12 +322,15 @@ if:
     deletions: "> 100"
     total: "> 200"
 
-  # "has_successful_status" is satisfied if the status checks that are specified
-  # are marked successful on the head commit of the pull request.
-  has_successful_status:
-    - "status-name-1"
-    - "status-name-2"
-    - "status-name-3"
```
bluekeyes: For the benefit of people with older policies, I'd like to keep the documentation for this predicate, but with a note that it is deprecated:

```
Deprecated: Use `has_status` instead, which is more flexible
```

policy/predicate/status.go
```diff
@@ -16,37 +16,76 @@ package predicate
...
 
-func (pred HasSuccessfulStatus) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
-	statuses, err := prctx.LatestStatuses()
+// UnmarshalYAML implements the yaml.Unmarshaler interface for HasStatus.
+// This supports unmarshalling the predicate in two forms:
+//  1. A list of strings, which are the statuses to check for. This is the
+//     deprecated `has_successful_status` format.
+//  2. A full structure with `statuses` and `conclusions` fields as in
+//     `has_status`.
+func (pred *HasStatus) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
bluekeyes: While this works, it does mean people could use the new `has_status` predicate by providing only a list of strings or use the old `has_successful_status` predicate with a `conclusions` list that doesn't contain `success`.

Instead, I suggest keeping the old `HasSuccessfulStatus` type and replacing the `Evaluate` method with a version that creates a `HasStatus` object and calls `Evaluate` on it. That would keep these as two separate predicates for policy parsing purposes, while avoiding most of the code duplication.

policy/predicate/predicate.go
```diff
@@ -27,3 +30,61 @@ type Predicate interface {
...
 	Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error)
 }
+
+type unit struct{}
+type set map[string]unit
+
+// allowedConclusions can be one of:
+// action_required, cancelled, failure, neutral, success, skipped, stale,
+// timed_out
+type allowedConclusions set
```
bluekeyes: Curious for your opinion here: there's a decent amount of code in this PR for converting between this set representation and a list of strings in different circumstances. Given that the number of elements in the set is almost always going to be less than 3, what do you think about using a simple `[]string` and testing for matching conclusions with `slices.Contains`?

That could allow duplicates, but duplicates would be obvious in the policy definition and shouldn't cause any problems (other than slightly awkward messages.)

policy/approval/approve.go
iainlane: Yes indeed. I've dropped it in 2d9c41c83e58ded7d2e42e25eed00bf6db0feeee.

README.md
iainlane: Yep sure: 06ae646f577a08c365931527addba40d63ee0e86

policy/predicate/predicate.go
iainlane: Yeah no worries. I'm used to writing such code so I tend to reach for it but it's fine to skip it in this case. See ebee53203ad74705387c4abe080d22ce31fe7f4e.

If needed we could dedupe when printing, when looking at the slice or any other time. But I doubt it'll be a big deal since we're talking about a small number of possible values.

policy/predicate/status.go
iainlane: Makes sense. Done here: 1793921421e208e47ae226e1d93934152b09bc84.

Since we brought the type back, I added it to the testsuite again. Then it started to look a bit unwieldy with all the cases going on, so I reworked that a bit to be hopefully clearer vs. my previous.

iainlane (commented):
Thanks for the second look. All the comments should be addressed now. 👍

> If you're able to share more details about what you want to achieve with your policy (maybe on the original issue), I can try to think about alternate approaches that might work.

Okey dokey, I'll do that.

bluekeyes (approved):
Looks good, thanks for iterating on this with me. Let me know when you have the CLA sorted out (or if you have any questions about it) and then we can merge this.

iainlane:
> Looks good, thanks for iterating on this with me. Let me know when you have the CLA sorted out (or if you have any questions about it) and then we can merge this.

I got the reply from our legal folks earlier. It's good to go and I've signed it.

Thanks - if you could hit the button for me, please. 👍

I'll get to the other PRs and issues tomorrow now I shouldn't be blocking from this side any more.

stevoland:
Thanks so much to both of you, excellent stuff!

bluekeyes:
@iainlane while looking at #807, I realized we may have a bug related to this: the handler for the `check_run` event, and probably the handler for the `status` event, exits early when the conclusion of the incoming status is not `success`. This was an optimization to make fewer requests, but it means we could miss evaluating a policy that requires some other status if there are no other triggering events.

Unfortunately, this could complicate the request optimizations you might be looking at. While we have the "trigger" concept for policies that deal with statuses, it doesn't include the different conclusions, so we may needlessly evaluate policies on pending events when they could only be satisfied by successes.

iainlane:
> @iainlane while looking at #807, I realized we may have a bug related to this: the handler for the `check_run` event, and probably the handler for the `status` event, exits early when the conclusion of the incoming status is not `success`. This was an optimization to make fewer requests, but it means we could miss evaluating a policy that requires some other status if there are no other triggering events.
> 
> Unfortunately, this could complicate the request optimizations you might be looking at. While we have the "trigger" concept for policies that deal with statuses, it doesn't include the different conclusions, so we may needlessly evaluate policies on pending events when they could only be satisfied by successes.

Ah, yeah, hmm. Good spot.

Would it be acceptable to look for `status` being `completed` instead? So we'd only react when runs are finished, but now if they finished with any conclusion.

Or would we be looking at figuring out to get at the set of interesting conclusions in the handler? Feels tricky.

bluekeyes:
Yeah, I think restricting `has_status` to completed conclusions and then looking for that would be reasonable. If someone has a compelling reason to write a policy that matches on pending statuses in the future, we can revisit at that point.

PR #794, opened 2024-07-07
feat(predicate): Add `has_workflow_result` predicate

iainlane:
This is a new predicate which matches on the results of an entire GitHub Actions workflow.

This is similar to (doesn't replace, both are complementary) `has_status` from #789, except `has_status` matches on statuses or check runs, which for GitHub Actions are basically equivalent to _jobs_ in a Workflow. `has_workflow_result` finds the check _suite_ for the PR and reports its conclusion, which is the conclusion of the whole workflow run - of all of its jobs.

With `has_workflow_result`, the workflows are given as paths in the repository, as these can't be dynamic and are easy to predict when writing policies, as well as easy to review for correctness.

With this, you can write

```yaml
has_workflow_result:
  workflows:
    - .github/workflows/ci.yml
```

and then we'll wait for all the jobs in this workflow to succeed (or with the optional `conclusions`, to conclude with whatever you want), which is more convenient than listing all the job names out. This new predicate is really to make writing rules easier. At least for the workflows I use daily it's much more common to want to wait for every job than just a subset of them. 🙂 

For the avoidance of doubt, it doesn't solve the problem I mentioned in #789 where we don't see completely skipped jobs (if the `on` filters resulted in a skip).

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,19 @@ if:
       - "status-name-2"
       - "status-name-3"
 
+  # "has_workflow_result" is satisfied if the GitHub Actions workflow runs that
+  # are specified all finished and concluded with one of the conclusions
+  # specified. "conclusions" is optional and defaults to ["success"].
+  # `workflows` contains the paths to the workflow files that are being checked.
+  # If a workflow is run more than once for a commit - for example for a `push`
+  # and `pull_request` event, the most recent completed run for each event type
+  # will be considered.
+  has_workflow_result:
+    conclusions: ["success", "skipped"]
+    workflows:
+      - ".github/workflows/a.yml"
+      - ".github/workflows/b.yml"
+
   # "has_labels" is satisfied if the pull request has the specified labels
   # applied
   has_labels:
@@ -995,6 +1008,7 @@ The app requires these permissions:
 
 | Permission | Access | Reason |
 | ---------- | ------ | ------ |
+| Actions| Read-only | Read workflow run events for the `has_workflow_result` predicate |
 | Repository contents | Read-only | Read configuration and commit metadata |
 | Checks | Read-only | Read check run results |
 | Repository administration | Read-only | Read admin team(s) membership |
@@ -1013,6 +1027,7 @@ The app should be subscribed to these events:
 * Pull request
 * Pull request review
 * Status
+* Workflow Run
 
 There is a [`logo.png`](https://github.com/palantir/policy-bot/blob/develop/logo.png)
 provided if you'd like to use it as the GitHub application logo. The background
--- a/policy/predicate/workflow.go
+++ b/policy/predicate/workflow.go
@@ -1,0 +1,94 @@
+// Copyright 2018 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package predicate
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/palantir/policy-bot/policy/common"
+	"github.com/palantir/policy-bot/pull"
+	"github.com/pkg/errors"
+)
+
+type HasWorkflowResult struct {
+	Conclusions AllowedConclusions `yaml:"conclusions,omitempty"`
+	Workflows   []string           `yaml:"workflows,omitempty"`
+}
+
+func NewHasWorkflowResult(workflows []string, conclusions []string) *HasWorkflowResult {
+	return &HasWorkflowResult{
+		Conclusions: conclusions,
+		Workflows:   workflows,
+	}
+}
+
+var _ Predicate = HasWorkflowResult{}
+
+func (pred HasWorkflowResult) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
+	workflowRuns, err := prctx.LatestWorkflowRuns()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list latest workflow runs")
+	}
+
+	allowedConclusions := pred.Conclusions
+	if len(allowedConclusions) == 0 {
+		allowedConclusions = AllowedConclusions{"success"}
+	}
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "workflow results",
+		ConditionPhrase: fmt.Sprintf("exist and have conclusion %s", allowedConclusions.joinWithOr()),
+	}
+
+	var missingResults []string
+	var failingWorkflows []string
+	for _, workflow := range pred.Workflows {
+		conclusions, ok := workflowRuns[workflow]
+		if !ok {
+			missingResults = append(missingResults, workflow)
+		}
+		for _, conclusion := range conclusions {
+			if !slices.Contains(allowedConclusions, conclusion) {
+				failingWorkflows = append(failingWorkflows, workflow)
+			}
+		}
+	}
+
+	if len(missingResults) > 0 {
+		predicateResult.Values = missingResults
+		predicateResult.Description = "One or more workflow runs are missing: " + strings.Join(missingResults, ", ")
+		predicateResult.Satisfied = false
+		return &predicateResult, nil
+	}
+
+	if len(failingWorkflows) > 0 {
+		predicateResult.Values = failingWorkflows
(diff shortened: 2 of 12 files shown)


palantirtech:

pull/github.go
```diff
@@ -26,6 +26,10 @@ import (
 )
 
 const (
+	// GitHubAppID is the GitHub Actions App ID. When a check suite is created
+	// by GitHub Actions, its `app.id` field is set to this value.
+	GitHubAppID = 15368
```
bluekeyes: If we have to use this (which might not be necessary if we switch to the REST API), it will need to be configurable. For example, on one of our internal GitHub Enterprise instances, the App ID for the `github-actions` app is 152.

Another alternative might be to filter based on the app name (which should be consistent) instead of the ID.

pull/github.go
```diff
@@ -829,6 +834,89 @@ func (ghc *GitHubContext) getCheckStatuses() (map[string]string, error) {
...
+
+	ghc.workflowRuns = workflowRuns
+
+	return workflowRuns, nil
+}
+
+func (ghc *GitHubContext) getWorkflowRuns() (map[string]string, error) {
+	// TODO: We could use the REST API to get the path, once
+	// https://github.com/google/go-github/commit/7665f317d3985efe6a7828e3af2751c41b3526b1
+	// is in a release.
```
bluekeyes: Looking at the APIs, I think the REST API makes this more straightforward, so it might be worth copying the relevant function and the partial struct we need in to this project while waiting for a release.

https://github.com/palantir/policy-bot/blob/67b611ab74ebfd63f98addb6105e06a71e65bb7f/pull/github_teams.go is an example of where we did something similar previously when an API we still used was removed upstream.

pull/github.go
```diff
@@ -829,6 +834,89 @@ func (ghc *GitHubContext) getCheckStatuses() (map[string]string, error) {
...
+										}
+									}
+									PageInfo *v4PageInfo
+								} `graphql:"checkSuites(first: 100, after: $checkSuiteCursor, filterBy: {appId: $appId})"`
+							}
+						}
+					}
+				} `graphql:"commits(last: 1)"`
+			} `graphql:"pullRequest(number: $number)"`
+		} `graphql:"repository(owner: $owner, name: $repo)"`
```
bluekeyes: If we stick with GraphQL, I think this query can be simplified by using the `object` query on `Repository` to directly get the head commit by SHA. This is also important for consistency - we want to make sure the commit we think we're evaluating is the one we actually get the check suites from.

policy/predicate/workflow.go
```diff
@@ -0,0 +1,95 @@
...
+			missingResults = append(missingResults, workflow)
+		}
+		if _, allowed := conclusions[result]; !allowed {
+			failingWorkflows = append(failingWorkflows, workflow)
+		}
+	}
+
+	if len(missingResults) > 0 {
+		predicateResult.Values = missingResults
+		predicateResult.Description = "One or more workflow runs are  missing: " + strings.Join(missingResults, ", ")
```
bluekeyes:
```suggestion
		predicateResult.Description = "One or more workflow runs are missing: " + strings.Join(missingResults, ", ")
```

README.md
```diff
@@ -329,6 +329,16 @@ if:
...
 
+  # "has_workflow_result" is satisfied if the GitHub Actions workflow runs that
+  # are specified all finished and concluded with one of the conclusions
+  # specified. "conclusions" is optional and defaults to ["success"].
+  # `workflows` contains the paths to the workflow files that are being checked.
+  has_workflow_result:
+    conclusions: ["success", "skipped"]
+    workflows:
+      - ".github/workflows/a.yml"
+      - ".github/workflows/b.yml"
```
bluekeyes, line 354: Since workflows must exist in the `.github/workflows` directory, would it make sense to exclude that from these paths? Or do you think it is clearer to specify the full path?

bluekeyes (commented):
Thanks for proposing this. I haven't used Actions enough to run into this problem, but it seems reasonable to have an Actions-specific predicate to simplify things.

Since this contains some overlap with #789, my comment from there about the `allowedConclusions` set also applies here. I think we'll want to pick one of these PRs to focus on and merge first, and then update the other one to resolve conflicts.

iainlane:
Cheers. I'll mark this as a draft until #789 is good to go, and then pivot over here, so we don't have too much in flight at once.

pull/github.go
bluekeyes: Since I left this comment, we upgraded to `go-github` v63, which should include the new REST API field you wanted to use originally.

iainlane: Just coming back to this. Yep, great! I'll switch over to REST. Cheers.

pull/github.go
iainlane: Ah, cheers for the info. If we use the `slug` field - which is `github-actions` for github.com - would that work in your environment too?

bluekeyes: Yeah, when I checked before it looked like the slug was consistent across both github.com and GitHub Enterprise

iainlane: [The API wants an ID though](https://pkg.go.dev/github.com/google/go-github/v63/github#ListCheckSuiteOptions), so I'm going to have to resolve it. Since that's an API call and the value isn't going to change, I should cache it for the lifetime of the whole program. Do you have a preference for where/how we should do that?

bluekeyes: Not sure if this will do everything else you needed, but if we switch to the [List workflow runs](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository) REST API, I think it bypasses this problem since we know it's only returning workflow details from actions and it includes the paths and the conclusions.

If you do end up needing to cache the ID, I think you could do it at server startup and pass it in to the base handler. Or maybe we make it a server configuration option at that point and operators will have to look it up before using this feature.

README.md
iainlane: I had a look at both, and for me having the full path makes it clearer. I find

```
- a.yml
- b.yml
```

to be a tiny bit more obscure. It's harder to get confused about which files we're talking about if it's a full path.

But it doesn't matter _that much_. WDYT?

pull/github.go
iainlane: Cheers. I'm just checking on a real installation but I think the REST approach will work.

pull/github.go
```diff
@@ -829,6 +830,68 @@ func (ghc *GitHubContext) getCheckStatuses() (map[string]string, error) {
...
+		if resp.NextPage == 0 {
+			break
+		}
+		opt.Page = resp.NextPage
+	}
+	workflowRuns := make(map[string]string, len(runsWithDate))
+
+	for path, run := range runsWithDate {
+		workflowRuns[path] = run.GetConclusion()
+	}
```
iainlane: I think this is not quite right. Imagine a workflow like this

```yaml
on:
  push:
  pull_request:
```

and one of those fails, one succeeds.

With the way I have it now, we would pass the policy if the success happens later than the failure.

I'm thinking we should do the "what came later" check, but group by event. So in the above case we'd return `["success", "failure"]` and the policy would check both of those against its list of allowed conclusions.

iainlane: That would be more aligned with what GitHub does. I constructed the situation in a test repo with a GitHub required status on `build b`. The `push` event fails and the `pull_request` one passes. The PR is blocked.

I'll do a change along the lines of ☝️ 

image

iainlane: Ran out of time to test this live today but I think the current version is more or less what I had in mind. I'll open up for review!

iainlane: Ok, I've now built an image with this change (and my other PRs) included and it seems to do what I expect

image

`policy-bot: main` is running the previous version of this change as per the first screenshot. `dev-policy-bot` is running this PR's code as of now.

In this PR, the successful `pull_request` run for "Build directory B" came _after_ the failed `push` event run. In the previous version, since that came later, we considered it a success. Now, with the current version, we consider them separately since they are triggered by different event types, and the PR is blocked.

When the `push` workflow passes, we give a check as expected:

image

README.md
bluekeyes: Using full paths sounds fine to me then

server/handler/workflow_run.go
```diff
@@ -0,0 +1,71 @@
...
+	"encoding/json"
+
+	"github.com/google/go-github/v63/github"
+	"github.com/palantir/go-githubapp/githubapp"
+	"github.com/palantir/policy-bot/policy/common"
+	"github.com/palantir/policy-bot/pull"
+	"github.com/pkg/errors"
+)
+
+type WorkflowRun struct {
```
bluekeyes, line 28: You'll also need to register this new handler with the `EventDispatcher` so it will actually get the events: https://github.com/palantir/policy-bot/blob/develop/server/server.go#L184-L192

bluekeyes (commented):
Looks good, thanks for adding tests for both the predicate and the `pull.Context` method. I think you missed registering the new event handler with the server, but should be good to merge once that's in place.

server/handler/workflow_run.go
iainlane: Oh yeah, thanks! [Here it is](https://github.com/palantir/policy-bot/compare/fd32004717210a737936d36f1039745bbeef1484..af8fa99485a8f8062b5ef9bdc0db05b6675489d2).

I suppose it mostly worked for us because you tend to get `check_run` events anyway...

bluekeyes (approved):

PR #796, opened 2024-07-09
feat(policy): Support marshalling of config

iainlane:
Sometimes it's convenient to generate a Policy Bot config. For example, when handling required but optional workflows, you need the config file to contain a rule for each workflow which replicates its path filters.

That job is made much easier if the Policy Bot types can be reused. But without `omitempty` the generated YAML contains all fields and it's hard to read/review.

Add `omitempty` where it's needed, and `IsZero` in a few places too - which is how the behaviour of `omitempty` can be customised.

### DIFF
--- a/policy/common/actor.go
+++ b/policy/common/actor.go
@@ -26,21 +26,21 @@ import (
 // team and organization memberships. The set of allowed actors is the union of
 // all conditions in this structure.
 type Actors struct {
-	Users         []string `yaml:"users" json:"users"`
-	Teams         []string `yaml:"teams" json:"teams"`
-	Organizations []string `yaml:"organizations" json:"organizations"`
+	Users         []string `yaml:"users,omitempty" json:"users"`
+	Teams         []string `yaml:"teams,omitempty" json:"teams"`
+	Organizations []string `yaml:"organizations,omitempty" json:"organizations"`
 
 	// Deprecated: use Permissions with "admin" or "write"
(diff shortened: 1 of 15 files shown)


palantirtech:

policy/common/actor.go
```diff
@@ -45,6 +45,10 @@ func (a *Actors) IsEmpty() bool {
 		len(a.Permissions) == 0 && !a.Admins && !a.WriteCollaborators)
 }
 
+func (a *Actors) IsZero() bool {
+	return a.IsEmpty()
+}
```
bluekeyes: I think it would also be fine to rename the `IsEmpty` method to `IsZero` here

bluekeyes (commented):
Thanks, I think it's fine to support this for people who'd like to use it. 

Note that we don't offer any guarantees about Go-level compatibility between different versions of Policy Bot. As long as the behavior of the running server is compatible, we may make breaking changes to any of the Go packages in minor/patch versions. Something to be aware of if you plan to consume this code in a tool to generate policies.

policy/common/actor.go
iainlane: Got it, cheers!

iainlane:
Just remembered about this / came back to it and rebased 👍

iainlane:
@bluekeyes would you be able to take another look at this one please? Have just rebased again.

bluekeyes (approved):
Thanks, sorry I missed your update back in September

PR #808, opened 2024-08-01
Don't look up approval candidates when no approval is required

iainlane:
We always check the approval candidates when evaluating an approval policy. This process ends up making an API call to list the comments on the PR, so that the approvers given in the policy can be checked against the commenters or reviewers on the PR and only the relevant people considered.

This search currently happens regardless of whether any approvals are required. If there aren't any then we can save the API calls and a bit of time by skipping the search. PRs can be evaluated fairly frequently depending on the policy, so this can add up to a lot of API calls.

I noticed this because we had a problem yesterday where we somehow managed to run out of GitHub API quota which I didn't think would be reasonably possible 😱 . This query stood out when I checked the logs, but I'm not sure it's the end of the story.

Hopefully I'm not missing a reason why this was the way it was. :-)

### DIFF
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -304,7 +304,7 @@ func (r *Rule) isApprovedByConditions(ctx context.Context, prctx pull.Context) (
 // FilteredCandidates returns the potential approval candidates and any
 // candidates that should be dimissed due to rule options.
 func (r *Rule) FilteredCandidates(ctx context.Context, prctx pull.Context) ([]*common.Candidate, []*common.Dismissal, error) {
-	if r.Requires.Count == 0 {
+	if r.Requires.Count <= 0 {
 		return nil, nil, nil
 	}
 
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -304,6 +304,10 @@ func (r *Rule) isApprovedByConditions(ctx context.Context, prctx pull.Context) (
 // FilteredCandidates returns the potential approval candidates and any
 // candidates that should be dimissed due to rule options.
 func (r *Rule) FilteredCandidates(ctx context.Context, prctx pull.Context) ([]*common.Candidate, []*common.Dismissal, error) {
+	if r.Requires.Count == 0 {
+		return nil, nil, nil
+	}
+
 	candidates, err := r.Options.GetMethods().Candidates(ctx, prctx)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "failed to get approval candidates")
--- a/policy/approval/approve_test.go
+++ b/policy/approval/approve_test.go
@@ -194,23 +194,6 @@ func TestIsApproved(t *testing.T) {
 		assertPending(t, prctx, r, "0/1 required approvals. Ignored 7 approvals from disqualified users")
 	})
 
-	t.Run("singleApprovalError", func(t *testing.T) {
-		prctx := basePullContext()
-		// The converse of the check in `noApprovalRequired`. If there are
-		// approvers required, we do expect to call `Comments()`, and therefore
-		// this error should be returned to show that it was called.
-		prctx.CommentsError = errors.New("Comments() was called")
-
-		r := &Rule{
-			Requires: Requires{
-				Count: 1,
-			},
-		}
-
-		_, _, err := r.FilteredCandidates(ctx, prctx)
-		require.Error(t, err)
-	})
-
 	t.Run("authorCannotApprove", func(t *testing.T) {
 		prctx := basePullContext()
 		r := &Rule{
--- a/policy/approval/approve_test.go
+++ b/policy/approval/approve_test.go
@@ -16,6 +16,7 @@ package approval
 
 import (
 	"context"
+	"errors"
 	"os"
 	"regexp"
 	"testing"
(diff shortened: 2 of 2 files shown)


bluekeyes:
> I noticed this because we had a problem yesterday where we somehow managed to run out of GitHub API quota which I didn't think would be reasonably possible 😱 . This query stood out when I checked the logs, but I'm not sure it's the end of the story.

Policy Bot can be pretty request-heavy. I think we've done most of the big optimizations, but I'm sure there's more stuff like this that we've missed.

Another note is that GitHub App rate limits scale based on users and repositories in the organization (up to 12,500 requests/hour.) If you work primarily in a monorepo, you may have a lower rate limit than you would if an equivalent level of development was spread across multiple repositories. GitHub Support may be able to make a custom adjustment to your rate limit in that case.

policy/approval/approve.go
```diff
@@ -304,6 +304,10 @@ func (r *Rule) isApprovedByConditions(ctx context.Context, prctx pull.Context) (
 // FilteredCandidates returns the potential approval candidates and any
 // candidates that should be dimissed due to rule options.
 func (r *Rule) FilteredCandidates(ctx context.Context, prctx pull.Context) ([]*common.Candidate, []*common.Dismissal, error) {
+	if r.Requires.Count == 0 {
```
bluekeyes: nit: for consistency with the check in `IsApproved`, I'd suggest using `<= 0` here

policy/approval/approve_test.go
```diff
@@ -188,6 +194,23 @@ func TestIsApproved(t *testing.T) {
 		assertPending(t, prctx, r, "0/1 required approvals. Ignored 7 approvals from disqualified users")
 	})
 
+	t.Run("singleApprovalError", func(t *testing.T) {
+		prctx := basePullContext()
+		// The converse of the check in `noApprovalRequired`. If there are
+		// approvers required, we do expect to call `Comments()`, and therefore
+		// this error should be returned to show that it was called.
+		prctx.CommentsError = errors.New("Comments() was called")
```
bluekeyes: I don't mind having a specific test for this, but I think the fact that rules with required approvals call `Comments()` should be covered by the other approval tests, right?

bluekeyes (commented):
We do check this in `IsApproved`, so I think what happened is that when we did some refactoring for dismissals, we ended up moving the candidate lookup to before that check happened. It's kind of unfortunate that we're doing it in two places now, but I'm not sure how to easily combine them.

policy/approval/approve_test.go
iainlane: That makes sense. I added it initially while developing so it'd go red if I made a mistake. It doesn't need to be committed. caa985cfeabaa10247c5c34f8e2f38d3a4fcccc4

policy/approval/approve.go
iainlane: Sure! 0b558a3b9d81c0b941b7faaec147e772ab552423

iainlane:
> > I noticed this because we had a problem yesterday where we somehow managed to run out of GitHub API quota which I didn't think would be reasonably possible 😱 . This query stood out when I checked the logs, but I'm not sure it's the end of the story.
> 
> Policy Bot can be pretty request-heavy. I think we've done most of the big optimizations, but I'm sure there's more stuff like this that we've missed.

One other one that I've noticed we could do is to allow the default repository to be disabled. It seems like 404s aren't cached and so we do a lot of requests to check for `grafana/.github` having a policy and find it doesn't every time. I'll send over a PR for that shortly. (edit: this was merged in https://github.com/palantir/policy-bot/commit/550e0326d417032a6a7ca64010e592913e0f3cda - thanks!)
 
> Another note is that GitHub App rate limits scale based on users and repositories in the organization (up to 12,500 requests/hour.) If you work primarily in a monorepo, you may have a lower rate limit than you would if an equivalent level of development was spread across multiple repositories. GitHub Support may be able to make a custom adjustment to your rate limit in that case.

Cheers for the tip. For us, it's not _only_ one monorepo that drives activity. It is one of those but also a few of the larger OSS repos we have.

Looking into what happened a bit more, I noticed a few things:

- As reported by the Prometheus metrics from the app, there are two rate limits in play: one with our installation ID of 15,000 (we are on GitHub Enterprise Cloud) and another with no installation ID (`0`) of 12,500
- It's that second one which we exhausted. The installation's rate limit did get used up proportionally but it bottomed out at 5,300 that hour. Bear in mind we were erroring so we probably would have made more requests if the other one hadn't been hit.
- I'm not sure what that 12,500 rate limit is, actually! [The docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28) don't mention a rate limit for non-installation (JWT) requests but perhaps it is that.  Maybe this could help me focus where I'm looking in the code and logs? Just from a quick look that might mean the installations service.
- This happening is fairly anomalous. Here's a screenshot of the time when this happened. I can see one or two other similar instances if I look back further, but never this steep. I didn't manage to pin down a cause yet, but I would suspect it's due to a particular pattern of activity in the org's repos.

image

I'm sending in fixes for what I can see and then continuing to observe the effects they have!

Just a brain dump 🙂

bluekeyes (approved):

bluekeyes:
The requests with no installation ID are interesting. It could be application requests using the JWT, but I don't think we're making many of those. If you have debug logging enabled, it might be helpful to look at the logs with the `github_request` message during that period to see if they also show the requests with no installation ID.

It's also possible we're not correctly attaching the installation ID to the metrics in some situations, but the fact that the maximum limit is different from your actual installation makes that seem less likely.

PR #812, opened 2024-08-02
server: Allow the shared location to be disabled

iainlane:
The shared location can't currently be switched off. There are a couple of reasons why being able to do is desirable:

1. When Policy Bot is run for a repo which doesn't have a config, the shared location is searched for every time, since 404s aren't cached.  If we know it doesn't exist, we can skip that lookup. This saves API calls.
2. Somebody not on the Policy Bot operating team might create a `.github` repository at a later date and put a `.policy.yml` file in there. This file would then start to be used without the team knowing about it. It should be possible to prevent this.

Here we solve this by turning the shared repo and path into pointers in the struct. Then we can tell the difference between if it was explicitly set to `""` or not specified by the administrator. In the former case, we preserve the empty string, which is treated as disabling by `go-githubapp`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,9 @@ instance of the server.
 - If the file does not exist in the repository, `policy-bot` tries to load a
   shared `policy.yml` file at the root of the `.github` repository in the same
   organization. You can change this path and repository name when running your
-  own instance of the server.
+  own instance of the server, or disable this feature by setting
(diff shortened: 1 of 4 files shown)


bluekeyes (approved):
Thanks, this makes sense to reduce requests in installations that don't want to use this feature.

PR #861, opened 2024-10-09
Don't look at workflow run results in other repositories

iainlane:
This is a copy of 536d20206c069699863acb75054283725b023c45. The same issue exists with the `workflow_run` event as with `check_run`. People can fork a repository and then propose pull requests from the original repository into their fork. This may result in webhooks being dispatched, but we want to ignore them.

I did have a quick think about whether there was somewhere common to put this, since it's in two places now, but I couldn't see anywhere decent. I'd still be happy to move this, though, if there is a better place.


bluekeyes (approved):
Thanks, I think it's fine to duplicate this logic in the handlers

PR #866, opened 2024-10-23
Add dismissed state

alankpatel:
Previously we skipping the check if the state was dismissed but after adding dismissed as one of the base states it will rerun the policy if we receive that event.


server/handler/pull_request_review.go
```diff
@@ -88,6 +88,9 @@ func (h *PullRequestReview) Handle(ctx context.Context, eventType, deliveryID st
 
 func (h *PullRequestReview) affectsApproval(reviewState pull.ReviewState, config *policy.Config) bool {
 	states := make(map[pull.ReviewState]struct{})
+	// Include dismissed state
```
bluekeyes: If we're going to comment here, let's explain why we need to include the dismissed state, maybe something like:

> Always process events for dismissed reviews because they can revert the overall approval or disapproval to a previous state

bluekeyes (approved):

PR #875, opened 2024-11-09
check PR changed files count directly from get pull request endpoint

dblinkhorn:
Implements a fix for the checking of the files changed count for a PR. Now the check will be against the git pull request endpoint `changed_files` property rather than the list of files returned from the list pull request files endpoint.


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @dblinkhorn! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

pull/github.go
```diff
@@ -318,12 +328,23 @@ func (ghc *GitHubContext) ChangedFiles() ([]*File, error) {
...
 	}
-	if len(ghc.files) >= MaxPullRequestFiles {
-		return nil, errors.Errorf("too many files in pull request, maximum is %d", MaxPullRequestFiles)
-	}
+
 	return ghc.files, nil
 }
 
+func (ghc *GitHubContext) ChangedFilesCount() (int, error) {
+	pullRequest, _, err := ghc.client.PullRequests.Get(ghc.ctx, ghc.owner, ghc.repo, ghc.number)
```
bluekeyes: I shouldn't be necessary to add a new method to get the changed files from the pull request object because we load the pull request as part of creating the `GitHubContext` object. Instead, I think we need to extend the stored object to include this field:

1. Modify the `v4PullRequestStruct` to include a new `ChangedFiles` field
2. Update `Locator.IsComplete` to return false if the `ChangedFiles` field is missing or empty
3. Update `Locator.toV4` to se the new `ChangedFiles` field from the `github.PullRequest` object in the `Locator`

After that, you should be able to use `ghc.pr.ChangedFiles` to reference this in the `ChangedFiles` method

bluekeyes (changes requested):
Thanks for submitting this! I think we can simplify the implementation by taking advantage of the fact that we already load the pull request object when creating the `pull.Context`

pull/github.go
dblinkhorn: Thanks for the suggestions on the simpler implementation. I pushed a new commit with these changes.

pull/context.go
```diff
@@ -255,4 +255,5 @@ type Body struct {
 	CreatedAt    time.Time
 	Author       string
 	LastEditedAt time.Time
+	ChangedFiles int
```
bluekeyes: Since it looks like nothing uses this field after calling the `Body()` method, I think should remove it for now and only add it in if it is required at some later point. The same thing would apply to the modification to the `v4PullRequestWithEditedAt` struct.

dblinkhorn: Thanks for pointing this out. I have removed both fields from these structs.

bluekeyes (approved):

bluekeyes:
Thanks for the fix! This should be available in the `palantirtechnologies/policy-bot:snapshot` image in a few minutes for testing and I'll look at making a release later today or tomorrow.

PR #885, opened 2024-12-09
add the missing method field for disapproval

### LINKED ISSUE #884: Viewing details of a disapproved PR generates a panic
Visiting the details page for a disapproved PR renders a partial page and an internal server error caused by a panic:

```
template: details.html.tmpl:187:20: executing "result-methods-details" at <getMethods .>: error calling getMethods: runtime error: invalid memory address or nil pointer dereference
```

Not sure exactly what is wrong yet, but there must be some property we set on results from approval rules that we do not set on results from disapproval rules.

admin822:
Before this change : 
Screenshot 2024-12-09 at 3 45 21 PM

After this change : 
Screenshot 2024-12-09 at 3 44 10 PM

Fixes #884.

### DIFF
--- a/config/policy-examples/team-with-disapproval.yml
+++ b/config/policy-examples/team-with-disapproval.yml
@@ -8,7 +8,7 @@ policy:
 
   disapproval:
     requires:
-      organization:
+      organizations:
       - "palantir"
 
 approval_rules:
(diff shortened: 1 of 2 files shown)


config/policy-examples/team-with-disapproval.yml
```diff
@@ -8,7 +8,7 @@ policy:
 
   disapproval:
     requires:
-      organization:
+      organizations:
```
admin822, line 11: Housekeeping, not related to the bug we are fixing in this PR.

bluekeyes (approved):

PR #887, opened 2024-12-12
Add retries when loading config

### LINKED ISSUE #506: Improve error behavior when loading policies
As reported in #405, Policy Bot will sometimes fail to discover a policy file due to a GitHub timeout (or other network error.) When this happens, we post a failed status check on the PR. This is undesirable for repositories that otherwise don't use Policy Bot, especially if the timeout or error happened while checking if there was an organization-level policy.

To make the issue worse, the failed status is never removed because when everything works correctly, the repository is ignored.

I think there are a few possible improvements:

1. Detect and retry timeout errors while loading policies, if we do not already
2. Make the timeout or retries configurable, if they are not already
3. Log errors from loading policies and only post failed status checks for errors that happens after we've read a policy file from GitHub. This requires that repositories expecting a Policy Bot status mark it as required. This could also create an opposite problem, where the Policy Bot status is missing from a PR where it is required and it is hard to find out why.

nikola-jokic:
This PR adds a retry mechanism with exponential backoff when fetching config returns a timeout error. 

Fixes #506

### DIFF
--- a/server/handler/fetcher.go
+++ b/server/handler/fetcher.go
@@ -16,6 +16,8 @@ package handler
 
 import (
 	"context"
+	"os"
+	"time"
 
 	"github.com/google/go-github/v66/github"
 	"github.com/palantir/go-githubapp/appconfig"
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @nikola-jokic! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

server/handler/fetcher.go
```diff
@@ -37,26 +39,50 @@ type ConfigFetcher struct {
...
+				fc.LoadError = err
+				return fc
+			}
 
-	var pc policy.Config
-	if err := yaml.UnmarshalStrict(c.Content, &pc); err != nil {
-		fc.ParseError = err
-	} else {
-		fc.Config = &pc
+			ticker.Reset(delay)
```
bluekeyes: Let's use `time.After` here instead of manipulating a ticker - I think it will be easier to understand and I don't think there are any performance concerns with allocating a new timer if we're about to sleep for multiple seconds. Go 1.23 also fixed the issue with timers not being GC'd until after they fire.

server/handler/fetcher.go
```diff
@@ -37,26 +39,50 @@ type ConfigFetcher struct {
...
+			Path:   c.Path,
+		}
 
-	c, err := cf.Loader.LoadConfig(ctx, client, owner, repository, branch)
-	fc := FetchedConfig{
-		Source: c.Source,
-		Path:   c.Path,
-	}
+		if err != nil {
+			if !os.IsTimeout(err) {
```
bluekeyes: I think it would be worth re-trying other errors too, for instance if GitHub returns a 500 error or something.

bluekeyes (commented):
Thanks for submitting this - had a few suggestions, but overall I think this approach looks good.

server/handler/fetcher.go
nikola-jokic: Sounds good, I guess I got used to using ticker before 1.23 :relaxed:

server/handler/fetcher.go
nikola-jokic: For this one, I guess the change needs to be done in [palantir/go-githubapp](https://github.com/palantir/go-githubapp) since [getFileContents](https://github.com/palantir/go-githubapp/blob/f40ba9b0944c4b7252c5a64fdea7f6565b9f3704/appconfig/appconfig.go#L270C17-L270C18) receives a response and only propagates the error. In order to check for 500, there must either be an error type to check for on the caller side, or some other way to propagate 500 error somehow.

bluekeyes: The `go-github` library returns an error for any non-successful status code, so we have two options here without changing the upstream:

* Retry on any error from loading the content and accept that maybe that will sometimes retry things that will still fail
* Use `errors.As` to test for [github.ErrorResponse](https://pkg.go.dev/github.com/google/go-github/v67/github#ErrorResponse) and then check the status code in the response

nikola-jokic: Updated with error check.
lmkwyt :relaxed:

bluekeyes (approved):
Thanks!

Issue #921, opened 2025-02-25
Inconsistent naming of Github Status Check related predicates


FloThinksPi:
According to how github names their functionality according to their [doc]( https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks)

The current predicates:
```yaml
has_status:
...
has_workflow_result:
...
```

do not really fit into the naming from github.
For github there are [`checks`](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#checks) which have an attribute called [`status`](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions) that can have multiple values. In case the `status` value is `completed`, a `check` has an additional attribute called [`conclusion`](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#check-statuses-and-conclusions) which gives more information over the termination reason.

So i would see some config similar to following example be more understandable as it uses the same terminology and hierarchy.

```yaml
has_checks:
 status: ["completed"] # By default contains all possible statuses
 conclusion: ["success", "skipped"] # default, logical and with status in case status contains "completed"
 checks:
      - "check-name-1"
      - "check-name-2"
      - "check-regex-name-.*"
has_workflows:
 status: ["completed"]  # By default contains all possible statuses
 conclusion: ["success", "skipped"] # default, logical and with status in case status contains "completed"
 workflows:
      - ".github/workflows/a.yml"
      - ".github/workflows/b.yml"
```

Which would i suppose result in the same, current behaviour yet giving a lot more options.

It would btw also allow to put a `has_check` predicate into the `if condition` of a policybot rule allowing e.g. to enforce a rule if a `check` exists with any `state` even when not having a conclusion yet and skipping it if there is no check at all by the if condition of a rule. 
I currently am fiddling around tying to implement support for that scenario where i want to require a check/workflow having certain conclusions but only if the check/workflow actually was triggered/created and im struggling to add this into the current naming schema of `has_workflow_results` as well as `has_status`.

Whats the opinions regarding the naming of these predicates ?

bluekeyes:
[...]

PR #922, opened 2025-02-26
Enable status/workflow predicates for use in if condition

FloThinksPi:
This change is a proposal for the use-cases described in #921

It introduces new predicates `status_check` and `workflow` that compared to the currently existing `status` and `workflow` introduce the capability to specify resources by regular expression as well as defining the desired `status` workflows and status_checks should be in. Additionally the changes in naming of the GitHub documentation([1](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks),[2](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28)) is applied to these  new predicates.

Also a status and conclusion `any` is introduced that will populate the statuses/conclusions with all possible values.

This now allows for more scenarios where a status check or workflow may be required in a rule but only if it exists.
```
policy:
  approval:
    - and:
      - Test Dependencies
      - All started workflow concluded(at least one)


approval_rules:
  - name: Test Dependencies
    description: Test Dependencies Checks Passed
    options:
      invalidate_on_push: true
    if:
      has_workflow:
        statuses: ["any"]
        conclusions: ["any"]
        workflows:
          - '\.github/workflows/test-dependencies\.yml'
    requires:
      conditions:
        has_workflow:
          conclusions: ["success", "skipped"]
          workflows:
            - '\.github/workflows/test-dependencies\.yml'
  - name: All started workflow concluded(at least one)
    description: All started workflow concluded(at least one)
    options:
      invalidate_on_push: true
    requires:
      conditions:
        has_workflow:
          conclusions: ["any"]
          workflows:
            - '.*'
```
Hereby the rule `All started workflow concluded(at least one)` makes sure all started workflows finished but at least one workflow needs to complete to fullfill the rule. This is to ensure the rules all does not get approved before all workflows are properly created in the api and can be used for rule evaluation. This assumes that it is "unlikely" for a workflow to be scheduled by github actions after a workflow completely finished timing wise.
The rule `Test Dependencies` requires the workflow `test-dependencies.yml`(attention regex so this will match also `random-test-dependencies.yml`) to be present with any status or conclusion and if so it must finish with conclusion `success` or `skipped`. The usual regex escape mechanism can be applied here.

This allows now to require certain status checks, repo statuses and workflows to be in a certain state/conclusion *if they exist at all*, wich with the current predicates is not possible to cover efficiently. Currenlty the only workaround is to define the running condition twice with `only_changed_files` predicate which does not work reliably with good user experience. Also for status checks it might not be able to mimic the running condition in policy-bot of arbitrary statuses that could possibly have complex logic when they appear and when not. 

Hereby the predicates `has_workflow_result`,`has_status` and `has_successful_status` remain identical and are just marked as deprecated. Internally they also use the meta predicate `has_status_check` and `has_workflow` so no code duplication exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,7 +372,7 @@ if:
     deletions: "> 100"
     total: "> 200"
 
-  # DEPRECATED: Use "has_status" below instead, which is more flexible.
+  # DEPRECATED: Use "has_status_check" below instead, which is more flexible.
   # "has_successful_status" is satisfied if the status checks that are specified
   # are marked successful on the head commit of the pull request.
   has_successful_status:
@@ -380,6 +380,7 @@ if:
     - "status-name-2"
     - "status-name-3"
 
+  # DEPRECATED: Use "has_status_check" below instead, which is more flexible.
   # "has_status" is satisfied if the status checks that are specified are
   # finished and concluded with one of the conclusions specified.
   # "conclusions" is optional and defaults to ["success"].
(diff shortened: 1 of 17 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @FloThinksPi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #934, opened 2025-03-11
add file_exists predicate

### LINKED ISSUE #246: feature request: Allow conditional policy on file deletion/additon
Today we allow conditional policies on aspects of file modification like number of lines added/deleted/modified. Certain workflows would benefit from extending this to also allow conditional policies on the addition of new files of deletion of existing files which cannot be accurately expressed in terms of existing file modification knobs.

@asvoboda

dblinkhorn:
This adds the `file_exists` predicate, which accepts a list of regex patterns used verify whether files required by a policy rule have been deleted as part of a PR. If any matching file is removed, the predicate fails, allowing rules that depend on the file's existence to be skipped.

- Added logic for `file_exists` predicate
- Updated unit tests to verify functionality
- Updated readme to document the new predicate

Closes #246

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,17 @@ if:
     paths:
       - "^config/.*$"
 
+  # "file_exists" is satisfied if none of the files matching any regular expression
+  # in the "paths" list have been deleted in the pull request. If any matching file
+  # is removed, the predicate fails, allowing rules that depend on the file's existence
+  # to be skipped.
+  #
+  # Note: Double-quote strings must escape backslashes while single/plain do not.
+  # See the Notes on YAML Syntax section of this README for more information.
+  file_exists:
+    paths:
+      - "^\\.github/workflows/.*\\.ya?ml$"
+
   # "has_author_in" is satisfied if the user who opened the pull request is in
   # the users list or belongs to any of the listed organizations or teams. The
   # `users` field can contain a GitHub App by appending `[bot]` to the end of
--- a/policy/predicate/file.go
+++ b/policy/predicate/file.go
@@ -181,6 +181,60 @@ func (pred *NoChangedFiles) Trigger() common.Trigger {
 	return common.TriggerCommit
 }
 
+type FileExists struct {
+	Paths []common.Regexp `yaml:"paths"`
+}
+
+var _ Predicate = &FileExists{}
+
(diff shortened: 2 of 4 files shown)


bluekeyes:
Thanks for the PR! Sorry for not commenting on #246 before you started this, but I'm not sure `file_exists` is the right predicate for this problem, for two reasons:

1. I think `file_exists` as a name is misleading about what it does. It doesn't actually care if the file exists in the repository or on the PR branch, as long as it is not being deleted by the current PR. I think something like `file_not_deleted` is more accurate to what this does and less likely to give the wrong expectation about policy behavior.
2. It only supports the "not deleted" case and it would be nice to have something (either a feature of this predicate or a parallel predicate) for the related "deleted", "added", and "not added" cases.

Right now, I have two ideas for alternate approaches but open to any suggestions you have too:

* Create `file_deleted`, `file_not_deleted`, `file_added`, and `file_not_added` predicates. Advantage here is that these are all simple and it's clear what they do. Disadvantage is adding four new predicates that are conceptually similar to the existing file predicates.
* Extend `changed_files`, `only_changed_files`, and `no_changed_files` in some way. Maybe the existing `paths` key is interpreted as meaning any modification and we add new `added` and `deleted` keys that take a list of patterns matching added or deleted files? Advantage is that it avoids expanding the number of predicates, but it may make understanding the existing predicates more complicated, especially when multiple of `paths`, `added`, and `deleted` are set.

What do you think?

dblinkhorn:
@bluekeyes thanks for the comments! Your suggestions make a lot of sense. I think I like first option to create separate predicates better as that would possibly be easier to use and reason about. I am happy to implement either solution as they will both solve a pain point for us.

bluekeyes:
I discussed this a bit with the team and the consensus is that we should add the four separate predicates. If you're willing to add all of them, that would be great, otherwise feel free to rename the one you added here to either `file_not_deleted` or `no_file_deleted` and I can add the remaining 3 predicates later on.

dblinkhorn:
> feel free to rename the one you added here to either `file_not_deleted` or `no_file_deleted` and I can add the remaining 3 predicates later on.

I'll start with this and create an issue on adding the others, which I should be able to get to.

PR #935, opened 2025-03-13
add file_not_deleted predicate

### LINKED ISSUE #246: feature request: Allow conditional policy on file deletion/additon

dblinkhorn:
This adds the `file_not_deleted` predicate, which accepts a list of regex patterns used verify whether files required by a policy rule have been deleted as part of a PR. If any matching file is removed, the predicate fails, allowing rules that depend on the file's existence to be skipped.

- Added logic for `file_not_deleted` predicate
- Updated unit tests to verify functionality
- Updated readme to document the new predicate

See discussion from [previous PR](https://github.com/palantir/policy-bot/pull/934).

Closes #246

### DIFF
--- a/policy/predicate/file.go
+++ b/policy/predicate/file.go
@@ -181,6 +181,60 @@ func (pred *NoChangedFiles) Trigger() common.Trigger {
 	return common.TriggerCommit
 }
 
+type FileNotDeleted struct {
+	Paths []common.Regexp `yaml:"paths"`
+}
+
+var _ Predicate = &FileNotDeleted{}
+
+func (pred *FileNotDeleted) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
+	var paths []string
+	for _, r := range pred.Paths {
+		paths = append(paths, r.String())
+	}
+
+	predicateResult := common.PredicateResult{
+		Satisfied:         true,
+		ValuePhrase:       "deleted files",
+		Values:            []string{},
+		ConditionPhrase:   "match path patterns",
+		ReverseSkipPhrase: true,
+		ConditionValues:   paths,
+	}
+
+	changedFiles, err := prctx.ChangedFiles()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list changed files")
+	}
+
+	if len(changedFiles) == 0 {
+		predicateResult.Description = "No files were changed"
+		return &predicateResult, nil
+	}
+
+	deletedFiles := []string{}
+	for _, f := range changedFiles {
+		if f.Status == pull.FileDeleted {
+			deletedFiles = append(deletedFiles, f.Filename)
+
+			if anyMatches(pred.Paths, f.Filename) {
+				predicateResult.Satisfied = false
+				predicateResult.Values = []string{f.Filename}
+				predicateResult.Description = f.Filename + " was deleted"
+				return &predicateResult, nil
+			}
+		}
+	}
+
+	predicateResult.Values = deletedFiles
+	predicateResult.Description = "No deleted files match the specified patterns"
+	return &predicateResult, nil
+}
+
+func (pred *FileNotDeleted) Trigger() common.Trigger {
+	return common.TriggerCommit
+}
+
 type ModifiedLines struct {
 	Additions ComparisonExpr `yaml:"additions"`
 	Deletions ComparisonExpr `yaml:"deletions"`
--- a/README.md
+++ b/README.md
@@ -268,6 +268,17 @@ if:
     paths:
       - "^config/.*$"
 
+  # "file_not_deleted" is satisfied if none of the files matching any regular expression
+  # in the "paths" list have been deleted in the pull request. If any matching file
+  # was deleted, the predicate fails, allowing rules that depend on the file's existence
+  # to be skipped.
+  #
+  # Note: Double-quote strings must escape backslashes while single/plain do not.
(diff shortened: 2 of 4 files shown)


dblinkhorn:
Created #936 to document the additional predicates to be added as part of this.

policy/predicate/file.go
```diff
@@ -181,6 +181,60 @@ func (pred *NoChangedFiles) Trigger() common.Trigger {
...
+
+	if len(changedFiles) == 0 {
+		predicateResult.Description = "No files were changed"
+		return &predicateResult, nil
+	}
+
+	var deletedFiles []string
+	for _, f := range changedFiles {
+		if anyMatches(pred.Paths, f.Filename) {
+			if f.Status == pull.FileDeleted {
```
bluekeyes:
```suggestion
		if f.Status == pull.FileDeleted {
			if anyMatches(pred.Paths, f.Filename) {
```

Let's check the file status first to avoid doing any regex evaluation for files that we already know have no impact.

policy/predicate/file.go
```diff
@@ -181,6 +181,60 @@ func (pred *NoChangedFiles) Trigger() common.Trigger {
...
+	for _, r := range pred.Paths {
+		paths = append(paths, r.String())
+	}
+
+	predicateResult := common.PredicateResult{
+		Satisfied:       true,
+		ValuePhrase:     "file not deleted",
+		Values:          []string{},
+		ConditionPhrase: "all specified paths exist",
+		ConditionValues: paths,
```
bluekeyes:
```suggestion
		Satisfied:         true,
		ValuePhrase:       "deleted files",
		ConditionPhrase:   "match path patterns",
		ReverseSkipPhrase: true,
		ConditionValues:   paths,
```

I think getting the information for the details page right is the hardest part of writing a predicate. In addition to this change, we should collect all of the deleted files to use as the `Values` when the predicate is satisfied. When the predicate is not satisfied (i.e. there's at least one deleted file matching the patterns), it's up to you if you want to set `Values` to only the first matching file and return immediately or if you want to collect all of the matching files. Most of the other file predicates return as soon as they find one matching file, which makes writing the `Description` message a bit easier too.

For context, these values are (roughly) used as:

> "The `{{.ValuesPhrase}}` `{{.Values}}` `{{.ConditionPhrase}}` `{{.ConditionValues}}`"

So for this predicate:

> "The deleted files `.github/workflows/bad.yaml` match path patterns `.github/workflows/.*\\.yaml`"
> "The deleted files `test.go` do not match path patterns `.github/workflows/.*\\.yaml`"

See `server/templates/details.html.tmpl` for the exact usage.

policy/predicate/file.go
```diff
@@ -181,6 +181,60 @@ func (pred *NoChangedFiles) Trigger() common.Trigger {
...
+			if f.Status == pull.FileDeleted {
+				deletedFiles = append(deletedFiles, f.Filename)
+			}
+		}
+	}
+
+	if len(deletedFiles) > 0 {
+		predicateResult.Satisfied = false
+		predicateResult.Values = deletedFiles
+		predicateResult.Description = fmt.Sprintf("Files matching defined patterns were deleted: %v", deletedFiles)
```
bluekeyes: If we exit on the first matching file, let's use `${filename} was deleted` as the description. Otherwise, `A deleted file matches the specified patterns` is good for this. The actual files are available via the other details output, so the description should be brief.

policy/predicate/file.go
```diff
@@ -181,6 +181,60 @@ func (pred *NoChangedFiles) Trigger() common.Trigger {
...
+		return &predicateResult, nil
+	}
+
+	predicateResult.Description = "All specified paths exist"
```
bluekeyes:
```suggestion
	predicateResult.Description = "No deleted files match the specified patterns"
```

bluekeyes (commented):
Thanks for including documentation and tests with this! I had a few suggestions, mostly about how the details message is formatted. Otherwise, the core logic of this predicate seems good.

policy/predicate/file.go
dblinkhorn: Good catch, thanks!

policy/predicate/file.go
dblinkhorn: Thanks for that context, makes a lot more sense now! I went with returning on the first matching deleted file to be consistent with the other file predicates. Also, `predicateResult.Values` will be set to all deleted files, if the PR has non-matching deleted files.

policy/predicate/file.go
dblinkhorn: Wasn't 100% sure on your meaning here so let me know if it's still now how it should be.

policy/predicate/file.go
dblinkhorn: Fixed.

policy/predicate/file.go
bluekeyes: Now that we return on L224 in the `Satisfied = false` case, I don't think we need this block starting on L229 at all. It should be enough to set `predicateResult.Values = deletedFiles`, set the description on L232, and return.

dblinkhorn: Fixed, thanks!

bluekeyes (approved):
Thanks!

PR #974, opened 2025-06-10
Expand Actor to support UserTypes and Not

south-mer:
## What

Expand the capabilities of `Actor` which is used in `requires`, `author`, etc.

## Why

Allow the expressing predicates such as "author must be user but not certain machine accounts" etc.


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @south-mer! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

south-mer:
Hello @asvoboda (apologize in advance for pinging directly because you have reviewed other PRs recently).

Could you please take a look and let us know what you think about potentially merging this PR? Thank you.

asvoboda:
I think it would be useful to describe in more detail what problems you are having first and if they can be expressed using current policy primitives.

south-mer:
@asvoboda thank you for your question.

We are writing rules that specifically target pull requests with authors/contributors that are NHIs like GitHub Apps and machine users. For the prior, we would like to use the `UserTypes`, while for the latter one we would like to use `Not`.

We are also writing some specific rules that further re-include some GitHub Apps, resulting in nested `Not`s, in order to smooth the onboarding of repositories with existing automation onto policy-bot.

PR #975, opened 2025-06-10
Implement ForceSharedPolicy

### LINKED ISSUE #377: [Question]: Prevent changes to policy
As an enterprise, we would like to use same policy for everyone. In order to do that we would need to centralize the policy.yml configuration, which i can see its possible to do it by pointing the policy to remote repo. But its possible to override that by adding policy.yml in the repo. Is there a way to prevent this, can we add some type of flag or setting that would prevent overriding policy that is defined in one central place?

south-mer:
## What

Add `ForceSharedPolicy` to force the use of shared policy for all repositories

## Why

To resolve https://github.com/palantir/policy-bot/issues/377

### DIFF
--- a/server/handler/eval_options.go
+++ b/server/handler/eval_options.go
@@ -32,6 +32,9 @@ type PullEvaluationOptions struct {
 	SharedRepository *string `yaml:"shared_repository"`
 	SharedPolicyPath *string `yaml:"shared_policy_path"`
 
+	// Ignore PolicyPath and use SharedRepository and SharedPolicyPath only.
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @south-mer! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes (approved):
Thanks!

PR #977, opened 2025-06-20
Add option to ignore edited comments on the server

asvoboda:
Add a server side option to configure ignore_edited_comments globally


jamestoyer (dismissed):

server/handler/eval_context.go
```diff
@@ -122,6 +123,11 @@ func (ec *EvalContext) ParseConfig(ctx context.Context, trigger common.Trigger)
 func (ec *EvalContext) EvaluatePolicy(ctx context.Context, evaluator common.Evaluator) (common.Result, error) {
 	logger := zerolog.Ctx(ctx)
 
+	// Apply server-side options to the context
+	if ec.Options != nil {
+		ctx = options.WithIgnoreEditedComments(ctx, ec.Options.IgnoreEditedComments)
+	}
+
 	result := evaluator.Evaluate(ctx, ec.PullContext)
```
asvoboda: Other ideas:
* Put this information into the `PullContext`
* Though I think its more invasive, modify the `Evaluate` interface to add an options param,

bluekeyes: I don't really like passing this through the context - since we fully control the interfaces here, think we should modify things to make this more explicit. My two suggestions:

1. Modify `policy.ParsePolicy` to take some kind of global options parameter. When parsing the policy, iterate through all of the rules and override their values based on the global options. This needs no changes to the evaluation code and should work for any future case where we want a global override for a particular rule option.
2. As you proposed, extend the `Evaluate` method to take either a struct or a variable list of `EvaluateOptions`

I think my preference is for modifying the parsing. That way, a given `policy.Policy` instance always evaluates the same way and doesn't change based on options passed at evaluation time. That also keeps the separation between policy evaluation and otherwise unrelated server-wide concerns.

bluekeyes (approved):

PR #990, opened 2025-07-16
Improve validation API docs

reegnz:
The `curl` uses `PUT` when using the `-T` or `--upload-file` flag, so it's unnecessary to explicitly set the method.

The complex logic using temporary files and shell variables is also not needed to determine if the validation failed. Using the `--fail` flag has curl return with a non-zero exit code if the response has a 400 or greater response code. The `--fail-with-body` prints the body, and in this case it makes most sense, as we're interested in the details of the validation failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -705,15 +705,17 @@ however it cannot validate that the rules are semantically correct for a given u
 The API can be used as such:
 
 ```sh
-$ curl https://policybot.domain/api/validate -XPUT -T path/to/policy.yml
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @reegnz! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes (approved):
Thanks!

PR #1043, opened 2025-09-06
Add file filtering support to modified_lines predicate with include/exclude patterns

thelastnode:
- Support three filtering modes:
  - No files config: count all files
  - Only exclude: count all files except excluded ones
  - Include (with optional exclude): count only included files, minus excluded ones
- Updated README with explanation of rules

This enables policies like:

```yaml
modified_lines:
  additions: "> 100"
  files:
    include:
      - ".*\\.go$"
      - ".*\\.js$"
    exclude:
      - ".*_test\\.go$"
      - ".*\\.generated\\..*"
```

🤖 Generated with Claude Code (https://claude.ai/code)

### DIFF
--- a/policy/predicate/file.go
+++ b/policy/predicate/file.go
@@ -362,6 +362,12 @@ type ModifiedLines struct {
 	Additions ComparisonExpr `yaml:"additions,omitempty"`
 	Deletions ComparisonExpr `yaml:"deletions,omitempty"`
 	Total     ComparisonExpr `yaml:"total,omitempty"`
+	Files     *FilesConfig   `yaml:"files,omitempty"`
+}
+
+type FilesConfig struct {
+	Include []common.Regexp `yaml:"include,omitempty"`
+	Exclude []common.Regexp `yaml:"exclude,omitempty"`
 }
 
 type CompareOp uint8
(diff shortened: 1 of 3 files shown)


policy/predicate/file.go
```diff
@@ -362,6 +362,12 @@ type ModifiedLines struct {
 	Additions ComparisonExpr `yaml:"additions,omitempty"`
 	Deletions ComparisonExpr `yaml:"deletions,omitempty"`
 	Total     ComparisonExpr `yaml:"total,omitempty"`
+	Files     *FilesConfig   `yaml:"files,omitempty"`
```
bluekeyes: To avoid some of the `nil` checks, this could be defined as a value (`FileConfig`). The nested slices are empty by default, so this doesn't change any behavior when the configuration doesn't specify the `files` key.

After making that change, we should also define a `IsZero` method on `FilesConfig` so that `omitempty` still works when marshalling:

```go
func (fc FilesConfig) IsZero() bool {
  return len(fc.Include) == 0 && len(fc.Exclude) == 0
}
```

policy/predicate/file.go
```diff
@@ -469,10 +475,41 @@ func (pred *ModifiedLines) Evaluate(ctx context.Context, prctx pull.Context) (*c
...
+			} else if len(pred.Files.Include) > 0 {
+				if len(pred.Files.Include) == 1 {
+					predicateResult.ConditionPhrase = fmt.Sprintf("meet the modification conditions for files matching %s", getPathStrings(pred.Files.Include)[0])
+				} else {
+					predicateResult.ConditionPhrase = "meet the modification conditions for files matching the included patterns"
+				}
+			} else {
+				predicateResult.ConditionPhrase = "meet the modification conditions excluding specified patterns"
+			}
+		}
```
bluekeyes: Getting the predicate descriptions correct is probably one of the most confusing parts of Policy Bot code. The existing descriptions for this predicate aren't that great, so one acceptable option is to remove all of this and not mention the include/exclude patterns in the description at all.

If you want to include full information about the predicate, we'll need to also modify the descriptions in the cases below. The `ConditionsMap` and `ConditionValues` properties are mutually exclusive, with the map winning if both are set. I think the correct way to do this will be:

1. Set `ConditionPhrase` to `meet`
2. If `Files.Include` is set, set `ConditionsMap["in files matching"] = getPathStrings(pred.Files.Include)`
3. If `Files.Exclude` is set, set `ConditionsMap["excluding files matching"] = getPathStrings(pred.Files.Exclude)`
4. Change the lines that currently set `ConditionValues` to instead set `ConditionsMap["the modification conditions"] = []string{/* existing values */}`

This still isn't perfect (the elements of the condition map may be rendered in an order that doesn't read cleanly), but I think its the best that can be done given how the rest of the system works today. It should also reduce the number of nested if statements here.

bluekeyes (commented):
I think this makes sense as an extension to the `modified_lines` predicate, since I can't think of a way to achieve exactly the same thing by combining existing predicates. The core implementation seems good but I had some comments about the config and the rendering for the details page.

bluekeyes (dismissed):

policy/predicate/file.go
```diff
@@ -469,39 +480,70 @@ func (pred *ModifiedLines) Evaluate(ctx context.Context, prctx pull.Context) (*c
 
 	var additions, deletions int64
 	for _, f := range files {
+		if len(pred.Files.Exclude) > 0 && anyMatches(pred.Files.Exclude, f.Filename) {
+			continue
+		}
+		if len(pred.Files.Include) > 0 && !anyMatches(pred.Files.Include, f.Filename) {
+			continue
+		}
```
alankpatel: nit: could possibly have a function here which also acts as a comment 
`func shouldIncludeFile(filters ModifiedLinesFileFilter, filename string) bool`

alankpatel (dismissed):

alankpatel (approved):

PR #1054, opened 2025-09-17
Custom properties predicate

south-mer:
## Before this PR
Custom properties are not supported in predicates.

## After this PR

Custom properties are supported in predicates, implemented in the way as proposed in https://github.com/palantir/policy-bot/discussions/1029#discussioncomment-14296382.

## Possible downsides?

New predicate are not active unless they are used in the policy, so there should be no negative impact.

## Screenshots
Some screenshots when testing the feature to cover the "testing" of `ConditionPhrase` and `ReverseSkipPhrase` that is not covered by `policy/predicate/custom_properties_test.go`

<details><summary>match</summary>
image
</details>
<details><summary>nomatch</summary>
image
</details>
<details><summary>empty</summary>
image
</details>

### DIFF
--- a/policy/predicate/custom_properties.go
+++ b/policy/predicate/custom_properties.go
@@ -1,0 +1,194 @@
+// Copyright 2025 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package predicate
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"slices"
+	"strings"
+
+	"github.com/palantir/policy-bot/policy/common"
+	"github.com/palantir/policy-bot/pull"
+	"github.com/pkg/errors"
+)
+
+type CustomPropertyIsNull []string
+type CustomPropertyIsNotNull []string
+type CustomPropertyMatchesAnyOf map[string][]string
+type CustomPropertyMatchesNoneOf map[string][]string
+
+var _ Predicate = (CustomPropertyIsNull)(nil)
+var _ Predicate = (CustomPropertyIsNotNull)(nil)
+var _ Predicate = (CustomPropertyMatchesAnyOf)(nil)
+var _ Predicate = (CustomPropertyMatchesNoneOf)(nil)
+
+func formatCustomProperties(customProperties map[string]pull.CustomProperty) []string {
+	result := []string{}
+	keys := make([]string, 0, len(customProperties))
+	for k := range customProperties {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
+	for _, k := range keys {
+		v := customProperties[k]
+		formatted := "(null)"
+		if v.String != nil {
+			formatted = *v.String
+		}
+		if v.Array != nil {
+			formatted = "[" + strings.Join(v.Array, ", ") + "]"
+		}
+		result = append(result, fmt.Sprintf("%s: %s", k, formatted))
+	}
+	return result
+}
+
+func (pred CustomPropertyIsNotNull) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
+	customProperties, err := prctx.RepositoryCustomProperties()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get repository custom properties")
+	}
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "specified custom properties",
+		Values:          formatCustomProperties(customProperties),
+		ConditionPhrase: "have a non-null value",
+		ConditionValues: pred,
+		Satisfied:       true,
+	}
+
+	for _, property := range pred {
+		if _, ok := customProperties[property]; !ok {
+			predicateResult.Satisfied = false
+			return &predicateResult, nil
+		}
+	}
+
+	return &predicateResult, nil
+}
+
+func (pred CustomPropertyIsNull) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
+	customProperties, err := prctx.RepositoryCustomProperties()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get repository custom properties")
+	}
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:       "specified custom properties",
+		Values:            formatCustomProperties(customProperties),
+		ReverseSkipPhrase: true,
+		ConditionPhrase:   "have a non-null value",
+		ConditionValues:   pred,
+		Satisfied:         true,
+	}
+
+	for _, property := range pred {
+		if _, ok := customProperties[property]; ok {
+			predicateResult.Satisfied = false
+			return &predicateResult, nil
+		}
+	}
+
+	return &predicateResult, nil
+}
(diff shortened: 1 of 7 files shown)


south-mer:
@bluekeyes hello, could you let us know if this implementation matches the design as suggested at https://github.com/palantir/policy-bot/discussions/1029#discussioncomment-14296382, and if any other modifications are needed?

bluekeyes:
Sorry, I was unavailable last week, but this is on my list to review this week. Thanks for submitting the PR!

policy/predicate/custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "specified custom properties",
+		Values:          formatCustomProperties(customProperties),
+		ConditionPhrase: "have a non-null value",
+		ConditionValues: pred,
+		Satisfied:       true,
+	}
+
+	for _, property := range pred {
+		if _, ok := customProperties[property]; !ok {
```
bluekeyes, line 72: The documentation for this predicate says it is satisfied if the properties "are set to a non-empty string or array on the target repository", but the code here is checking only if the property exists in the properties map. Does the GitHub API automatically exclude or prevent setting properties that have empty values?

policy/predicate/custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+var _ Predicate = (CustomPropertyMatchesAnyOf)(nil)
+var _ Predicate = (CustomPropertyMatchesNoneOf)(nil)
+
+func formatCustomProperties(customProperties map[string]pull.CustomProperty) []string {
+	result := []string{}
+	keys := make([]string, 0, len(customProperties))
+	for k := range customProperties {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
```
bluekeyes:
```suggestion
	keys := slices.Sorted(maps.Keys(customProperties))
```

policy/predicate/custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get repository custom properties")
+	}
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "specified custom properties",
+		Values:          formatCustomProperties(customProperties),
+		ConditionPhrase: "have a non-null value",
+		ConditionValues: pred,
+		Satisfied:       true,
```
bluekeyes: The current structure for these messages will render a bit awkwardly. I think there are two options to improve it:

1. Use the predicate to populate the values and omit the condition values:
   ```go
   ValuePhrase:     "custom properties",
   Values:          pred,
   ConditionPhrase: "exist",
   ```
   This is the simplest, but does not show the actual properties that are set on the repository.

2. Adjust the messages so that you can show all of the actual properties:
   ```go
   ValuePhrase:     "custom properties",
   Values:          formatCustomProperties(customProperties),
   ConditionPhrase: "contain",
   ConditionValues: pred,
   ```

I prefer the first option. I think the message is clearer and the actual values of the properties are not as important when the predicate check only if they are present or not.

policy/predicate/custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, errors.Wrap(err, "failed to get repository custom properties")
+	}
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:       "specified custom properties",
+		Values:            formatCustomProperties(customProperties),
+		ReverseSkipPhrase: true,
+		ConditionPhrase:   "have a non-null value",
+		ConditionValues:   pred,
+		Satisfied:         true,
```
bluekeyes: This message should also be updated in the same way as the message for `CustomPropertyIsNotNull`, but adding `ReverseSkipPhrase: true`.

policy/predicate/custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		conditionsMap[property] = strValues
+	}
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "specified custom properties",
+		Values:          formatCustomProperties(customProperties),
+		ConditionPhrase: "match one or more of the specified values",
+		ConditionsMap:   conditionsMap,
+		Satisfied:       true,
```
bluekeyes:
```suggestion
		ValuePhrase:     "custom properties",
		Values:          formatCustomProperties(customProperties),
		ConditionPhrase: "match one or more of the patterns",
		ConditionsMap:   conditionsMap,
		Satisfied:       true,
```

policy/predicate/custom_properties.go
```diff
@@ -0,0 +1,189 @@
...
+		conditionsMap[property] = strValues
+	}
+
+	predicateResult := common.PredicateResult{
+		ValuePhrase:       "specified custom properties",
+		Values:            formatCustomProperties(customProperties),
+		ReverseSkipPhrase: true,
+		ConditionPhrase:   "match one or more of the specified values",
+		ConditionsMap:     conditionsMap,
+		Satisfied:         true,
```
bluekeyes:
```suggestion
		ValuePhrase:       "custom properties",
		Values:            formatCustomProperties(customProperties),
		ReverseSkipPhrase: true,
		ConditionPhrase:   "match one or more of the patterns",
		ConditionsMap:     conditionsMap,
		Satisfied:         true,
```

bluekeyes (commented):
Thanks for submitting this! I had a few comments related to how the details messages are produced, but I think the core logic of the predicates all looks good.

south-mer:
Thank you for the review. I will fix the conflict, address the issue and then request another review.

policy/predicate/custom_properties.go
south-mer: Accepted in https://github.com/palantir/policy-bot/pull/1054/commits/ea7bbfccdb8a51a6e2a5010e9eb7d861a1ef2270.

policy/predicate/custom_properties.go
south-mer: Accepted in https://github.com/palantir/policy-bot/pull/1054/commits/ea7bbfccdb8a51a6e2a5010e9eb7d861a1ef2270.

policy/predicate/custom_properties.go
south-mer: Accepted in https://github.com/palantir/policy-bot/pull/1054/commits/ea7bbfccdb8a51a6e2a5010e9eb7d861a1ef2270.

policy/predicate/custom_properties.go
south-mer: I chose the second option in https://github.com/palantir/policy-bot/pull/1054/commits/baf70efc1eb12a5ef3021613d0c652850ef3698e, because it might be a bit confusing to debug without the current values.

If the first option is strongly preferred, I can switch to it.

policy/predicate/custom_properties.go
south-mer: I chose the second option in https://github.com/palantir/policy-bot/pull/1054/commits/baf70efc1eb12a5ef3021613d0c652850ef3698e, because it might be a bit confusing to debug without the current values.

If the first option is strongly preferred, I can switch to it.

policy/predicate/custom_properties.go
south-mer: It is possible to set empty strings and arrays through the [`PATCH /repos/{owner}/{repo}/properties/values`](https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28#create-or-update-custom-property-values-for-a-repository) API, but those are considered unset and not returned in [`GET /repos/{owner}/{repo}/properties/values`](https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28#get-all-custom-property-values-for-a-repository).

I have added a comment in https://github.com/palantir/policy-bot/pull/1054/commits/ec350697d99bb94d3228d96b08acb6b6c24099aa to clarify the behavior.

bluekeyes (approved):
Thanks! We have some other pending features that I need to review and test, but I'll try to make a release with these new predicates soon.

PR #1066, opened 2025-10-06
Review comments should affect approvals

### LINKED ISSUE #1007: Status update issue when sending comments with a review
When reviewing a PR with a comment but without approving or requesting changes, policy-bot seems to ignore the event, and doesn't evaluate the policy. I have verified that Github is sending the event.

After enabling debug logging, we noticed it is hitting the following line: https://github.com/palantir/policy-bot/blob/v1.38.2/server/handler/pull_request_review.go#L76

**Reproduction**

The following config should be enough for replicating the issue.

Example config:
```
policy:
  approval:
  - Review
approval_rules:
- name: Review
  requires:
    count: 1
    teams:
    - "org/review-team"
  options:
    allow_contributor: true
    invalidate_on_push: true
    ignore_update_merges: true
    methods:
      comments: []
      comment_patterns:
      - "^:unicorn:"
      - "^🦄"
      github_review_comment_patterns:
      - "^:unicorn:"
      - "^🦄"
  ```

Send a 🦄 via review comment as in the screenshot below.

Image

The status check isn't updated.

Going to the link of the status check forces re-evaluation, and the status goes green.

psalaberria002:
## Before this PR
Fixes https://github.com/palantir/policy-bot/issues/1007

## After this PR
PR review comments will trigger a re-evaluation of the policy if github_review_comment_patterns is set on a policy.

## Possible downsides?
Can't think of any.

### DIFF
--- a/server/handler/pull_request_review.go
+++ b/server/handler/pull_request_review.go
@@ -94,10 +94,18 @@ func (h *PullRequestReview) affectsApproval(reviewState pull.ReviewState, config
 
 	for _, rule := range config.ApprovalRules {
 		states[rule.Options.GetMethods().GithubReviewState] = struct{}{}
+		// Process review comments if review comment patterns are configured
+		if len(rule.Options.GetMethods().GithubReviewCommentPatterns) > 0 {
+			states[pull.ReviewCommented] = struct{}{}
+		}
 	}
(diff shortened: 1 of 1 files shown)


bluekeyes:
Per discussion on the issue, closing this in favor of #1073. Thanks for raising this and submitting a potential fix!

Issue #41, opened 2018-12-13
Support rule defaults for common options


bluekeyes:
It would be helpful if policy-bot support a way to set default options for all rules in a policy without copying code or using YAML references. I'm imagining something like this:

```yaml
policy:
  approval:
    - rule using defaults

approval_defaults:
  options:
    invalidate_on_push: true
    ignore_update_merges: true

approval_rules:
  - name: rule using defaults
    requires:
      count: 1
      teams: ["palantir/devtools"]
```

PR #1098, opened 2025-11-07
Add accessor methods for common.Methods fields

bluekeyes:
Most of these can already be nil, removing the need to add pointers to distinguish between unset, set to non-zero, and set to zero values. However, the accessors are still useful to allow getting default values in a future PR. It also helps in the one case of a boolean pointer where we had to check for nil each time.

Required as part of adding default support for #41.

### DIFF
--- a/policy/disapproval/disapprove.go
+++ b/policy/disapproval/disapprove.go
@@ -92,12 +92,18 @@ func (p *Policy) Trigger() common.Trigger {
 
 	if !p.Requires.IsZero() {
 		dm := p.Options.GetDisapproveMethods()
-		rm := p.Options.GetRevokeMethods()
-
-		if len(dm.Comments) > 0 || len(rm.Comments) > 0 {
+		if len(dm.GetComments()) > 0 || len(dm.GetCommentPatterns()) > 0 {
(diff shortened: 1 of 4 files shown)


policy/disapproval/disapprove.go
```diff
@@ -92,12 +92,18 @@ func (p *Policy) Trigger() common.Trigger {
...
 		dm := p.Options.GetDisapproveMethods()
-		rm := p.Options.GetRevokeMethods()
+		if len(dm.GetComments()) > 0 || len(dm.GetCommentPatterns()) > 0 {
+			t |= common.TriggerComment
+		}
+		if dm.IsGithubReview() || len(dm.GetGithubReviewCommentPatterns()) > 0 {
+			t |= common.TriggerReview
+		}
 
-		if len(dm.Comments) > 0 || len(rm.Comments) > 0 {
```
bluekeyes, line 97: I'm pretty sure it was a bug that disapproval didn't consider the "pattern" variants here, likely unnoticed because disapproval (especially with non-default options) is rarely used.

admin822:
👍

asvoboda (approved):

PR #1099, opened 2025-11-07
Add internal support for default options

bluekeyes:
This is not exposed to users yet, but provides a cleaner way to set default options that are not the zero value, like the standard approval methods. A future commit will connect these to user-visible defaults in policies and in the server as part of #41.


admin822:
👍

jamestoyer (approved):

Issue #1096, opened 2025-11-07
Reduce confusion for comments that use common emojis


asvoboda:
Occasionally we have internal discussions about the use of `:+1:` or 👍 as approval mechanisms. Usually this is because it can feel a bit magic if policy-bot accepts this in the context of a comment with other text where someone might not intend to approve a PR.

We should consider if there are ways to reduce this confusion:
* Should we modify the approval comment matching to not look for substrings, but instead match on exact text (probably without whitespace?) This would be a breaking change.
* Should there be a server option to "restrict" comments to exact matches? Similar to above but as an option rather than always modifying the behaviour in code?
* Remove +1s in our internal policies and don't change anything in policy-bot?
* Something else?

bluekeyes:
I agree the existing default behavior of substring matching was probably a mistake, but I'm not sure we want to change the behavior of the app at this point given existing usage. Either way, I don't think we need to change how the `comments` method works, given you can use `comment_patterns` to achieve the "full match" behavior.

My proposal for this is that we:

1. Implement #41 to allow default approval methods for all rules in a policy
2. Add support for a server-level default configuration that applies when policies don't specify their own
3. For our internal usage, change the default approval methods at the server level to allow only GitHub reviews and comment patterns that match the full comment

We could then discuss switching the hardcoded defaults from using `comments` to using `comment_patterns`. This would be a behavior change, but people could restore the old behavior if they wanted using the server-level default mechanism above.

bluekeyes:
As of 1.40.0, it's now possible to override the default approval methods in the server. Internally, we're using the following defaults:

```yaml
  approval_defaults:
    options:
      methods:
        comments: []
        comment_patterns:
          - "^\\s*:\\+1:\\s*$"
          - "^\\s*👍[\\x{1F3FB}-\\x{1F3FF}]?\\s*$"
```

To maintain compatibility, I don't think we'll change the hardcoded defaults, but it's something to consider if we ever release a `2.0.0` version with other breaking changes.

PR #1100, opened 2025-11-07
Add support for policy and server default options

### LINKED ISSUE #41: Support rule defaults for common options

bluekeyes:
Polices and server administrators can now define default values for rule options that apply to all rules in the policy or all rules globally, respectively. Rules can still override default values as needed.

This will make it much easier to do things like make sure all rules are invalidated by pushes or use custom comments for approval.

Closes #41.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,10 @@ instance of the server.
   This means it is safe to enable `policy-bot` on all repositories in an
   organization.
 
+Server administrators can also set the `force_shared_policy` option to disable
+loading policies from local repositories, requiring that all repositories use a
+policy supplied in a shared organization repository.
+
 ### policy.yml Specification
 
 The overall policy is expressed by:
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -613,6 +617,81 @@ requires:
...
+      teams: ["palantir/plus-oners"]
+```
+
+Option values are not merged. For example, defining the `ignore_commits_by` or
+`methods.comment_patterns` options in a rule completely replaces any default
+values for these options.
+
+The only exception to this is the `methods` property, which supports limited
+merging. This matches the legacy behavior for how rule values combined with the
+hardcoded server defaults. For example:
```
bluekeyes, line 668: This is probably the most suspicious part of this change. I think we need to keep this to match existing behavior, but I can see an argument that each level should fully override earlier levels so that you don't have to explicitly reset values you don't want.

The existing behavior is that we always set `comments` and `github_reviews` to the server defaults if they're not set by the rule, so this block-level merging is required to replicate that behavior using only the new logic for default values.

I don't think this is that confusing on its own, but it is a bit confusing that the other block-level options, `ignore_commits_by` and `request_review`, don't merge. I could add block-level merging for those, but I think in practice that would also be weird and no one really wants that behavior.

README.md
```diff
@@ -613,6 +617,81 @@ requires:
...
 
+#### Default Options
+
+Policies and server administrators can define new default values for approval
+rule options. This makes it easier to set common options accross all rules.
+Policy Bot tries values in the following order until it finds a set value:
+
+1. Explicit rule options
+2. Policy-level default options
+3. Configurable server-level default options
```
bluekeyes, line 628: TODO: add this to the example server configuration and consider if we should support setting these values via environment variables. We support most other properties, but these are pretty nested and there are a lot of them.

README.md
```diff
@@ -613,6 +617,81 @@ requires:
...
+approval_defaults:
+  # "options" can contain any property available as a rule-level option
+  options:
+    invalidate_on_push: true
+    methods:
+      comments: []
+      comment_patterns: ["^LGTM$"]
+
+approval_rules:
+  # This rule is invalided by new commits and uses the "LGTM" comment to
```
admin822: Nit: "Existing approvals are invalidated by new commits ...."

README.md
admin822: So, how merging works for `options`'s `methods` property is that : if the previous level has the same key as the current one, then overrides using the value from the current level. But for keys that don't exist on the current level but exist on the previous one, uses the value from the previous level?

Using your example, for key `comment_patterns`, rule-level value `["^:+1:$"]` was used to override the previous policy-level default. But all the other keys used values from the previous level.

bluekeyes: Yes, that's correct. Deciding how to clearly explain it was hard without implying that the contents of lists might be merged as well (which they aren't)

README.md
```diff
@@ -613,6 +617,81 @@ requires:
         - "vulnerability scan"
 ```
 
+#### Default Options
+
+Policies and server administrators can define new default values for approval
+rule options. This makes it easier to set common options accross all rules.
```
asvoboda:
```suggestion
rule options. This makes it easier to set common options across all rules.
```

asvoboda (commented):
I think it might be useful to add some tests to help validate and show the hierarchy of overrides that can be allowed, and shows the "limited merging". 

Overall though I think this would work well in practice to help close out #41 and #1096

admin822 (approved):

Issue #1062, opened 2025-09-30
Adding new features  built-in predicates (changed_files_count, commit_count, commit_messages)


Kamalesh-Seervi:
Sample policy snippet
Drop something like this into your policy YAML (for example under a rule’s requires → predicates):

```
predicates:
  changed_files_count:
    total: "< 20"
    added: "<= 5"
    files:
      include:
        - "^src/.*"
      exclude:
        - "^docs/.*"
  commit_count:
    total: "<= 10"
  commit_messages:
    mode: all
    scope: subject
    matches:
      - "^JIRA-[0-9]+:"
    not_matches:
      - "(WIP|fixup)"
 ```

Kamalesh-Seervi:
Shall I work in this too

bluekeyes:
PRs for new predicates are generally welcome and these seem like reasonable things to add. Can you share an example of a policy you are trying to implement where these would help?

If you'd like to work on this, please submit a separate PR for each new predicate to make it easier to review and discuss.

Kamalesh-Seervi:
@bluekeyes I have created the pr plz check this added testing test cases and attached test ss.

I need this feature for my product

PR #1105, opened 2025-11-14
feat: add commit_count predicate

Kamalesh-Seervi:
Add predicate to count total commits in a PR with comparison operators (<, >, =). Implements part of GitHub issue #1062.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/policy/predicate/commit_count.go
+++ b/policy/predicate/commit_count.go
@@ -1,0 +1,72 @@
+// Copyright 2018 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package predicate
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/palantir/policy-bot/policy/common"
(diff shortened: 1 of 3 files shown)


policy/predicate/commit_count.go
```diff
@@ -0,0 +1,72 @@
...
+	Total ComparisonExpr `yaml:"total,omitempty"`
+}
+
+var _ Predicate = &CommitCount{}
+
+func (pred *CommitCount) Evaluate(ctx context.Context, prctx pull.Context) (*common.PredicateResult, error) {
+	predicateResult := common.PredicateResult{
+		ValuePhrase:     "commits",
+		ConditionPhrase: "meet",
+		ConditionsMap:   make(map[string][]string),
```
bluekeyes, line 36: Since there's only one condition in this predicate, I think you can simplify it like this:

```
ValuePhrase: "commit counts",
ConditionPhrase: "meet the conditions",
ConditionValues: []string{pred.Total.String()},
```

The `ConditionsMap` key is useful when you have multiple conditions.

policy/predicate/commit_count.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, errors.Wrap(err, "failed to list commits")
+	}
+
+	count := int64(len(commits))
+
+	if !pred.Total.IsEmpty() {
+		value := fmt.Sprintf("%d", count)
+		cond := fmt.Sprintf("total commits %s", pred.Total.String())
+
+		predicateResult.Values = []string{value}
```
bluekeyes, line 50: Let's set this outside of the `if` so that it is always available

bluekeyes (commented):
Thanks for the contribution! Can you please also add this new predicate to the documentation in the README?

Otherwise, the main logic looks good but I had two small comments about how to construct the the result used in the details view.

Kamalesh-Seervi:
ok will check and update

Issue #862, opened 2024-10-15
Command to validate policy.yml


reegnz:
Today one can only validate `.policy.yml` by running a policy-bot service, and allowing whoever needs to validate to invoke the `/api/validate` HTTP endpoint. But the network path isn't necessarily available from every host that wants to validate the config.

Policy-bot should provide a basic cli command to validate a `.policy.yml`. Eg.: `policy-bot validate .policy.yml`. This way validation can be done fully offline (eg. no network path necessary to a policy-bot service instance).

### Alternatives considered

I've considered running policy-bot locally in a github action as a [service container](https://docs.github.comlocally/en/actions/use-cases-and-examples/using-containerized-services/about-service-containers). But I cannot just launch the container because on startup it requires a valid github app configuration, Which is overkill and hurts reuse if I want to just use it for config validation.

I also considered declaring a schema for the `.policy.yml` and use json schema validation. But that doesn't validate the regexes for validity.

A simple cli command to validate config would be easier on the end-users.

### Context

I'm building a github action to validate `.policy.yml`. I don't want my action to have to call out to a policy-bot service to perform validation, as it makes the action more complex. I don't want my github action runners to have a clear network path to the policy-bot service if it's not absolutely necessary.

bluekeyes:
Fixed by #1114.

PR #1114, opened 2025-11-22
feat: Add validate command

AtzeDeVries:
This is an implementation of adding validate command to policybot. This will implement #862.

FOr now it is a bit rough but if this is the way we want to go we can make it a big nicer. It copies we way how validate is handled as a request.

## Before this PR

Validations of policy required access to a running policybot server. Running a server requires a github applicaton or a mock of a github application. 

## After this PR

A user can run `policy-bot validate -p < path to policy file>`. This is useful if you want to run it in CI.

## Possible downsides?

It adds a little bit of code to maintain i guess.

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -1,0 +1,79 @@
+// Copyright 2018 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


cmd/validate.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func readPolicyBotPolicy(cfgFile string) ([]byte, error) {
+	fi, err := os.Stat(cfgFile)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed fetching policy file: %s", cfgFile)
+	}
+	if os.IsNotExist(err) {
+		return nil, errors.Wrapf(err, "policy file does not exist: %s", cfgFile)
+	}
```
pravindahal:
```suggestion
	if os.IsNotExist(err) {
		return nil, errors.Wrapf(err, "policy file does not exist: %s", cfgFile)
	}
	if err != nil {
		return nil, errors.Wrapf(err, "failed fetching policy file: %s", cfgFile)
	}
```

bluekeyes (approved):
Thanks!

PR #1124, opened 2025-12-03
feat: support glob patterns

### LINKED ISSUE #853: Support glob filename patterns
When working with files, policy-bot should be providing basic glob support.

When working with common unix tools, glob-support tends to be either built-in, or augmented by the wrapping shell evaluating globs. It's not that common using regex to match filename patterns.

Using filename pattern matching is arguable also more secure than raw regex.

Please support either globs or gitignore patterns (which also resemble globs).

Seems like go has glob functionality already built into it's standard library, and is much more performant in matching filenames than regex:
https://github.com/golang/go/blob/master/src/path/filepath/match.go#L44

I imagine config could look like this:

```
changed_files:
  glob: true
  paths:
  - **/*.yaml
  - **/some/**/pattern/*.py
  
```

reegnz:
Other tools in the ecosystem tend to work with globs, makes sense to
make policy-bot also support globs next to regexes.
Globs are also safer to use than regexes as they're less prone to
unintentional file matches.

Fixes #853

Signed-off-by: Zoltán Reegn <zoltan.reegn@gmail.com>

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,9 +241,10 @@ description: "A rule that explains how to configure all of the features"
 # exist, the rule applies to every pull request.
 if:
   # "changed_files" is satisfied if any file in the pull request matches any
-  # regular expression in the "paths" list. If the "ignore" list is present,
-  # files in the pull request matching these regular expressions are ignored
-  # by this rule.
+  # regular expression in the "paths" list or any glob in the "globs" list. If
+  # the "ignore" list is present, files in the pull request matching these
+  # regular expressions are ignored by this rule. If the "ignore_globs" list is
+  # present, files matching those globs are ignored by this rule.
   #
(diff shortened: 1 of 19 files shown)


bluekeyes:
Thanks for the contribution! Although I wish we didn't have to duplicate these properties for globs, I don't see a good alternative to add glob support while maintaining compatibility with existing policies. Maybe this is something we can revisit it if we ever release a 2.x version with breaking changes.

Before I merge this, could you please update the README with the new properties for each predicate?

reegnz:
Yeah I've struggled a lot with how to add it, and it's quite difficult maintaining backwards compatibility otherwise...
I'll add the docs in a day or two.

bluekeyes (approved):

PR #1133, opened 2025-12-09
improvement: parallelize calls to load pr reviewers

bwaldrep:
## Before this PR
A common workflow when contributing to an unfamiliar repo is checking the policybot ui to determine which team(s) to get approval from. Sometimes the mapping is not obvious, and requires clicking to expand the set of valid reviewers for the corresponding node in the policy tree. In orgs with large numbers of teams and/or members, loading the reviewers for a given node of the tree can take multiple minutes, if it succeeds at all. The current implementation simply iterates over the list of relevant orgs and teams and loads their membership serially, which can be very slow.

## After this PR
Parallelize calls to the GH api to load org and team membership.

==COMMIT_MSG==
parallelize calls to load pr reviewers
==COMMIT_MSG==

## Possible downsides?
Increased load on the GH api. For now we'll bound the number of concurrent loads per context to 20, but can adjust later if needed.

### DIFF
--- a/pull/github.go
+++ b/pull/github.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/go-github/v79/github"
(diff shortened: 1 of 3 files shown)


bluekeyes:
Thanks for attempting to improve this. After testing things out locally, I don't think this will help. In the example we discussed, the problem is that the preceding GraphQL query has to execute 90 times, with each request taking about a second. Once that was done, loading the teams serially took less than 200ms total.

I don't think we can easily parallelize the queries due to the way pagination works, so I've filed #1135 to track this. I think fixing the performance here requires either limiting the results to direct collaborators or finding an alternate API to use.

PR #1139, opened 2025-12-15
Add Helm Chart

alankpatel:
This sets up an initial helm chart for policy bot which can be deployed via cloning this repository. Next steps would be to figure out publishing once a name/location is decided.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -1,0 +1,12 @@
+apiVersion: v2
+name: policy-bot
+description: GitHub App for enforcing approval policies on pull requests
+type: application
+version: 0.1.0
+appVersion: "1.40.0"
+maintainers:
+  - name: Palantir Technologies
+    url: https://github.com/palantir/policy-bot
+home: https://github.com/palantir/policy-bot
+sources:
+  - https://github.com/palantir/policy-bot
--- a/chart/README.md
+++ b/chart/README.md
@@ -1,0 +1,265 @@
+# policy-bot Helm Chart
+
+A Helm chart for deploying [policy-bot](https://github.com/palantir/policy-bot) to Kubernetes.
+
+> **Note:** This chart is currently **experimental** and may change in future releases.
+
+## Installation
+
+We recommend deploying policy-bot in its own namespace:
+
+```bash
+# Add your configuration
+cp values.yaml my-values.yaml
+# Edit my-values.yaml with your settings
+
+# Install the chart in a dedicated namespace
+helm install policy-bot ./chart -f my-values.yaml -n policy-bot --create-namespace
+```
+
+## Configuration
+
+### GitHub App Configuration
+
+Before deploying policy-bot, you need to create a GitHub App. See the [main documentation](https://github.com/palantir/policy-bot#github-app-configuration) for detailed instructions.
+
+You'll need the following values from your GitHub App:
+- **Integration ID** (App ID)
+- **Webhook Secret**
+- **Private Key** (PEM format)
+- **OAuth Client ID**
+- **OAuth Client Secret**
+
+### Providing Secrets
+
+There are two ways to provide sensitive configuration:
+
+#### Option 1: Let the chart create the Secret (default)
+
+```yaml
+secrets:
+  create: true
+  github:
+    integrationId: "12345"
+    webhookSecret: "your-webhook-secret"
+    privateKey: |
+      -----BEGIN RSA PRIVATE KEY-----
+      ...
+      -----END RSA PRIVATE KEY-----
+  oauth:
+    clientId: "your-client-id"
+    clientSecret: "your-client-secret"
+  session:
+    key: "a-random-string-at-least-32-characters"
+```
+
+#### Option 2: Use an existing Secret
+
+Create a Secret with the following keys:
+- `github-integration-id`
+- `github-webhook-secret`
+- `github-private-key`
+- `oauth-client-id`
+- `oauth-client-secret`
+- `session-key`
+
+Then reference it:
+
+```yaml
+secrets:
+  create: false
+  existingSecret: "my-policy-bot-secret"
+```
+
+### Providing Configuration
+
+#### Option 1: Let the chart create the ConfigMap (default)
+
+Configure options under the `config` key in values.yaml:
+
+```yaml
+config:
+  create: true
+  server:
+    address: "0.0.0.0"
+    port: 8080
+    publicUrl: "https://policy-bot.example.com"
+  github:
+    webUrl: "https://github.com"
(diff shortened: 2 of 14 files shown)


alankpatel:
some examples of it working with the ingress set up as well installed in k8s via local minikube cluster
image
image

chart/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v2
+name: policy-bot
```
alankpatel, line 2: this or helm-chart-policy-bot or etc? if we wanted to deploy the helm chart using docker hub having a different name could be useful

chart/values.yaml
```diff
@@ -0,0 +1,191 @@
...
+  hosts:
+    - host: policy-bot.local
+      paths:
+        - path: /
+          pathType: Prefix
+  # TLS configuration (3 options):
+  # Option 1: Provide cert/key directly, chart creates the secret
+  # Option 2: Reference an existing secret via existingSecret
+  # Option 3: Use cert-manager by adding annotation above and leave false
+  tls:
```
alankpatel, line 122: tried to give as many options as possible since different people like to have their k8s ingress set up differently. can disable if they want to control it

chart/values.yaml
```diff
@@ -0,0 +1,191 @@
...
+  # Datadog metrics (optional)
+  datadog:
+    enabled: false
+    address: "127.0.0.1:8125"
+    interval: "10s"
+    tags: []
+
+  # Prometheus metrics (optional)
+  prometheus:
+    enabled: false
```
bluekeyes: I think this is a bit misleading because the Prometheus endpoint is always enabled (and we always attach the scrape annotations to the deployment.) While overriding the quantiles is optional, metrics are always available.

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,126 @@
...
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: {{ include "policy-bot.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args:
+            - server
+            - --config
+            - /secrets/policy-bot.yml
+          env:
+            - name: POLICYBOT_PUBLIC_URL
+              value: {{ include "policy-bot.publicUrl" . | quote }}
```
bluekeyes: This is also set in the config file - is the env variable necessary?

chart/Chart.yaml
bluekeyes: I feel like `policy-bot` makes the most sense, but I'm not familiar with the requirements to publish to different locations.

bluekeyes (commented):
Thanks, I think this looks good but I want to take a closer look all of the details - leaving some initial comments for now.

chart/Chart.yaml
jamestoyer: `policy-bot` is the right name for this as the fact this is in a `Chart.yaml` indicates that it is a helm chart.

chart/values.yaml
```diff
@@ -0,0 +1,191 @@
...
+    expandRequiredReviewers: false
+    # -- Force use of shared policy for all repositories
+    forceSharedPolicy: false
+
+  # Datadog metrics (optional)
+  datadog:
+    enabled: false
+    address: "127.0.0.1:8125"
+    interval: "10s"
+    tags: []
```
jamestoyer: I'd be tempted to omit this completely. Most apps in k8s are expected to use an OpenMetrics/Prometheus endpoint and I'm inclined not to allow this as an option.

chart/values.yaml
```diff
@@ -0,0 +1,191 @@
...
+nodeSelector: {}
+
+# -- Tolerations for pod assignment
+tolerations: []
+
+# -- Affinity rules for pod assignment
+affinity: {}
+
+# Policy-bot specific configuration
+config:
```
jamestoyer, line 134: Can we have an "existingConfigMap" or similar option? There are likely to be cases where people want to set options that aren't in the config, which is something I've done in the past for other helm charts that support it. 

[Here](https://github.com/grafana/alloy/blob/e2e96e95ff0dab600befbe63165e10eea096b968/operations/helm/charts/alloy/values.yaml#L31-L41) is a pretty standard example of how others set configs.

jamestoyer (commented):
We should include a README.md in the chart directory that gives more information about how to use the chart. I'd also look to decorate the `values.yaml` file with more in depth information about what each does. We could use something like https://github.com/bitnami/readme-generator-for-helm to output these to the README.

chart/values.yaml
alankpatel: followed the grafana chart example

chart/values.yaml
alankpatel: reworded the description and the values to make it more clear but open to other ways of letting users pass that information

alankpatel:
> We should include a README.md in the chart directory that gives more information about how to use the chart. I'd also look to decorate the `values.yaml` file with more in depth information about what each does. We could use something like https://github.com/bitnami/readme-generator-for-helm to output these to the README.

the readme generator seems pretty helpful. I changed the values file to use the format the generator requires so we can run the command to update the file

chart/values.yaml
```diff
@@ -0,0 +1,235 @@
...
+## @param replicaCount Number of replicas to deploy. policy-bot is stateless and safe to run multiple instances.
+replicaCount: 2
+
+## @param image.repository Docker image repository
+## @param image.pullPolicy Image pull policy (Always, IfNotPresent, Never)
+## @param image.tag Overrides the image tag whose default is the chart appVersion
+image:
+  repository: palantirtechnologies/policy-bot
+  pullPolicy: IfNotPresent
+  tag: ""
```
jamestoyer, line 16: We should set a real tag here that is pinned to the release that is about to be cut. We either need to see if godel can do this without failing a verify or update how we do releases.

chart/README.md
```diff
@@ -0,0 +1,265 @@
...
+      paths:
+        - path: /
+          pathType: Prefix
+```
+
+### Metrics
+
+Policy-bot exposes Prometheus metrics at `/api/metrics`
+
+## Parameters
```
jamestoyer, line 137: Have we documented how we generated these? Or at least update the godel config to run this on generate.

chart/values.yaml
alankpatel: Currently I have it set to use the app version set in the Chart.yaml. I think it makes sense to update the chart and app versions during release automatically but want to include that in a separate pr that handles publishing this to docker hub

chart/README.md
alankpatel: added a section in the readme to include generate steps

PR #1140, opened 2025-12-16
Optimise collaborator permission filtering

### LINKED ISSUE #1135: Loading reviewers by permission can be slow in large organizations
The details view provides an option to display the users who can approve particular rules. Loading this list is relatively fast when using team or organization members (provided the teams or organization aren't that large), but can be very slow when loading by permissions. From some local testing, this seems to be due to the query to list collaborators:

https://github.com/palantir/policy-bot/blob/d80588c1d5ed7301ff25ca859469c06596cb05cc/pull/github.go#L519-L536

In one internal example, this required 90 requests, taking a total of about 92 seconds.

There are probably some minor optimizations here. Splitting in to two queries might improve things as each query could use a larger page size. We could also potentially run these queries in parallel with the other parts of this function, but the rest of the function completed in about 300ms in this test.

But I think the real fix is re-evaluating how this method works. The current strategy was devised in 2021 based on API limitations at the time. We should see if GitHub provides better APIs for the information we need or ways to limit the data returned.

Internally, we generally use this feature with teams, so another option may be to limit things to direct collaborators. This would ignore collaborators who get their permissions through the organization and would significantly limit the number of users we process in most cases. This is already a documented restriction for requesting reviews, so it might be reasonable to adopt it here as well.

Notably, this is not a problem for evaluation, because it's fast to check if any given user has particular permissions. The problem is in listing all of the users who have a certain permission level.

leny-mi:
We're using https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators instead of the graphql request at https://github.com/palantir/policy-bot/blob/d80588c1d5ed7301ff25ca859469c06596cb05cc/pull/github.go#L519-L536. The REST api allows us to filter for minimum permissions. Given a policy that determines approval through permissions, we find the "lowest" permission and use it to filter results which significantly reduces the number of requests we need to make in cases where few members have the required permission level.

## Before this PR
Currently it takes ~80s to fetch users that can approve some PRs (depending on member count)

## After this PR

After this PR it should take ~6-10s, depending on team sizes. Should close #1135.
==COMMIT_MSG==
==COMMIT_MSG==


pull/github.go
```diff
@@ -516,77 +516,62 @@ func (ghc *GitHubContext) RepositoryCollaborators() ([]*Collaborator, error) {
...
-					Nodes []v4Actor
-				} `graphql:"all: collaborators(affiliation: ALL, first: 50, after: $allCursor)"`
-			} `graphql:"repository(owner: $owner, name: $name)"`
-		}
-		qvars := map[string]interface{}{
-			"owner":        githubv4.String(ghc.owner),
-			"name":         githubv4.String(ghc.repo),
-			"directCursor": (*githubv4.String)(nil),
-			"allCursor":    (*githubv4.String)(nil),
+		permissionToGitHub := map[Permission]string{
```
bluekeyes: Instead of having this map here, let's define a `GitHubString()` method on the `Permission` type that returns the strings as used in the GitHub API.

pull/github.go
```diff
@@ -610,7 +592,7 @@ func (ghc *GitHubContext) RepositoryCollaborators() ([]*Collaborator, error) {
 		}
 
 		fillPermissions := func(c *Collaborator) {
-			overall := c.Permissions[0].Permission // from above, every collaborator has at least one permission
```
bluekeyes, line 613: Any reason to remove this comment?

pull/github_test.go
```diff
@@ -777,3 +787,27 @@ func (c *MockGlobalCache) GetPushedAt(repoID int64, sha string) (time.Time, bool
 func (c *MockGlobalCache) SetPushedAt(repoID int64, sha string, t time.Time) {
 	c.PushedAt[fmt.Sprintf("%d:%s", repoID, sha)] = t
 }
+
+// PathAndQueryMatcher matches both path and query parameters
+type PathAndQueryMatcher struct {
```
bluekeyes: Let's put this in `pull/responseplayer_test.go` with the other matcher implementations, since it is pretty generic

pull/github_test.go
```diff
@@ -777,3 +787,27 @@ func (c *MockGlobalCache) GetPushedAt(repoID int64, sha string) (time.Time, bool
...
+	Query url.Values
+}
+
+func (m PathAndQueryMatcher) Matches(r *http.Request, body []byte) bool {
+	if r.URL.Path != m.Path {
+		return false
+	}
+	for key, expectedValues := range m.Query {
+		actualValues := r.URL.Query()[key]
+		if len(actualValues) != len(expectedValues) {
```
bluekeyes: Can this use `slices.Equal` to compare these instead of checking both the length and each item?

policy/reviewer/reviewer.go
```diff
@@ -241,7 +241,16 @@ func selectUserReviewers(ctx context.Context, prctx pull.Context, selection *Sel
 		}
 	}
 
-	collaborators, err := prctx.RepositoryCollaborators()
+	minPerm := pull.PermissionNone
+	if len(result.ReviewRequestRule.Permissions) > 0 {
+		minPerm = result.ReviewRequestRule.Permissions[0]
```
bluekeyes: Can this use `slices.Min` to get the minimum value?

server/handler/details_reviewers.go
```diff
@@ -98,10 +98,15 @@ func (h *DetailsReviewers) ServeHTTP(w http.ResponseWriter, r *http.Request) err
 		}
 	}
 
-	// Add reviewers with permissions
 	perms := requires.Actors.GetPermissions()
 	if len(perms) > 0 {
-		userCollaborators, err := prctx.RepositoryCollaborators()
+		minPerm := perms[0]
```
bluekeyes: Can this use `slices.Min` to get the minimum value?

bluekeyes (commented):
Thanks, this approach seems promising and the use of the REST API means we should get some benefits from the client's HTTP cache too, which is not enabled on the GraphQL client.

pull/github.go
leny-mi: Was too eager with clauding. I've restored it.

leny-mi:
I've noticed that the cache didn't respect the provided permission level. In practice I think this wouldn't have been an issue but I've still fixed the cache to be per-permission

bluekeyes (approved):

PR #1148, opened 2026-01-08
fix: make IsActor handle 'permissions' via 'AND' condition

smudge:
## Before this PR

Given the following rule:

```yaml
  - name: reviewer is an org member and has write access
    requires:
      count: 1
      organizations:
        - "our-org"
      permissions:
        - write
```

I would expect the following truth table to apply:

| org       | permissions | IsActor |
|-----------|-------------|---------|
| our-org   | write       | true    |
| our-org   | read        | false   |
| other-org | write       | false   |
| other-org | read        | false   |

However, the current code effectively treats these as an "OR" condition and produces the following truth table:

| org       | permissions | IsActor  |
|-----------|-------------|----------|
| our-org   | write       | true     |
| our-org   | read        | **true** |
| other-org | write       | **true** |
| other-org | read        | false    |

## After this PR

In reading the implementation, I understand why the other identity conditions (users, teams, organizations) all use OR logic. With the exception of permissions, it's essentially the same mental model as the "Collaborators and teams" page in a GitHub repo's settings:

Screenshot 2026-01-08 at 1 15 01 PM

Note how `'permissions'` are treated in the GitHub model. This PR brings the `policy-bot` behavior in line with what seems to be the more likely expectation of the developer/maintainer of a rule set.

==COMMIT_MSG==
fix: make IsActor handle 'permissions' via 'AND' condition
==COMMIT_MSG==

## Possible downsides?

This is technically a breaking change for anyone who had been _intentionally_ relying on permissions to apply via an "OR" condition.

## Alternatives Considered

- The YML config could be expanded to support more explicit `- and:` and `-or:` mappings.
- The YML config could be expanded to nest permissions beneath individual identity requirements. (This is closest to the "Collaborators and teams" settings page.)
- Without any code changes, we can _sort of_ (but not entirely) get what we need via something like this:

```yaml
  - name: reviewer is an org member
    requires:
      count: 1
      organizations:
        - "our-org"
  - name: reviewer has write access
    requires:
      count: 1
      permissions:
        - write
```

However, our use case ideally enforces that **the reviewer from a given org be the _same_ reviewer as the one with the write permissions**. The above example results in the ability for _two reviewers_ (one with `write` access and the other with read-only org membership) to combine two reviews into a single policy bot approval.

Therefore, without some form of code change, it does not appear to be possible to enforce both an identity check and a permissions check for the same "actor". (Which, at the very least, is not in line with the way that permissions are typically expressed within the GitHub interface -- but please point me in the right direction if I'm missing anything!)

### DIFF
--- a/policy/common/actor.go
+++ b/policy/common/actor.go
@@ -73,6 +73,32 @@ func (a *Actors) GetPermissions() []pull.Permission {
 // IsActor returns true if the given user satisfies at least one of the
 // conditions in this structure.
 func (a *Actors) IsActor(ctx context.Context, prctx pull.Context, user string) (bool, error) {
+	permissions := a.GetPermissions()
+	if len(permissions) > 0 {
+		userPerm, err := prctx.CollaboratorPermission(user)
+		if err != nil {
+			return false, errors.Wrap(err, "failed to get collaborator permission")
+		}
+		if userPerm == pull.PermissionNone {
+			return false, nil
+		}
+
+		satisfiesPermissions := false
+		for _, p := range permissions {
+			if userPerm >= p {
+				satisfiesPermissions = true
+				break
+			}
+		}
+		if !satisfiesPermissions {
+			return false, nil
+		}
+
+		if len(a.Users)+len(a.Teams)+len(a.Organizations) == 0 {
+			return true, nil
+		}
+	}
+
 	for _, u := range a.Users {
 		if user == u {
 			return true, nil
@@ -96,23 +122,6 @@ func (a *Actors) IsActor(ctx context.Context, prctx pull.Context, user string) (
 		}
 		if member {
 			return true, nil
-		}
-	}
-
-	permissions := a.GetPermissions()
-	if len(permissions) > 0 {
-		userPerm, err := prctx.CollaboratorPermission(user)
-		if err != nil {
-			return false, err
-		}
-		if userPerm == pull.PermissionNone {
-			return false, nil
-		}
-
-		for _, p := range permissions {
-			if userPerm >= p {
-				return true, nil
-			}
 		}
 	}
 
--- a/policy/common/actor_test.go
+++ b/policy/common/actor_test.go
@@ -113,6 +113,24 @@ func TestIsActor(t *testing.T) {
 		assertActor(t, a, "jstrawnickel")
 		assertNotActor(t, a, "ttest")
 	})
+
+	t.Run("identityAndPermissions", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @smudge! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

bluekeyes:
Thanks for the contribution. Unfortunately, I don't think this is something we can accept as-is due to the breaking behavior change in a core part of Policy Bot. That said, I'm open to improving documentation if you found it lacking or discussing other ways you might be able to achieve the result you want.

For some background, the OR design here is intentional and the `permissions` key is meant to be an alternative to the other options rather than a filter. The way we normally use this internally is by requiring the `maintain` permission. We then assign the teams that should be allowed to review the `maintain` permission on the repository in GitHub. This makes it easier to share the same policy between repositories that have different approvers and allows granting write permissions to anyone who needs them.

When projects need more specific approval conditions, they usually use policies that specify teams. The `organizations` property is something we rarely use (organizations generally have too many members to be useful) and is a feature I'm not sure I would add again. I don't think I've seen a case yet where we couldn't combine identity and access into a single element, either through a team or with the `maintain` permission.

smudge:
Totally understood -- yeah I suspected this would be the case, because at a high level the OR logic generally makes sense. It just gets gnarly when a policy appears to read like "reviewers on team X with write access" when it's actually "reviewers on team X _or_ that have write access" -- adding specificity to a policy may feel natural but turns out to behave differently than intended, even if the functionality itself is internally consistent.

In addition to some form of documentation, would you be open to an optional "nested" flavor that looks like this:

```yaml
- name: reviewer is an org member and has write access
    requires:
      count: 1
      organizations:
        - name: "our-org"
          role: "write"
```

Backwards-compatibility would be easy to maintain for rules that provide a list of strings instead of key-value pairs.

Ultimately, that extended syntax is most in line with GitHub's built-in "Collaborators and teams" settings, but I wanted to propose the smaller code change before asking about a bigger change to the YAML DSL.

bluekeyes:
Yes, I'm open to modifying the `organizations` property to accept a new optional syntax that allows filtering by role, but I'm still unclear on the situations in which using an organization and permission is preferable to using teams (or using only a more exclusive permission like `maintain`.) Could you share some more details about your setup and the outcome you want?

smudge:
Realizing I never closed this! We found some reasonable workarounds, and I haven't had a chance to circle back. Closing for now 👍

Issue #1152, opened 2026-01-14
N `has_workflow_result` causes N duplicate calls to the `/actions/runs` endpoint


denysvitali-kaiko:
In our fork, we enabled tracing for policy-bot (would be very useful to add it here too!) because we were facing an incredibly slow policy evaluation - 8s+.

After adding tracing and analyzing the trace, the problem was obvious:

<img width="1810" height="919" alt="Image" src="https://github.com/user-attachments/assets/3ff452c1-4126-4419-852e-d0d957ae2d67" />

In our policies we heavily rely on `has_workflow_result`. It looks like each `has_workflow_result` entry causes an HTTP call to the exact same endpoint, that generally takes 600ms+ to process.

Since there is no caching, despite the endpoint being exactly the same, the request takes 8s+.

I'm working on a fix, but wanted to report the bug here first.

bluekeyes:
Thanks for the report, this is definitely a bug. We don't really use the `has_workflow_result` predicate in any of our internal policies and I missed that the cache wasn't correctly implemented when originally reviewing this feature.

PR #1153, opened 2026-01-14
fix(github): API calls causing performance issues

denysvitali-kaiko:
## Before this PR

Multiple bugs and performance bottlenecks cause the policy evaluation to take >8s

See for example:
- #1152

Concurrent GitHub API work could bypass caching and trigger duplicate actions or racey behavior under load.

image

## After this PR

Fewer requests, it now take ~1s to process the request:

image

## Possible downsides?

None I can think of.

### DIFF
--- a/pull/github_test.go
+++ b/pull/github_test.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/google/go-github/v81/github"
+	"github.com/h2non/gock"
 	"github.com/shurcooL/githubv4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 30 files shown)


pull/github_test.go
```diff
@@ -686,6 +687,53 @@ func TestLatestWorkflowRuns(t *testing.T) {
...
 }
 
+func TestLatestWorkflowRunsCaching(t *testing.T) {
+	defer gock.Off()
+
+	pr := defaultTestPR()
+
+	// Mock the workflow runs endpoint - Times(1) means only 1 call allowed
+	// If caching doesn't work, the second call will fail with "gock: cannot match any request"
+	gock.New("http://github.localhost").
```
bluekeyes, line 697: Please avoid introducing new dependencies for testing unless they are necessary. The existing tests already provide a way to mock requests and count the number of matches for any particular route in order to verify that caching works.

server/server.go
```diff
@@ -161,7 +161,11 @@ func New(c *Config) (*Server, error) {
 	basePolicyHandler := handler.Base{
 		ClientCreator: cc,
 		BaseConfig:    &c.Server,
-		Installations: githubapp.NewInstallationsService(appClient),
+		Installations: githubapp.NewCachingInstallationsService(
+			githubapp.NewInstallationsService(appClient),
+			5*time.Minute,  // expiry - installation ID rarely changes
```
bluekeyes, line 166: We typically use 10 minutes here, but something even longer is probably safe

bluekeyes (commented):
Thanks for proposing some potential performance improvements. Right now, it looks like this is doing 4 different things that need various levels of discussion:

1. It fixes the workflow caching bug, but does this by changing how loading works in general. Since this is an obvious bug, I created #1159 to fix only that without changing other behavior.
2. It parallelizes some parts of loading PR data. This is effectively trading speed against the number of API requests, which requires more thought. The implementation here leads to more API requests in two situations:
   1. When policy-bot is installed on repositories that do not define policy files, it starts fetching PR details before determining if a policy exists. This means the results of these API requests may be discarded if the repository does not define a policy.
   2. By picking a fixed set of data to load preemptively, policy-bot will now make API requests for data that may go unused by the repository's policy. As a specific case, our internal policy-bot instance does 10,000-20,000 evaluations per day and we do not use the `has_workflow_result` predicate in any policy. With this change, policy-bot would now add at least one more API request to all of those evaluations for data that it doesn't need. A similar thing might be true of the modified files. Only the "paged data" (commits, comments, and reviews) are used by every evaluation.
   
   Historically, we've seen that API request rate is usually the biggest problem. GitHub rate limits can be strict and are not easy to increase, so we tend to get more reports about hitting rate limits than we do about increasing evaluation speed.

   GitHub also enforces secondary rate limits, which apply when an app makes too many requests in a short period of time, even if the total number of requests is still within the primary rate limit. I'm concerned that parallel requests scaled to many evaluations across many repositories will make us more likely to trigger secondary limits.
3. It parallelizes parts of loading data for the details page. I think this is safer than the parallelization in PR data (we only make one unnecessary request in the case where the permission check fails) and it speeds up a page load where a user is actively waiting for a response. I'd want to measure the impact this actually has in practice, but it seems like it could be a worthwhile improvement.
4. It enables caching for installation IDs. I think this is a free performance gain with no downside and is something we forgot to come back and update after introducing the caching variant of `InstallationsService`

Summarizing:

* The workflow caching bug should be fixed by #1159
* I don't think I can accept the parallelization changes in the `pull` package as-is, but the idea is worth considering if there is a way to address the concerns with rate limits without adding too much complexity
* If the parallelization in the details handler makes a meaningful difference in page load time, I'd consider accepting that as its own PR
* I'm happy to accept the installation service caching change as its own PR

PR #1149, opened 2026-01-09
Add CI, publish to GHCR, add linters and address linter errors

denysvitali-kaiko:
# CI Improvements and Linting Fixes

## Summary

This PR enhances the CI/CD pipeline with additional code quality checks and container registry publishing options, while also fixing all existing linter warnings in the codebase.

#### New Quality Check Workflows
Added 6 new GitHub Actions workflows for comprehensive code quality checks:

| Workflow | Purpose |
|----------|---------|
| `go-mod-tidy.yml` | Ensures `go.mod` and `go.sum` are properly tidied |
| `go-vet.yml` | Runs `go vet` to catch suspicious constructs |
| `gofmt.yml` | Validates code formatting with `gofmt` |
| `goimports.yml` | Checks import statement formatting |
| `golangci-lint.yml` | Runs comprehensive linting via golangci-lint v9 |
| `staticcheck.yml` | Performs advanced static analysis |

All workflows:
- Trigger on pushes to `develop` and pull requests
- Automatically detect Go version from `go.mod`
- Exclude vendor directory where applicable

#### Container Registry Improvements
- **GHCR Publishing**: Added GitHub Container Registry (ghcr.io) as an additional publish target alongside Docker Hub
- **Conditional Docker Hub Push**: Docker Hub publishing now only occurs when `DOCKERHUB_USERNAME` secret is configured, preventing CI failures in forks without Docker Hub credentials. This allows forks to sort of immediately get started with their builds / docker images, without having to set Docker Hub credentials.

#### Deprecated API Updates
- Replaced deprecated `github.String()`, `github.Bool()`, `github.Int()`, `github.Int64()` helpers with `github.Ptr()` (go-github v66+ migration)
- Replaced deprecated `ioutil.ReadAll()` and `ioutil.ReadFile()` with `io.ReadAll()` and `os.ReadFile()` (Go 1.16+ deprecation)
- Replaced deprecated `strings.Title()` with `cases.Title(language.English).String` from `golang.org/x/text`
- Updated datadog import path from `github.com/palantir/go-baseapp/baseapp/datadog` to `github.com/palantir/go-baseapp/appmetrics/emitter/datadog`

#### Code Quality Fixes
- Removed unnecessary `fmt.Sprintf()` with no format arguments (`policy/approval/evaluate.go`)
- Removed unused variable `signerList` (`policy/predicate/signature.go`)
[...]

PR #1157, opened 2026-01-14
ci: add go vet and staticcheck workflows

denysvitali-kaiko:
## Before this PR

Code was not validated against `go vet` or `staticcheck`

## After this PR

Two new workflows take care of this.

## Possible downsides?

Existing issues might cause the workflow to fail. #1156 should fix this though.

### DIFF
--- a/godel/config/golangci-lint-plugin.yml
+++ b/godel/config/golangci-lint-plugin.yml
@@ -1,0 +1,4 @@
+linters:
+  enable:
+   - govet
+   - staticcheck
--- a/godel/config/golangci-lint-plugin.yml
+++ b/godel/config/golangci-lint-plugin.yml
@@ -1,4 +1,3 @@
 linters:
   enable:
-   - govet
    - staticcheck
--- a/policy/approval/evaluate.go
+++ b/policy/approval/evaluate.go
@@ -179,7 +179,7 @@ func (r *AndRequirement) Evaluate(ctx context.Context, prctx pull.Context) commo
 	switch {
 	case approved > 0 && pending == 0:
 		status = common.StatusApproved
-		description = fmt.Sprintf("All rules are approved")
+		description = "All rules are approved"
 	case pending > 0:
 		status = common.StatusPending
 		description = fmt.Sprintf("%d/%d rules approved", approved, approved+pending)
--- a/pull/pulltest/context.go
+++ b/pull/pulltest/context.go
@@ -211,11 +211,7 @@ func (c *Context) OrganizationMembers(org string) ([]string, error) {
 	inverted := make(map[string][]string)
 	for user, orgs := range c.OrgMemberships {
 		for _, o := range orgs {
-			if _, ok := inverted[o]; ok {
-				inverted[o] = append(inverted[o], user)
-			} else {
-				inverted[o] = []string{user}
-			}
+			inverted[o] = append(inverted[o], user)
 		}
 	}
 
@@ -231,11 +227,7 @@ func (c *Context) TeamMembers(team string) ([]string, error) {
 
 	for user, teams := range c.TeamMemberships {
 		for _, t := range teams {
-			if _, ok := inverted[t]; ok {
-				inverted[t] = append(inverted[t], user)
-			} else {
-				inverted[t] = []string{user}
-			}
+			inverted[t] = append(inverted[t], user)
 		}
 	}
 
--- a/server/handler/details.go
+++ b/server/handler/details.go
@@ -135,7 +135,7 @@ func (h *Details) getStateIfAllowed(w http.ResponseWriter, r *http.Request) *Det
 		return nil
 	}
 
-	client, err := h.ClientCreator.NewInstallationClient(installation.ID)
+	client, err := h.NewInstallationClient(installation.ID)
 	if err != nil {
 		hatpear.Store(r, errors.Wrap(err, "failed to create github client"))
 		return nil
--- a/server/handler/frontend.go
+++ b/server/handler/frontend.go
@@ -173,9 +173,7 @@ func getMethods(result *common.Result) map[string][]string {
 	)
 
 	patternInfo := make(map[string][]string)
-	for _, comment := range result.Methods.GetComments() {
-		patternInfo[commentKey] = append(patternInfo[commentKey], comment)
-	}
+	patternInfo[commentKey] = append(patternInfo[commentKey], result.Methods.GetComments()...)
 	for _, commentPattern := range result.Methods.GetCommentPatterns() {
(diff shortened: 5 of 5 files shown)


jamestoyer:
Both of these are included in golangci-lint, which is called by Godel, as mentioned in #1149. Instead of adding these new checks, please update the configuration for Godel and golangci-lint to run these.

denysvitali-kaiko:
I'm new to Godel - but it looks like that specific plugin is only used [in one place](https://github.com/search?q=godel-golangci-lint-plugin++NOT+repo%3Apalantir%2Fgodel-golangci-lint-plugin&type=code) in the entirety of GitHub. On top of that, it seems like you already have your config here:
https://github.com/palantir/godel/blob/f3258547239f5ab2cfa70483e0943549b18509a0/framework/godellauncher/defaulttasks/defaulttasks.go#L101-L125

What approach should I take here?

jamestoyer:
The docs for the plugin can be found at https://github.com/palantir/godel-golangci-lint-plugin. An empty `godel/config/golangci-lint-plugin.yml` tells Godel to use the default settings, anything you want to change should be added to that file.

denysvitali-kaiko:
Perfect! Now it works!

```
$ ./godelw lint                               
policy/approval/evaluate.go:182:17: S1039: unnecessary use of fmt.Sprintf (staticcheck)
                description = fmt.Sprintf("All rules are approved")
                              ^
policy/predicate/file_test.go:31:29: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (staticcheck)
                        common.NewCompiledRegexp(regexp.MustCompile("app/.*\\.go")),
                                                 ^
policy/predicate/file_test.go:32:29: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (staticcheck)
                        common.NewCompiledRegexp(regexp.MustCompile("server/.*\\.go")),
                                                 ^
policy/predicate/file_test.go:35:29: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (staticcheck)
                        common.NewCompiledRegexp(regexp.MustCompile(".*/special\\.go")),
                                                 ^
policy/predicate/signature.go:113:16: SA4010: this result of append is never used, except maybe in other appends (staticcheck)
                signerList = append(signerList, signer)
                             ^
policy/predicate/status_test.go:166:23: SA1019: github.Bool is deprecated: use Ptr instead. (staticcheck)
                        overrideSatisfied: github.Bool(true),
                                           ^
pull/github_test.go:447:7: SA1019: github.Int64 is deprecated: use Ptr instead. (staticcheck)
                ID: github.Int64(12345),
                    ^
pull/github_test.go:449:11: SA1019: github.String is deprecated: use Ptr instead. (staticcheck)
                        Login: github.String("testorg2"),
                               ^
pull/github_test.go:451:9: SA1019: github.String is deprecated: use Ptr instead. (staticcheck)
                Name: github.String("testrepofork"),
                      ^
pull/github_test.go:739:14: SA1019: github.String is deprecated: use Ptr instead. (staticcheck)
                Title:     github.String("test title"),
                           ^
pull/github_test.go:741:14: SA1019: github.Int is deprecated: use Ptr instead. (staticcheck)
                Number:    github.Int(123),
                           ^
pull/github_test.go:743:14: SA1019: github.Bool is deprecated: use Ptr instead. (staticcheck)
                Draft:     github.Bool(false),
                           ^
pull/github_test.go:751:9: SA1019: github.Int64 is deprecated: use Ptr instead. (staticcheck)
                                ID: github.Int64(1234),
                                    ^
pull/github_test.go:761:9: SA1019: github.Int64 is deprecated: use Ptr instead. (staticcheck)
                                ID: github.Int64(1234),
                                    ^
pull/github_test.go:768:17: SA1019: github.Int is deprecated: use Ptr instead. (staticcheck)
                ChangedFiles: github.Int(1),
                              ^
pull/pulltest/context.go:214:4: S1036: unnecessary guard around map access (staticcheck)
                        if _, ok := inverted[o]; ok {
                        ^
pull/pulltest/context.go:234:4: S1036: unnecessary guard around map access (staticcheck)
                        if _, ok := inverted[t]; ok {
                        ^
pull/responseplayer_test.go:19:2: SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package [io] or package [os], and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
server/handler/details.go:138:19: QF1008: could remove embedded field "ClientCreator" from selector (staticcheck)
        client, err := h.ClientCreator.NewInstallationClient(installation.ID)
                         ^
server/handler/frontend.go:94:17: SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead. (staticcheck)
                        "titlecase": strings.Title,
                                     ^
server/handler/frontend.go:176:2: S1011: should replace loop with patternInfo[commentKey] = append(patternInfo[commentKey], result.Methods.GetComments()...) (staticcheck)
        for _, comment := range result.Methods.GetComments() {
        ^
server/handler/validate.go:19:2: SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package [io] or package [os], and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
server/handler/validate.go:66:3: S1023: redundant return statement (staticcheck)
                return
                ^
server/server.go:31:2: SA1019: "github.com/palantir/go-baseapp/baseapp/datadog" is deprecated: Use the appmetrics/emitter/datadog package instead. (staticcheck)
        "github.com/palantir/go-baseapp/baseapp/datadog"
        ^
server/server.go:280:13: SA1019: datadog.StartEmitter is deprecated: Use the appmetrics/emitter/datadog package instead. (staticcheck)
                if err := datadog.StartEmitter(s.base, s.config.Datadog); err != nil {
                          ^
25 issues:
* staticcheck: 25
```

These are fixed in #1156

godel/config/golangci-lint-plugin.yml
```diff
@@ -0,0 +1,4 @@
+linters:
+  enable:
+   - govet
```
jamestoyer: This is not required as it run by default. You can check this by running `./godelw linters`.

jamestoyer (commented):
In principal this looks fine once the change I've requests has been made. We'd need to have a further discussion internally as to whether we ant to enable `staticcheck` on this repo, and more broadly on our other repos.

PR #1185, opened 2026-02-24
return full result tree through simulation endpoint

alankpatel:
## Before this PR
When Policy Bot's simulation API encountered partial failures (e.g., an or requirement where one child errors but a sibling is pending or approved), the child error is ignored. The API returns status: "pending" or status: "approved" with no error, giving callers no way to see that a branch is broken and impossible to satisfy.

## After this PR
The simulation API response includes a partial_errors field that returns suppressed child errors. Callers can check len(partial_errors) > 0 to detect broken branches and trigger fallback behavior, if needed. This change should be backwards compatible since it will be ignored unless the caller decides to deserialize partial_errors.

## Possible downsides?
Since it is backwards compatible there should be little downside. It may increase the response time but it only traversing the tree once so that should be minimal.

Tested by creating a pr with over 3000 files and hitting the simulation endpoint when the overall policy is returning status approved but still has child errors

### DIFF
--- a/server/handler/simulate.go
+++ b/server/handler/simulate.go
@@ -34,14 +34,23 @@ type Simulate struct {
 	Base
 }
 
+// PartialError represents an error from a child rule that was suppressed by its
+// parent (e.g. an Or requirement that cleared the error because a sibling was
+// pending or approved)
+type PartialError struct {
+	Rule  string `json:"rule"`
+	Error string `json:"error"`
+}
+
 // SimulationResponse is the response returned from Simulate, this is a trimmed down version of common.Result with json
 // tags. This struct and the newSimulationResponse constructor can be extended to include extra content from common.Result.
 type SimulationResponse struct {
-	Name              string `json:"name"`
-	Description       string `json:"description:"`
-	StatusDescription string `json:"status_description"`
-	Status            string `json:"status"`
-	Error             string `json:"error"`
+	Name              string          `json:"name"`
+	Description       string          `json:"description:"`
+	StatusDescription string          `json:"status_description"`
+	Status            string          `json:"status"`
+	Error             string          `json:"error"`
+	PartialErrors     []*PartialError `json:"partial_errors,omitempty"`
 }
 
 type ErrorResponse struct {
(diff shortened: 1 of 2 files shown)


server/handler/simulate.go
```diff
@@ -34,14 +34,23 @@ type Simulate struct {
...
 // SimulationResponse is the response returned from Simulate, this is a trimmed down version of common.Result with json
 // tags. This struct and the newSimulationResponse constructor can be extended to include extra content from common.Result.
 type SimulationResponse struct {
-	Name              string `json:"name"`
-	Description       string `json:"description:"`
-	StatusDescription string `json:"status_description"`
-	Status            string `json:"status"`
-	Error             string `json:"error"`
+	Name              string          `json:"name"`
+	Description       string          `json:"description:"`
```
alankpatel: noticed that this field had a colon but did not change it since that would be a breaking change

server/handler/simulate.go
```diff
@@ -34,14 +34,23 @@ type Simulate struct {
...
+	StatusDescription string          `json:"status_description"`
+	Status            string          `json:"status"`
+	Error             string          `json:"error"`
+	PartialErrors     []*PartialError `json:"partial_errors,omitempty"`
```
alankpatel: other option could be to just return HasPartialErrors if we do not want to return all the errors

server/handler/simulate.go
bluekeyes: I think it might be alright to "break" this, as this is clearly a bug. The simulation API should be a less-critical piece of functionality and the description is also not super important field, so I think fixing this is reasonable.

server/handler/simulate.go
bluekeyes: What if we used a `Children []SimulationResult` list? The top level `SimulationResponse` is the overall value for the policy, but then `Children` gives the results for individual rules at each level, following the policy's nesting. This mirrors how `common.Result` is used.

Our use case would traverse this list looking for errors, but other users could use it to see if particular rules were skipped or to inspect other inner details of the policy. The concept of "partial errors" seems very specific to the particular problem we're trying to solve, rather than a generalized concern.

server/handler/simulate.go
alankpatel: alright will edit the field name

server/handler/simulate.go
alankpatel: replaced it with this approach instead. cleans up the code a bit in policy bot and leaves it up to the caller how they want to parse it

bluekeyes (approved):

PR #1196, opened 2026-03-06
Expand Actor to support UserTypes and Not

south-mer:
Reopen of https://github.com/palantir/policy-bot/pull/974, since it was mistakenly rebased to a wrong branch.

## What

Expand the capabilities of `Actor` which is used in `requires`, `author`, etc.

## Why

Allow the expressing predicates such as "author must be user but not certain machine accounts" etc.


asvoboda (commented):
Thank you for splitting up the earlier large PR. Can you also update the documentation here as well?

south-mer:
@asvoboda thank you for the reminder.

Could you take a look at the updated documentation here? I am open to any suggestions.

djahandarie:
@asvoboda Sorry for the repeated ping, would you have a moment to look at this?

The main motivation is to allow us to define a "human identity" (which requires negation in our envrionment), and this is quite a critical thing to be able to do in terms of defining safe rulesets, given lots of non-human identities (AI agents etc) these days.

We'd love to have this upstreamed if possible, let us know if anything extra is required 🙏

PR #1206, opened 2026-03-15
fix(load-error): suppress status for branches without seen policy

### LINKED ISSUE #1179: LoadError posts failing policy-bot status in repos with no policy
When `policy-bot` has installation access to a repo, it attempts to discover the repo's policy before evaluating a PR. If no policy exists, the current behavior is to skip posting status. But if policy discovery hits a non-404 GitHub API error, `policy-bot` posts a failing `error` status. This causes noisy failures in repos that do not define a policy and do not require `policy-bot` as a check.

The README states that a missing policy should not post a status:
```text
If a policy does not exist in the repository or in the shared organization
repository, `policy-bot` does not post a status check on the pull request.
```

The error noise in repos that don't utilize `policy-bot` is caused by this code in `eval_context.go`:
```go
case fc.LoadError != nil:
    msg := fmt.Sprintf("Error loading policy from %s", fc.Source)
    logger.Warn().Err(fc.LoadError).Msg(msg)

    ec.PostStatus(ctx, "error", msg)
    return nil, errors.Wrapf(fc.LoadError, "failed to load policy: %s: %s", fc.Source, fc.Path)
```

Ideally, the fix would distinguish “missing policy” from “unknown due to API error” without reporting failures in repos that don’t use `policy-bot`. One option is to stop posting a failing status on `LoadError` entirely, but where `policy-bot` is required, users will see `policy-bot` not reporting/completing for that commit, without any error details provided. Another option is to post `LoadError` status only for repos/branches where `policy-bot` is required by branch protection/rulesets.

Hoping to get some advice/discussion here from maintainers. Thanks!

dblinkhorn:
## Before this PR

When `policy-bot` is installed at the org level, it attempts to discover policy config in any repo it can access. If that lookup hits a non-404 load error, `policy-bot` posts a failing status even for repos and branches that have never actually used `policy-bot`. This can create noise and confusion in the CI of such repos.

## After this PR

`policy-bot` now tracks whether it has previously found a policy file for a given `owner/repo/base-branch`.

If config loading fails then branches that have previously seen a policy file will still get the load-error status, but branches with no prior seen-policy do not get a failing status.

### Changes
- add a seen-policy cache keyed by owner, repo, and base branch
- mark a branch as having seen a policy when `policy-bot` finds a policy file
- suppress load-error status unless a branch has previously seen a policy
- add unit tests covering cache and status suppression behaviors

## Possible downsides?

There are some pertinent limitations to the "seen-policy" signal:
- `policy-bot` will suppress the first load-error status for a branch that has a policy if that branch has not yet populated the cache
- the cache is per-instance and in-memory, so it is not shared across instances and is reset on restart

Closes #1179.


server/handler/seen_policy_cache.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// SeenPolicyKey identifies a repo branch where policy-bot has seen a policy file.
+type SeenPolicyKey struct {
+	Owner      string
+	Repository string
+	BaseBranch string
+}
+
+type SeenPolicyMetadata struct {
```
bluekeyes: I think we can remove this metadata for now and keep the cache as a simple present/absent test. The one usage of `Get` ignores the metadata and because the cache is in-memory, we can easily change the format in future versions if needed without compatibility concerns.

server/handler/fetcher.go
```diff
@@ -51,6 +65,9 @@ func (cf *ConfigFetcher) ConfigForRepositoryBranch(ctx context.Context, client *
 		}
 
 		if err != nil {
+			if cf.SeenPolicyCache != nil {
```
bluekeyes: It looks like the implementation of `Get` and `Set` on `SeenPolicyCache` also handle the case where the cache is `nil`. I'd prefer to have this check in one place: either a `nil` cache is valid and we don't have to check here or operations on a `nil` cache panic due to nil pointer dereferences and we check here before calling methods.

server/handler/seen_policy_cache.go
```diff
@@ -0,0 +1,70 @@
...
+	SeenAt time.Time
+	Source string
+	Path   string
+}
+
+type SeenPolicyCache struct {
+	entries *lru.Cache
```
bluekeyes: I'm not sure an LRU cache is most appropriate for this usage. We're not storing much data per entry, so I'm not too worried about the size of that cache. At the same time, the correctness of this feature relies on having complete knowledge of seen policies, so arbitrarily expiring entries could lead to skipped statuses when they should be posted.

I think for those reasons, either a simple unbounded cache or a time-based cache may be better. The main reason to expire entries is to account for the case where a repo/branch had a policy that was later removed. That seems rare to me, so I think I'd be fine if those repos either continued to get failed statuses until a Policy Bot server restart or if they continued to get failed statuses until some relatively high expiration time (maybe in the 7-30 day range.)

bluekeyes (commented):
Thanks for proposing this. I'm still a little concerned that this will cause the Policy Bot status to appear "stuck" if a repository that should get status checks is not present in the seen policy cache for some reason (e.g. if the errors happen shortly after restarting the server), effectively shifting the confusion from "why did I get a failed policy-bot status on my repo that doesn't use it" to "why didn't policy-bot post a status". I guess the later is less confusing overall than randomly seeing a status from something you think you don't use.

dblinkhorn:
> effectively shifting the confusion from "why did I get a failed policy-bot status on my repo that doesn't use it" to "why didn't policy-bot post a status". I guess the later is less confusing overall than randomly seeing a status from something you think you don't use.

Thanks for your review!

Yeah this is a legit concern for sure. I guess evaluating the impact of this shift comes down to the more common usage of policy-bot: having access to all repos in the org by default vs access only to specific repos that have a policy defined. I am not sure if this change will solve a niche scenario but create a broader issue, or vice versa.

I think if you're not comfortable with this shift I won't bother addressing the implementation details you highlighted in your comments, and honestly I totally understand if this shift is not acceptable. I really have no idea how many other orgs default to giving policy-bot access to all repos regardless if they use it or not.

bluekeyes:
I discussed this with the team today and the consensus was that the new behavior enabled by this PR is less confusing than the current behavior, i.e. it's less confusing to have a required policy-bot status never appear than it is to get a failed status when you aren't expecting any status. There are also usually several opportunities for Policy Bot to try again in the case where a policy exists, whereas in the case where a policy does not exist, Policy Bot will never replace (and can't remove) the failed status.

I'm willing to try this out (with the improvements from the review) and then we can revisit it or make adjustments if the seen policy cache ends up causing new problems.

server/handler/seen_policy_cache.go
dblinkhorn: Removed the LRU cache in favor of a simple unbounded cache.

server/handler/fetcher.go
dblinkhorn: Removed `nil` check from callers.

server/handler/seen_policy_cache.go
dblinkhorn: Removed metadata.

bluekeyes (dismissed):

bluekeyes (approved):

alankpatel (approved):

PR #1263, opened 2026-06-16
Fail closed when a PR contributor cannot be attributed to a GitHub user

jamestoyer:
## Before this PR

policy-bot's default contributor approval restriction (`allow_contributor` / `allow_non_author_contributor`, both off by default) bans a pull request's contributors from supplying its required approval. The banned set was built solely from `Commit.Users()`, which returns only resolved GitHub logins.

A commit whose author and committer emails map to no GitHub account resolved to an empty login meaning that `Commit.Users()` returned nothing and that contributor silently dropped out of the banned set. A contributer could therefore approve code they had contributed by pushing a commit with an author email tied to no account, defeating the restriction. The bypass left no trace in the approval audit trail as there was no logging around this edge case.

## After this PR

When the contributor restriction is enforced, a pull request containing a commit whose author or committer cannot be attributed to a GitHub account the policy condition no longer passes. This is communicated via the status and logged as a warning for auditing purposes. Commits made through the GitHub API by apps and bots resolve to a `name[bot]` identity and are unaffected, and web-flow commits remain exempt.

==COMMIT_MSG==
Fix a bypass of the contributor approval restriction.
==COMMIT_MSG==

## Possible downsides?

- A pull request containing a commit whose author/committer email is tied to no GitHub account (e.g. CI/automation committing with an unlinked email, or imported/vendored history) will now be held pending under the default restriction, where previously it would have evaluated. The remedy is to re-author the commit with a GitHub-linked email.
- There is currently no way to have a legitimately unattributable commit as `ignore_commits_by` matches on a GitHub login, which such a commit lacks.

### DIFF
--- a/policy/approval/approve.go
+++ b/policy/approval/approve.go
@@ -104,6 +104,34 @@ func (r *Rule) Evaluate(ctx context.Context, prctx pull.Context) (res common.Res
 	}
 	res.PredicateResults = predicateResults
 
+	// When contributors are banned from approving their own changes, every
+	// contributing commit must be attributable to a GitHub user. A commit whose
+	// author or committer cannot be resolved to an account could hide a
+	// contributor who is also an approver, so the contributor restriction
+	// cannot be enforced — fail the rule closed.
+	if r.Requires.Count > 0 && !r.Options.IsAllowContributor() && !r.Options.IsAllowNonAuthorContributor() {
+		commits, err := r.filteredCommits(ctx, prctx)
+		if err != nil {
+			res.Error = errors.Wrap(err, "failed to list commits")
+			return
+		}
+		for _, c := range commits {
+			if desc, ok := c.UnattributedContributor(); ok {
+				log.Warn().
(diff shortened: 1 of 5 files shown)


bluekeyes:
We discussed this more internally, but I think this only partially fixes the underlying issue. Even with the change here, commit authorship can still be forged and attributed to any other arbitrary user with a known email address, bypassing the new protections.

Ultimately, I think this is something we need GitHub to fix. On GHES, the `commit-current-user` pre-receive hook effectively blocks this. For GitHub.com, I think requiring signed commits is the only existing rule that might help. It's unfortunate that there's no ruleset equivalent for the `commit-current-user` check yet.

We also have this limitation documented in [the README](https://github.com/palantir/policy-bot#commit-users-), although it doesn't explicitly explain how it can lead to a contributor-approval bypass.

If we do go through with this limited version of the fix, I think it would be good to make it configurable via a rule-level option so people can preserve the old behavior if needed (say in forks that merge upstream changes frequently.)

PR #1302, opened 2026-07-13
Allow validate command to test remote config syntax

akerl:
## Before this PR
The Validate API endpoint will test both remote and local configs, but the validate subcommand errors on remote configs.

## After this PR

==COMMIT_MSG==
Extend the validate subcommand to handle remote configurations.

As with API endpoint, only checks remote config for syntax; does not load the actual referenced config.
==COMMIT_MSG==

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -18,6 +18,7 @@ import (
 	"io/fs"
 	"os"
 
+	"github.com/palantir/go-githubapp/appconfig"
 	"github.com/palantir/policy-bot/policy"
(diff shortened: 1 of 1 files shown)


bluekeyes (approved):

akerl:
Would it be possible to cut a release to get this into the `@latest` docker image

bluekeyes:
[v1.41.2](https://github.com/palantir/policy-bot/releases/tag/v1.41.2)

PR #1323, opened 2026-07-28
Request reviews from teams with inherited repository permissions

### LINKED ISSUE #165: request_review expect a team to be a contributor
Hello,

I face an issue with `request_review` feature. At least I consider it buggy.

I have a team of team (let's call it MegaTeam) in github organization and my repository has read only for MegaTeam. The RequestorTeam is part of MegaTeam and I configured policy bot with the following.

*Given*:
```
policy:
  approval:
    - policy

approval_rules:
  - name: policy
    options:
      request_review:
      enabled: true
      mode: all-users
    requires:
      count: 2
      teams: ["myorg/requestorteam"]
      admins: false
      write_collaborators: false
```
*When* pull request is created by anyone
*Then* 0 users are assigned to the pull request
*When* `requestorteam` is added as read collaborator
*And* Pull request is created
*Then* all member of the *requestorteam* are assigned

I expect policy to be able to "see" that requestorteam is part of collaborator, instead to explicitly add them.

hiyosi:
## Before this PR

When `request_review` is enabled with `mode: teams`, policy-bot only requests reviews from teams that appear in the repository team list (`GET /repos/{owner}/{repo}/teams`). Teams whose access is only inherited from a parent team do not reliably appear in that list, so they are silently skipped — even though the GitHub API accepts review requests for such teams (confirmed manually, and noted in #165 where a maintainer reproduced requesting a child team whose parent had read access via the GitHub UI).

Fixes #165

## After this PR
==COMMIT_MSG==
Request reviews from teams with inherited repository permissions
==COMMIT_MSG==

Teams listed in `requires.teams` now receive review requests even when their repository permission is inherited from a parent team.

Implementation:

- After matching against the repository team list as before, any remaining requested teams are resolved individually via the [check team permissions endpoint](https://docs.github.com/en/rest/teams/teams#check-team-permissions-for-a-repository) (`Teams.IsTeamRepoBySlug`), which is documented to include permissions inherited through parent teams
- Adds `TeamPermission(team)` to `pull.Context`, implemented with a per-team cache on `GitHubContext`; extra API calls are only made for requested teams missing from the repository team list
- Teams with no access at all resolve to 404 → `PermissionNone` and remain excluded; teams from other organizations are ignored, as before; permission-based selection (`requires.permissions`) is unchanged
- Updates the README caveat for `mode: teams` accordingly

Verification, in addition to the new unit test:

- Verified against a real organization where child teams inherit access from an umbrella parent team: `IsTeamRepoBySlug` + `ParseRepositoryPermissions` resolves inherited-only teams to `write` (404 for unknown teams), with both a user token and a GitHub App installation token holding org Members read — the same class of credentials policy-bot uses
- Verified that `POST /pulls/{n}/requested_reviewers` succeeds for an inherited-access team that is not a direct repository collaborator

## Possible downsides?

- One additional API request per requested team that is not in the repository team list, per evaluation (cached within an evaluation context)
- Behavior change: teams that were previously skipped silently will start receiving review requests. Organizations relying on the old behavior to suppress requests for inherited-access teams will see new notifications

### DIFF
--- a/README.md
+++ b/README.md
@@ -1030,8 +1030,8 @@ supported options:
 
  * `all-users` to request all users who can approve
  * `random-users` to randomly select the number of users that are required
- * `teams` to request teams for review. Teams must be repository collaborators
-   with at least read access.
+ * `teams` to request teams for review. Teams must have at least read access
+   to the repository, either directly or inherited from a parent team.
 
 ```yaml
 options:
--- a/policy/reviewer/reviewer.go
+++ b/policy/reviewer/reviewer.go
@@ -20,6 +20,7 @@ import (
 	"math/rand"
 	"slices"
 	"sort"
+	"strings"
 
 	"github.com/palantir/policy-bot/policy/common"
 	"github.com/palantir/policy-bot/pull"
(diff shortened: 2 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @hiyosi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1330, opened 2026-08-03
Trigger release workflow on tag push instead of a release event

tabboud:
## Before this PR
GitHub's immutable-releases feature rejects asset uploads to a release once it's published, so releases must be created as drafts, have artifacts uploaded, then get un-drafted.

## After this PR

==COMMIT_MSG==
Trigger release workflow on tag push instead of release event to support immutable releases
==COMMIT_MSG==

The workflow now triggers on push of a version tag and assets are published appropriately for snapshot/releases. Autorelease will now create a draft release, `./godelw publish github` will reuse the existing draft release to upload all assets, and will then finalize (un-draft) the release, making it immutable. The version of `godelw` was verified to support this new draft/upload/finalize workflow.


asvoboda (dismissed):

tabboud:
@asvoboda I had to fix up a merge conflict which invalidated the approval. Mind taking another look?

jamestoyer (approved):

Issue #1333, opened 2026-08-05
Log GitHub request IDs and HTTP status via custom http.RoundTripper


dblinkhorn:
We have recently been having intermittent issues with failures loading policies from GitHub. When this occurs, the initial responses are `500` errors and Policy Bot retries the requests as expected, but since logging runs outside the `httpcache` layer, it records only the response returned by the cache on these subsequent retries. It does not expose the HTTP status or `X-GitHub-Request-Id`. GitHub Support asked us to provide the request IDs to help investigate this issue, but we aren't able to do so with this limitation.

What would be your thoughts on adding a custom `http.RoundTripper` so that logging could expose the raw HTTP status and request ID for each GitHub response? I think this could provide some useful diagnostic info when investigating problems involving the GitHub API and how Policy Bot is handling its responses.

bluekeyes:
I'm open to improving things here to make it easier to debug issues with GitHub, but I think these are changes that belong in https://github.com/palantir/go-githubapp.

Adding the request ID to the logs should be a simple change that seems safe - I'm not sure why we didn't include that originally. Take a look at [middleware_logging.go](https://github.com/palantir/go-githubapp/blob/d6142bde76e4eb6ffe841c3b14ea00aa8eff93f0/githubapp/middleware_logging.go#L41).

I think I'll need more information to evaluate the status and retry issue. We order the logging middleware outside the caching middleware so that we can log the cache status. That said, from what I remember the `go-githubapp` client doesn't do any retries by default and we don't configure client-level retries in Policy Bot. The retries for loading policies happen outside the client, which means each request that's part of a retry sequence should appear as a distinct request in the logs. I don't think the cache should be caching 500 responses. If it is, that's a different bug that we should fix.

dblinkhorn:
Reviewing our Policy Bot logs is where we saw it reporting that we were getting cached 500s:

```
{"level":"debug","github_event_type":"workflow_run","method":"GET","elapsed":559.319095,"cached":false,"status":500,"size":101,"time":"2026-08-05T15:01:10.667137762Z","message":"github_request"}
{"level":"debug","github_event_type":"workflow_run","method":"GET","elapsed":238.37456,"cached":true,"status":500,"size":101,"time":"2026-08-05T15:01:11.906511442Z","message":"github_request"}
[...]

PR #1345, opened 2026-08-17
Fix 500 errors being cached by httpcache

psalaberria002:
## Problem

When GitHub API returns a 500 error, the httpcache library caches the error response. Consecutive requests for the same resource return the cached 500 error instead of retrying the API, causing policy evaluation to fail repeatedly.

**Root cause:** The httpcache library's `canStore()` function only checks for `no-store` directives and doesn't validate HTTP status codes. This violates RFC 7231, which requires that 5xx responses must not be cached by default.

## Solution

Add an error filtering transport middleware that intercepts responses and strips cache headers from 5xx responses before they reach httpcache. This prevents error caching while preserving conditional request support (ETags, 304 Not Modified) for successful responses.

### How it works

- Transport strips ETag, Cache-Control, Expires, Last-Modified headers from 5xx responses
- Headers remain intact on success responses (200, 404, etc.)
- Allows conditional requests to work normally
- RFC 7231 compliant

### Impact

- Fixes repeated 500 errors blocking policy evaluation
- Maintains rate limit efficiency via conditional requests
- No impact on other API calls or functionality

Related to https://github.com/palantir/policy-bot/issues/1333

### DIFF
--- a/server/error_filter_transport.go
+++ b/server/error_filter_transport.go
@@ -1,0 +1,39 @@
+package server
+
+import (
+	"net/http"
+)
+
+// errorFilteringTransport wraps an http.RoundTripper and prevents 5xx error
+// responses from being cached. This fixes the issue where httpcache caches
+// error responses in violation of RFC 7231.
+type errorFilteringTransport struct {
+	transport http.RoundTripper
+}
+
+// NewErrorFilteringTransport creates a new error filtering transport that wraps
+// the provided RoundTripper.
+func NewErrorFilteringTransport(transport http.RoundTripper) http.RoundTripper {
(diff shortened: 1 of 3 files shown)


asvoboda (commented):
I think there is a bug here, but looking at the code I think it would be better to put this in [go-githubapp ](https://github.com/palantir/go-githubapp/) more directly.

psalaberria002:
> I think there is a bug here, but looking at the code I think it would be better to put this in [go-githubapp ](https://github.com/palantir/go-githubapp/) more directly.

I am not very familiar with the codebase, and I do not have time to look into it unfortunately. I will close this for now.

Issue #111, opened 2019-07-19
Support remote policy files in private repositories


NargiT:
Hello,

I would like to know why the remote policy configuration has to be full public.
We would like to use it with a github entreprise on premise version where everything is private.

Since policy bot is installed on the organization level, I do not see any blocker to also allow the remote configuration policy to be access privately (using token).

Thx in advance for clarification, I can help to provide this feature.

bluekeyes:
When you say that "everything is private" do you mean that your GitHub Enterprise instance is in [private mode](https://help.github.com/en/enterprise/2.17/admin/installation/enabling-private-mode) (you must authenticate to see any content) or that all repositories are [private](https://help.github.com/en/enterprise/2.17/user/articles/setting-repository-visibility#making-a-repository-private) (even if you are authenticated, you can only see repositories where you have permission)?

If you are using "private mode", Policy Bot will work fine with remote policies; this is how we run it internally.

If you are using actual private repositories, I believe Policy Bot already supports reading from private repositories in the _same_ organization. I haven't tested this, but the code uses an authenticated organization client when fetching configuration. This requires that the app is enabled for the target repository (or for all repositories in the organization.)

Private repositories in _other_ organizations are not supported. While we could support this if Policy Bot is installed on the target organization, it makes the code more complicated and I'm not sure it is required.

NargiT:
we use it in "private mode" and with "private permission". But we are working in multiple organizations and policy bot is already installed in all of them. This feature is already supported by "has_contributor_in:"  that why I though it would not be a problem for remote policy.

In our case we have multiple organizations because we manage a lot of thing automatically and we do not want to mix "developer friendly" repo vs "robot friendly" repo.

Do you think it would be possible to integrate this feature?

NargiT:
Little clarification: https://github.com/palantir/policy-bot#remote-policy-configuration
I am talking about this case

> If you are using "private mode", Policy Bot will work fine with remote policies; this is how we run it internally.

Maybe another permission is required ? Is there a way to run policy bot in verbose/debug mode to see this kind of errors ?

bluekeyes:
[...]

PR #1359, opened 2026-08-25
Support cross-org remote policy references

clarkemn:
## Before this PR
A `.policy.yml` can reference a remote policy file via `remote: org/repo`, but this only works if the remote repo is in the same GitHub org as the requesting repo. The fetch reuses the requesting repo's own app installation client, which has no access outside its org, so a cross-org reference fails (typically a 404) even when policy-bot is installed on the target org too.

This blocks a common setup: one centrally managed policy repo referenced via `remote:` from many other orgs in the same GitHub Enterprise. (The vendored `go-githubapp/appconfig` package has a commented-out `WithPrivateRemotes` stub referencing palantir/policy-bot#111 describing this exact gap.)

## After this PR

A `remote:` reference can now point to a repository in a different GitHub org than the one defining it, as long as policy-bot is also installed on that org — resolved by looking up the target org's own app installation
(the same lookup `CrossOrgMembershipContext` already uses for cross-org team/org membership checks). Same-org remotes are unaffected: no installation lookup happens, the original client is reused exactly as before. Only 1 level of remote configuration is still supported, and a missing/misconfigured remote file still produces the same "invalid remote reference: file does not exist" error as before.

## Possible downsides?

None expected for existing setups — same-org remote references and local policy files go through unchanged code paths. For a genuinely cross-org `remote:` reference, the resolution now depends on policy-bot's app being installed on the target org: if it isn't, the failure mode changes from a generic 404 to a clearer wrapped error ("policy-bot is not installed on organization ..."), which is a behavior change for anyone currently relying on that specific 404 message/error type, though none is known.


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @clarkemn! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

asvoboda:
Policy bot definitely works with cross org policies today, as we use this functionality extensively internally on GHES (currently 3.19.x).

Is policy-bot installed on both organizations, so it can read the policy?

clarkemn:
Hi @asvoboda 

We are not on GHES (we are on Github.com) and have a single GitHub App registered at the enterprise level, installed separately on each organization (so it has its own installation per org, not one shared cross-org installation). 

We hit an error when testing this cross-org access where a PR in one of our orgs referenced remote: `<other-org>/policy-bot` (a private repo), and got:
```
Error loading policy from <other-org>/policy-bot
invalid remote reference: file does not exist
```
The file does exist — the message is misleading. Tracing `appconfig.Loader.loadRemoteConfig`: it fetches the remote file using the requesting repo's own installation client, not a client for the org that owns the remote repo. Since `<other-org>/policy-bot` is private and our requesting org's installation was never granted access to it, GitHub returns a 404 (by design — it doesn't distinguish "doesn't exist" from "no access" for private repos), which the vendored loader maps directly to `invalid remote reference: file does not exist`.

So back to your questions - yes, policy-bot is installed on both organizations — but each installation only has access to its own org's repos. The existing code never looks up the target org's installation for a cross-org remote: reference, it just reuses whatever client the caller already has.

bluekeyes:
Thanks for clarifying. As you found (and as noted in #111), it's expected that Policy Bot cannot load remote policies from private (or internal) repositories right now. 

To add this support, I think the correct fix is actually in https://github.com/palantir/go-githubapp, another library that we maintain. As noted in your PR body, something like the `WithPrivateRemotes` option is likely a better fit: it avoids duplicating config loading logic here and means anyone else using the library can also take advantage of this support.

At a high level, when this option is present, I think you want to modify the `loadRemoteConfig` method so that it attempts to use an alternate client when the owner in the remote reference differs from the owner passed to the `LoadConfig` call. Because both `ClientCreator` and `InstallationsService` have caching variants, we can probably omit caching from `ConfigLoader` (unless it's trivial to add) and recommend using the caching versions in the option function docs. If there's no installation (or constructing the client fails) make sure to log this and then fallback to the existing client to handle the public repository case.

Once we release the change as part of `go-githubapp`, we can update the version here and add the new option.

clarkemn:
@bluekeyes Thanks for taking a look! https://github.com/palantir/go-githubapp/pull/618 is now open. If we're able to get that merged and released, I'll update this PR to take advantage of the new option

bluekeyes:
Version 0.47.2 of `go-githubapp` has this change, if you want to update this PR. Per the security note on the new option, I think we'll want to also add a server option (default disabled) to Policy Bot that controls whether it can load policies from private remotes or not.

PR #1386, opened 2026-09-15
Support GitHub reactions as approval candidates (#1254)

### LINKED ISSUE #1254: Support GitHub reactions as approval candidates
## Feature request

Please consider supporting GitHub reaction objects as approval candidates or policy predicates.

## Use case

Some GitHub Apps and review bots signal a successful review by adding a `+1`
reaction to the pull request issue instead of posting a pull request comment or
submitting an approving review. In that case, Policy Bot cannot currently model
the bot signal directly with policy configuration.

For example, a policy author may want a rule like:

```yaml
approval_rules:
  - name: review bot reacted
    options:
      methods:
        reactions:
          issue:
            content: "+1"
    requires:
      count: 1
      users:
        - "chatgpt-codex-connector[bot]"
```

The exact shape is only a strawman. The important part is being able to count a
GitHub reaction from a required user/app account the same way Policy Bot can
currently count comments, comment patterns, GitHub reviews, review-comment
patterns, and pull-request body patterns.

## Current workaround

The workaround is to run a separate bridge check that calls GitHub's reactions
API, verifies that the required bot account added a `+1` reaction to the PR
issue, then publishes a commit status or check run that Policy Bot can require
through `has_status`.

That works, but it means every repository or organization that wants to use this
signal needs to maintain extra automation just to translate a GitHub approval
signal into something Policy Bot already understands.

## Semantics to consider

- The first useful target would be reactions on the pull request issue itself:
  `GET /repos/{owner}/{repo}/issues/{issue_number}/reactions?content=%2B1`.
- It would be helpful if reaction approval candidates honored existing
  `requires.users`, `requires.organizations`, and `requires.teams` checks.
- If `invalidate_on_push` is enabled, a reaction created before the latest
  relevant push should probably be treated similarly to a stale approval
  comment, using the reaction `created_at` timestamp.
- If GitHub does not send webhook events for reaction changes, it would still be
  useful for Policy Bot to read reactions during normal policy evaluation and
  document the re-evaluation behavior.

## Why this belongs in Policy Bot

Policy Bot already lets teams model nuanced approval signals declaratively. A
reaction approval method would keep the policy in one place and avoid requiring
custom status-bridge automation for bots whose only public approval signal is a
GitHub reaction.

ZayanKhan-12:
Fixes #1254.

## First, the webhook question

@asvoboda asked whether GitHub supports webhooks for reactions. **It does not.** There is no
`reaction` event in [the webhook events and payloads reference][webhook-events] — the list runs
`branch_protection_configuration` through `workflow_run` with nothing for reactions, and none of
the `issues`, `pull_request`, `issue_comment` or `pull_request_review` payloads carry reaction
changes. The linked discussion generalises correctly.

So the concern is real and I did not design around it by pretending otherwise. What I did
instead:

- A rule that uses reactions reports **`TriggerAll`**, so it is re-evaluated on *every* other
  event for the pull request. A reaction is picked up by the next comment, review, push, status
  or check run, and opening the details page re-evaluates immediately.
- The README states the limitation plainly in its own section, including the case where it
  would behave badly, and recommends comments or reviews when a reaction would be the *only*
  signal.

For the use case in the issue — a bot whose review finishes — this works well in practice,
because such bots almost always also produce a check run or status, which does deliver an
event. The honest failure mode is a human reacting on an otherwise idle pull request: the
status will not move until something else happens. That is documented rather than hidden.

If you would still rather not ship a signal without a webhook, that is a reasonable call and
I am happy for this to be closed — the answer to the webhook question above is the part worth
keeping either way.

## The change

```yaml
approval_rules:
  - name: review bot approved
    options:
      methods:
        comments: []
        comment_patterns: []
        github_review: false
        reactions: ["+1"]
    requires:
      count: 1
      users: ["review-bot[bot]"]
```

`reactions` counts a reaction on the pull request itself. Values are GitHub's content names
(`+1`, `-1`, `laugh`, `confused`, `heart`, `hooray`, `rocket`, `eyes`).

Reaction candidates run through the same pipeline as every other method, so the semantics the
issue asked about all fall out rather than being special-cased:

- `requires.users`, `requires.organizations` and `requires.teams` apply unchanged.
- `invalidate_on_push` dismisses a reaction created before the last push, using the reaction's
  `created_at` — exactly as suggested.
- Reactions cannot be edited, so `LastEditedAt` is always zero and `ignore_edited_comments`
  never applies to them.
- Because `common.Methods` is shared, disapproval and revocation get the method too.

## Design notes

**It costs nothing unless you use it.** `Reactions()` is only called when a rule configures the
method. It loads on demand over REST instead of joining `loadPagedData`'s GraphQL query — that
query is tuned to one rate-limit point and its comment asks that the cost be verified before
adding fields, so adding a fourth connection would have made every installation pay for a
feature almost none of them use. This follows the existing `Labels()` precedent.

**Invalid reactions fail at parse time.** `reactions: [":+1:"]` is the likely mistake, since
that is how the `comments` method is spelled. Rather than let it become a rule that silently
never matches, `ReactionContent` validates on unmarshal, following the `common.Regexp`
precedent:

```
invalid reaction ":+1:": must be one of +1, -1, laugh, confused, heart, hooray, rocket, eyes
(GitHub's reaction content values, not emoji or emoji codes)
```

**`TriggerAll` is deliberate**, not laziness. `Trigger` is documented as the set of events that
could change a rule's value; for reactions that is genuinely "all of them", because the change
itself is invisible. It is also why the method is opt-in — no existing policy changes trigger
behaviour.

## Testing

`./godelw verify --apply=false` passes — format, mod, license, lint (0 issues) and the full
suite.

26 new subtests:

- `policy/common/methods_test.go` — candidate extraction, multiple contents, dedup across a
  comment and a reaction by the same user, and a guard that `Reactions()` is **not** called when
  no reaction method is configured (the context returns an error if it is, so the test fails if
  the lazy-loading regresses).
- `policy/common/reaction_test.go` — all eight content values parse, unknown values and emoji
  codes are rejected, and the defaults chain (inherit / override / explicit-empty-disables).
- `policy/approval/approve_test.go` — approval by reaction, non-matching content, `requires.users`
  filtering, `invalidate_on_push` both dismissing and sparing a reaction, `ignore_edited_comments`
  not touching reactions, and the `TriggerAll` assertion.
- `pull/github_test.go` — `Reactions()` over a paginated `ResponsePlayer` fixture, plus a
  caching test asserting the endpoint is hit exactly once for a PR with no reactions. That one
  is load-bearing: with `var reactions []*Reaction` instead of an empty slice literal, an empty
  result is never cached and every evaluation re-requests it. I checked the test fails on that
  mutation (`expected: 1, actual: 2`).
- `policy/policy_test.go` — end-to-end from a `policy.yml` string through `ParsePolicy` to an
  approved/pending status.

## Note on the second commit

`Add CLAUDE.md` is unrelated to the feature and separated so it can be dropped with a single
`git rebase`.

[webhook-events]: https://docs.github.com/en/webhooks/webhook-events-and-payloads


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @ZayanKhan-12! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

asvoboda:
Regarding `If you would still rather not ship a signal without a webhook, that is a reasonable call`

I think it would be better to hapy policy-bot be entirely event-driven, and to not use reactions unless github supports sending webhooks. If and when github supports this feature, we can re-open the issue and re-investigate the PR.

PR #1387, opened 2026-09-15
Explain why an approval was ignored (#766)

### LINKED ISSUE #766: Clarify why users are "disqualified" when approval is ignored
Hit the same issue as in https://github.com/palantir/policy-bot/issues/174. It is hard to tell that approval is disqualified because approver contributed to the PR - the only way is to look through all commits.

It would be good to have a clarification saying "disqualified due to contributing, [link to commit]".

ZayanKhan-12:
Fixes #766.

## The problem

When an approval does not count towards a rule, the status says only:

```
0/1 required approvals. Ignored 1 approval from disqualified users
```

That does not say *who* was ignored or *why*. As the issue puts it, the only way to find out
whether an approver was disqualified for contributing is to read every commit on the pull
request by hand.

The interesting part is that `policy-bot` already knows the answer. `isApprovedByActors`
distinguishes three cases — the author, contributors, and users the rule does not require —
and then throws the distinction away into a debug log:

```go
if banned[c.User] {
    log.Debug().Str("user", c.User).Msg("rejecting approval by banned user")
    continue
}
```

So this is mostly a matter of keeping information that was being discarded, not computing
anything new.

## The change

The banned-user map now records *why* a user is banned and, for a contributor, the commit
that caused it. `isApprovedByActors` returns a `Disqualification` per rejected candidate,
carried on `RequiresResult` next to `Approvers` — mirroring the existing `Dismissal{Candidate,
Reason}` type.

The status names the reason:

```
0/1 required approvals. Ignored 1 approval from a contributor to this pull request
0/1 required approvals. Ignored 2 approvals from users this rule does not require
0/1 required approvals. Ignored 4 approvals: 1 author, 2 contributors, 1 not required
```

The full wording is used when every ignored approval shares a reason, and a per-reason
breakdown otherwise. Both stay inside GitHub's 140-character limit — there is a test asserting
that, since the description is built by concatenation and it would be easy to grow past it.

This also fixes the old message saying "disqualified **users**" when only one user was ignored.

The details page lists each ignored approver with their reason and **links the commit** that
made someone a contributor, which is what the issue asked for:

```
These approvals were not counted:
  mhaypenny             Opened this pull request
  contributor-author    Contributed commit 6748325   <- links to the commit
  outsider              Not one of the users this rule requires
```

The README gains a short section mapping each phrase back to the option that caused it
(`allow_author`, `allow_contributor`, `requires.users`, …).

## Compatibility

No evaluation behaviour changes — exactly the same candidates are approved and ignored as
before. Only the description changes.

The simulation API is unaffected: `SimulationResponse` is a trimmed struct that does not
include `RequiresResult`, so its JSON shape is unchanged.

## Testing

`./godelw verify --apply=false` passes — format, mod, license, lint (0 issues) and the full
suite.

New tests:

- `TestStatusDescriptionDisqualifications` — a table over the wording: singular vs plural for
  each reason, the mixed-reason breakdown, the fixed reason ordering, and the case where a rule
  requires no actors so nothing should be said at all. Every case also asserts the result is
  ≤ 140 characters.
- `reportsWhyEachApprovalWasIgnored` — asserts the reason for all seven candidates in the
  existing fixture, keyed by user so it does not depend on candidate ordering.
- `namesTheCommitThatDisqualifiedAContributor` — asserts the exact SHA recorded for each
  contributor.
- `allowedContributorsAreNotDisqualified` — guards the `allow_contributor` path.
- `TestDetailsTemplateRendersDisqualifications` / `...OmitsEmptyDisqualifications` in
  `server/handler` — these render `details.html.tmpl` for real and assert the commit link is
  present. There was no test parsing the templates before, so a typo in one was only
  discoverable by loading the page; while writing these I hit exactly that, which is why they
  are here.

14 existing assertions in `approve_test.go` are updated for the new wording. I updated them
per-subtest rather than with a blanket replace, because the same old string maps to different
new strings depending on the rule. As a check that this is description-only, the counts in
every new message add up to the count the old message reported (`7` → `1 author, 2
contributors, 4 not required`, `8` → `1 + 1 + 6`, and so on), and I verified the seven-candidate
breakdown by hand against the fixture before trusting the output.

## Note on the second commit

`Add CLAUDE.md` is unrelated to the fix and separated so it can be dropped with a single
`git rebase`. It is byte-identical to the file in #1386, so the two PRs will not conflict.


palantirtech:
Thanks for your interest in [palantir/policy-bot](https://github.com/palantir/policy-bot), @ZayanKhan-12! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

